PR #149, opened 2014-05-31
Bump racc dep to 1.4.11

bbatsov:
There was a TODO regarding this.


bbatsov:
Just a quick update here - `racc` doesn't seem to be maintained these days. There has been zero activity after the 1.8 breakage/fix. Wish it had a maintainer like you, but alas - it clearly doesn't.

whitequark:
:tada:

PR #178, opened 2015-01-02
Upgrade to slop 4

deivid-rodriguez:
Maybe @leejarvis can review this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,6 @@
 source 'https://rubygems.org'
 
+gem 'slop', '~> 4.0', platforms: [:ruby_20, :ruby_21, :ruby_22]
+
 # Specify your gem's dependencies in parser.gemspec
 gemspec
--- a/lib/parser/runner.rb
+++ b/lib/parser/runner.rb
@@ -12,7 +12,7 @@ module Parser
     end
 
     def initialize
-      @slop         = Slop.new(:strict => true)
+      @options      = Slop::Options.new
       @parser_class = nil
       @parser       = nil
       @files        = []
@@ -38,10 +38,10 @@ module Parser
     end
 
     def setup_option_parsing
-      @slop.banner "Usage: #{runner_name} [options] FILE|DIRECTORY..."
+      @options.banner = "Usage: #{runner_name} [options] FILE|DIRECTORY..."
 
-      @slop.on 'h', 'help', 'Display this help message and exit', :tail => true do
-        puts @slop.help
+      @options.boolean '-h', '--help', 'Display this help message and exit' do
+        puts @options.help
         puts <<-HELP
 
   If you specify a DIRECTORY, then all *.rb files are fetched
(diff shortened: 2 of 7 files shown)


lib/parser/runner.rb
```diff
@@ -52,50 +52,49 @@ def setup_option_parsing
...
       end
 
-      @slop.on 'V', 'version', 'Output version information and exit', :tail => true do
+      @options.boolean '-V', '--version', 'Output version information and exit' do
         puts "#{runner_name} based on parser version #{Parser::VERSION}"
         exit
       end
 
-      @slop.on '18', 'Parse as Ruby 1.8.7 would' do
+      @options.boolean '-18', 'Parse as Ruby 1.8.7 would' do
```
whitequark: Should be `--18`, or even better, `--1.8`.

lib/parser/runner.rb
```diff
@@ -52,50 +52,49 @@ def setup_option_parsing
...
 
-      @slop.on 'e=', 'Process a fragment of Ruby code' do |fragment|
+      @options.string '-e', 'Process a fragment of Ruby code' do |fragment|
         @fragments << fragment
       end
     end
 
     def parse_options(options)
-      @slop.parse!(options)
+      @slop = Slop::Parser.new(@options).parse(options)
```
leejarvis: You could do `@options.parse(options)` here I think

deivid-rodriguez:
Ooooops, `slop 4` requires at least `Ruby 2.0`...

leejarvis:
Yep, that might be a problem here ;)

whitequark:
Requested in #179.

Can you please update PR and add a conditional that would fail the CLI if ran on <=1.9? I think it's not important to support CLI on <=1.9.

bbatsov:
Yeah, this seems sensible.

bbatsov:
Alternatively the simple cli can be extracted in another gem - say `parser-utils` or something like that.

whitequark:
Actually, that wouldn't quite work, since slop 4 is not installable on <=1.9. I think we can use this to make slop an optional dependency; ruby-parse should either require slop or complain.

I don't want to extract CLI into another gem.

bbatsov:
> I don't want to extract CLI into another gem.

Nobody would want to do something like that, but it's one way to resolve such a problem once and for all. As parser is used mostly as a library, having such a dependency is a bit of a problem (as we see with slop 4). Perhaps another way would be to simplify the cli tools and stop using slop in them.

whitequark:
Well, slop is the only dependency, so we won't have this problem ever again. Could as well bite it in this case and add workarounds for slop being optional.

bbatsov:
Fair enough. Btw, why did you chose it over the built-in `optparse`?

whitequark:
I never considered it. It sounded like too much work. I'm open to converting CLI to optparse if it doesn't make the interface (either of) too terrible.

deivid-rodriguez:
@whitequark So your decision is to drop CLI support for rubies under 2.0.0?

whitequark:
@deivid-rodriguez Yes. I don't think there is much value in having it, and it seems a straightforward solution.

leejarvis:
Whilst I'd of course prefer that parser used Slop, it'll probably always be a bit of a pain to cover all Ruby versions (if that's the intention of Parser, of course). In that instance we should just use optparse.

deivid-rodriguez:
I'm updating the PR in any case. I guess I need to skip some tests, but I don't have Ruby 1.9.3 installed so I'll wait for Travis to tell me.

deivid-rodriguez:
Travis is failing in ruby-head and jruby. Don't know whether it has the do with this PR or with recent changes in the failing rubies...

whitequark:
The failures are unrelated...

bbatsov:
@leejarvis Yeah, having the same behaviour on all supported rubies is generally the best option.

Gemfile
```diff
@@ -1,4 +1,6 @@
 source 'https://rubygems.org'
 
+gem 'slop', '~> 4.0', platforms: [:ruby_20, :ruby_21, :ruby_22]
```
whitequark: Uh, Gemfile? You do know that Gemfile is completely ignored by Rubygems?

deivid-rodriguez: Right, it will work if installed through bundler but not with `gem install`. I'd just migrate to optparse then...

bbatsov: @deivid-rodriguez :+1:

whitequark: I actually don't think it will work even if installed via Bundler.

deivid-rodriguez: I'll try it to satisfy my curiosity! :smile:

I'm currently very busy but I'll try to migrate to optparse when I get time if nobody does it before. I recently migrated byebug and it was pretty painless.

deivid-rodriguez: You were right @whitequark , it doesn't work via Bundler either! :+1:

PR #183, opened 2015-01-11
Fix anonymous restarg and kwrestarg in same signature

JoshCheek:


test/test_parser.rb
```diff
@@ -1980,6 +1980,13 @@ def test_kwarg_combinations
         s(:blockarg, :b)),
       %q{**baz, &b},
       ALL_VERSIONS - %w(1.8 1.9))
+
+    assert_parses_args(
+      s(:args,
+        s(:restarg),
+        s(:kwrestarg)),
+      %q{*, **},
+      ALL_VERSIONS - %w(1.8 1.9))
```
bbatsov: This is repeated quite often. Guess we can introduce a constant `RUBY_2_OR_NEWER`.

whitequark: I prefer listing them explicitly.

Issue #185, opened 2015-02-02
Comment.associate 


kubicle:
Code like below is not associating the comment:
```
def foo
  p "hello" # decoration comment on last line 
end
```

kubicle:
Similarly, but less frequent, comments at the end of a file are not associated. 
I am planning to fork the repo and propose an "associator" that can make the difference between "decoration" comments (attached at the end of a line) and other comments. 
If I am the only one with this need, feel free to close my PR.

JoshCheek:
I hit the end of file comments, too. I deal with it by checking the tokens https://github.com/JoshCheek/seeing_is_believing/blob/267b9e8f5a8cdff268658376bb10670de102c58f/lib/seeing_is_believing/code.rb#L123-130

whitequark:
This is really a parser bug :/

Issue #184, opened 2015-01-24
Comment.associate "merging" comments for identical nodes


kubicle:
Hi and thanks for Parser, it works great!

I seem to have a problem with the comments "associate" method. For example the code below:
```ruby
a=1 # first time
f()
a=1 # second time
f()
```
will give me the following mapping:
```
{(send nil :f)=>[
 #<Parser::Source::Comment test1.rb:2:5 "# first time">,
 #<Parser::Source::Comment test1.rb:4:5 "# second time">
]}
```
So it looks like both calls to "f" have 2 lines of comment instead of 1 line each.

whitequark:
Wow. That was a really dumb decision on my side. Indeed #associate is broken.

whitequark:
@bbatsov @yujinakayama @YorickPeterse  have you bumped into this bug? (How could you possibly have not?!)

What do you think is the best course of action here? On a second thought it looks like `AST::Node#hash` and `AST::Node#eql?` as currently defined are a bug factory. I want to switch them to compare-by-identity instead, but not sure yet what would be the consequences.

YorickPeterse:
@whitequark Not that I know of. In my case I only use comments for method definitions where this problem is probably less/not likely to occur.

whitequark:
No, they will not if I implement this change, which is why I am asking.

bbatsov:
I concur that the current behaviour has to be changed. The impact to RuboCop's codebase should be minimal.

whitequark:
On reflection, I think that:
  * The change to _ast_ would be incredibly hard to safely roll out, as most gems depend on _ast_ transitively. It is probably not an option.
  * _#associate_ actually abuses the hash. Logically it returns not a hash table but a list of associations. So it probably should be fixed to return an array of comment-node pairs. As this changes the public API, a new function will have to be created.

YorickPeterse:
@whitequark To give some context, I depend on separate instances being considered equal in tests like this: https://github.com/YorickPeterse/oga/blob/f94461a9cadd3858bbf2bc5ee39484b865a5af96/spec/oga/xpath/parser/calls_spec.rb#L6

If two separate instances were no longer the same I'd have to fix a few thousand specifications spread across different projects.

whitequark:
No, that's `==`. I've never talked about changing `==`, but only `eql?` and `hash`.

YorickPeterse:
Ah ok, I thought it would apply to `==` as well. In that case this should be fine at least for me.

[...]

PR #188, opened 2015-03-07
* Changed Comment::Associator#associate API and way of associating comments

kubicle:
This PR makes the "Comment Associator" behave differently to solve 2 issues and allow callers to make the difference between "heading" comments, which come before a piece of code, and "decorative" comments which come right after the code they are associated to.

Changes:
- associator#associate now returns 2 hashes: @commentMap and @decorative_comment
- @commentMap corresponds to the hash which the function used to return, but with a different key to fix the issue of "merged" comments for similar nodes (issue #184). The key is now the node's location instead of the node itself (the value is an array of comments, this did not change).
- The new returned hash @decorative_comment uses the comment as key. The value is a boolean which is true if the comment is "decorative" (false for heading comments).

NB: Most of the time, a decorative comment is on the same line as the node it is associated to. The exception is for trailing comments in a file, which are after the last function of a file. The code in this PR will associate these comments correctly to the last function. This fixes issue #185.

Fixes https://github.com/whitequark/parser/issues/184
Fixes https://github.com/whitequark/parser/issues/185

Example of use:
```code:ruby
  # Parse the source and comments
  ast, comments = Parser::CurrentRuby.new.parse_with_comments(srcFile)
  associator = Parser::Source::Comment::Associator.new(ast, comments)
  @commentMap, @decorative_comment = associator.associate
  ...

  # Returns comments of a node, ready for JavaScript format (a paragraph of heading comments, and a line of decorative comments)
  def getComments(node)
    comments = @commentMap[node.location] # used to be comments = @commentMap[node]
    parag = ""
    deco = ""
    comments.each do |c|
      txt = c.text[1..-1] # get rid of leading "#"
      txt = txt[1..-1] if txt[0]==" " # get rid of first space if any
      if @decorative_comment[c]
        deco << " // #{txt}"
      else
        parag << "// #{txt}\n"
      end
    end
    return parag, deco
  end
```

### DIFF
--- a/lib/parser/source/comment/associator.rb
+++ b/lib/parser/source/comment/associator.rb
@@ -51,6 +51,7 @@ module Parser
         @comments    = comments
 
         @skip_directives = true
+        @map_using_node = true
       end
 
       ##
@@ -80,7 +81,6 @@ module Parser
       #
       def associate
         @mapping     = Hash.new { |h, k| h[k] = [] }
-        @decorative_comment = {}
         @comment_num = -1
         advance_comment
 
@@ -88,7 +88,16 @@ module Parser
 
         process(nil, @ast)
 
-        return @mapping, @decorative_comment
+        return @mapping
+      end
+
+      # #associate is broken for nodes which have the same content.
+      # e.g. 2 identical lines in the code will see their comments "merged".
+      # Using #associate_locations prevents this. The returned hash uses
+      # node.location as a key to retrieve the comments for this node.
+      def associate_locations
+        @map_using_node = false
+        return associate
       end
 
       private
@@ -154,8 +163,8 @@ module Parser
         else
           n = prev_node ? prev_node : node
         end
-        @mapping[n.location] << @current_comment
-        @decorative_comment[@current_comment] = (n == prev_node)
+        key = @map_using_node ? n : n.location
+        @mapping[key] << @current_comment
         advance_comment
       end
 
--- a/lib/parser/source/comment/associator.rb
+++ b/lib/parser/source/comment/associator.rb
@@ -80,74 +80,93 @@ module Parser
       #
       def associate
         @mapping     = Hash.new { |h, k| h[k] = [] }
-        @comment_num = 0
+        @decorative_comment = {}
+        @comment_num = -1
+        advance_comment
 
         advance_through_directives if @skip_directives
 
         process(nil, @ast)
 
-        @mapping
+        return @mapping, @decorative_comment
       end
 
       private
 
-      def process(upper_node, node)
-        if node.type == :begin
-          prev_node, next_node = nil, upper_node
-        else
+      def process(prev_node, node)
+        if node.type != :begin
           while current_comment_between?(prev_node, node)
-            associate_and_advance_comment(node)
+            associate_and_advance_comment(prev_node, node)
           end
-
-          prev_node, next_node = nil, upper_node
-        end
-
-        node.children.each do |child|
-          if child.is_a?(AST::Node) && child.loc && child.loc.expression
-            prev_node, next_node = next_node, child
-
-            process(prev_node, child)
+          if current_comment_decorates?(node)
+            associate_and_advance_comment(node, nil)
           end
         end
-      end
 
-      def current_comment
-        @comments[@comment_num]
+        if node.children.length > 0
+          node.children.each do |child|
+            if child.is_a?(AST::Node) && child.loc && child.loc.expression
+              process(prev_node, child)
+              prev_node = child
+            end
+          end
+          while current_comment_at_end?(node, nil)
+            associate_and_advance_comment(prev_node, nil)
+          end
+        end
       end
 
       def advance_comment
         @comment_num += 1
+        @current_comment = @comments[@comment_num]
       end
 
       def current_comment_between?(prev_node, next_node)
-        return false if current_comment.nil?
+        return false if !@current_comment
+        comment_loc = @current_comment.location.expression
 
-        comment_loc = current_comment.location.expression
-        next_loc    = next_node.location.expression
-
-        if prev_node.nil?
-          comment_loc.end_pos <= next_loc.begin_pos
-        else
-          prev_loc  = prev_node.location.expression
-
-          comment_loc.begin_pos >= prev_loc.end_pos &&
-                comment_loc.end_pos <= next_loc.begin_pos
+        if next_node
+          next_loc = next_node.location.expression
+          return false if comment_loc.end_pos > next_loc.begin_pos
         end
+        if prev_node
+          prev_loc = prev_node.location.expression
+          return false if comment_loc.begin_pos < prev_loc.end_pos
+        end
+        return true
       end
 
-      def associate_and_advance_comment(node)
-        @mapping[node] << current_comment
+      def current_comment_decorates?(prev_node)
+        return false if !@current_comment
+        return @current_comment.location.line == prev_node.location.line
+      end
+
+      def current_comment_at_end?(parent, last_node)
+        return false if !@current_comment
+        comment_loc = @current_comment.location.expression
+        parent_loc = parent.location.expression
+        return comment_loc.end_pos <= parent_loc.end_pos
+      end
+
+      def associate_and_advance_comment(prev_node, node)
+        if prev_node and node
+          n = (@current_comment.location.line == prev_node.location.line) ? prev_node : node
+        else
+          n = prev_node ? prev_node : node
+        end
+        @mapping[n.location] << @current_comment
+        @decorative_comment[@current_comment] = (n == prev_node)
         advance_comment
       end
 
       def advance_through_directives
         # Skip shebang.
-        if current_comment && current_comment.text =~ /^#!/
+        if @current_comment && @current_comment.text =~ /^#!/
           advance_comment
         end
 
         # Skip encoding line.
-        if current_comment && current_comment.text =~ Buffer::ENCODING_RE
+        if @current_comment && @current_comment.text =~ Buffer::ENCODING_RE
           advance_comment
         end
       end
--- a/lib/parser/source/comment/associator.rb
+++ b/lib/parser/source/comment/associator.rb
@@ -51,7 +51,6 @@ module Parser
         @comments    = comments
 
         @skip_directives = true
-        @map_using_node = true
       end
 
       ##
@@ -80,15 +79,8 @@ module Parser
       # @return [Hash(Parser::AST::Node, Array(Parser::Source::Comment))]
       #
       def associate
-        @mapping     = Hash.new { |h, k| h[k] = [] }
-        @comment_num = -1
-        advance_comment
-
-        advance_through_directives if @skip_directives
-
-        process(nil, @ast)
-
-        return @mapping
+        @map_using_node = true
+        associate_comments
       end
 
       # #associate is broken for nodes which have the same content.
@@ -97,10 +89,22 @@ module Parser
       # node.location as a key to retrieve the comments for this node.
       def associate_locations
         @map_using_node = false
-        return associate
+        associate_comments
       end
 
       private
+
+      def associate_comments
+        @mapping     = Hash.new { |h, k| h[k] = [] }
+        @comment_num = -1
+        advance_comment
+
+        advance_through_directives if @skip_directives
+
+        process(nil, @ast)
+
+        @mapping
+      end
 
       def process(prev_node, node)
         if node.type != :begin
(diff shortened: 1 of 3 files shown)


whitequark:
No, unfortunately this is not a solution. The proper solution would:

  * Associate *all* comments with the corresponding nodes, with the "trailing" comments becoming also associated with the preceding node (like you did for the trailing comment in entire file).
  * Return an associative array, i.e. an array of pairs, where each pair consists of the node and the corresponding comment.

kubicle:
1: I might have been unclear : this PR associates all comments. 

2: And what would be the index of the array? Do you really mean an integer index 0..n? This would make the caller do useless efforts to retrieve comments associated with a node. 
I commented on this earlier but did not get a reply.

whitequark:
@kubicle 

1. I've put emphasis on "all" but it probably should've been on "corresponding nodes". I strongly dislike the distinct maps for "normal" and "decorative" comments. It's not a decision that is parser's to make.
2. The index would not have any meaning. Since the node cannot be used as a hash key in our circumstances and that will not be changed (see the original PR for why), it's the only way.

kubicle:
So I *have* been unclear, sorry for that. See my (imperfect) code sample, though, all comments are in the same map ("commentMap").
The other map ("decorativeComment") is just a convenience for the caller : since the comment associator in this PR is already comparing the position of the comments it makes sense to store the result so callers do not need to do it again.

whitequark:
OK. My last comment on that still applies--there is no place for this distinction in parser's API. It is an implementation detail at most.

kubicle:
Well, this is "your baby" so you decide,  of course. 

The current implementation (or the one you proposed) is less interesting for a "client" of parser like me, and I suppose it will be true for all the clients who have to do something with the comments parsed. 

I still cannot exactly understand what bothers you with the idea of "decorative comments". Especially in a language like ruby where comments are on a single line.
Or is it because comments don't classify as important results of the parsing job in your eyes? For any translation or pretty printing task, I think they are just as important. No?

whitequark:
"Decorative comments" are a semantic distinction that is not present at source level. It is not parser's job to opine on whether the comment is "normal" or "decorative" or whatever. Its job is merely to present all comments and the nodes they're closest to.

Besides, it's completely trivial to reproduce the distinction if wanted. `node.loc.expression.begin <=> comment.loc.expression.begin`.

kubicle:
Ah, good, I think we are getting to a better understanding.
I completely agree with you about the job:
"present all comments and the nodes they're closest to."

This is on the "closest" point that I think the current implementation can be improved. We could call them by any other name than "decorative comments", which was probably a wrong choice from me, but I do believe these comments should be associated with the previous node. Systematically associate a comment with the next node makes it wrong. But maybe I am missing something.
When you say it is trivial,  this is if (and only if) you are given both the node and the comment, *then* you can compare their location and see if the comment is on the same line than the node.
*However*, in the current implementation, for a code like below, the comment gets associated to the node of the next function (f2).
```ruby
x = f1(y,z) # call f1 on y & z

def f2(n)
...
```
And, correct me if I am wrong, the algorithm to find that this comment was actually intended for the call of f1 is not trivial; you would have to backtrack the previous nodes - actually doing a similar tree visiting that comment associator has already done.
Do I make more sense to you now?

whitequark:
Sure. In my view a proper implementation would associate a comment not on the beginning of line but at the end of line with the preceding node. Is this not what yours does?

wizcorp-olivier:
Yes this is what it does. So we agree on this part, great!
 
Then the remaining issues are: (let me see if I got it correctly) 

1- You would like the API to return an array of pairs [node,  comment]; my PR is currently returning a hash [key=node-location, value=comment].
If we go your way,  I can add an extra step to recreate the hash I need in my calling code (i.e. not in my PR but the code in my repo that uses Parser). I find this less good than returning my hash because I suppose that all callers will have to recreate this hash too.  Reason is that they will visit the parsing tree finding each node in turn,  and only at this moment (having the node) they will want to find the associated comment. With your array they would need a loop to find the right comment,  while my hash would give it to them directly. Let me know if I am missing something here. 

2- You are against returning the extra array of boolean, one per comment,  with true for a comment associated with a preceding node, and false for one associated with a following node. I used to call the former "decorative" and I agree with you this is not right. I also agree this is easy for callers to compare the comment location with the node location and figure out this way if the comment is on the same line (or *after* the node,  for the case of comments that are after the last statement of a function or the last function of a class,  etc.).
I feel a bit (without 100% certitude)  that other Parser users will have the same need as me (i.e. people for who just need to have the boolean "before-or-after-node" to they can decide how to use it without messing around with the location info)... But again,  this is your call.

If the above is more or less correct,  then would you be OK if I modify 1 & 2 according to what I described? 
Namely,  the API will simply return an array of pairs [node,  comment]

kubicle:
Ah,  posted the above with my work account...  This is still me ;)

whitequark:
Oh! So you suggest using `node.loc` as the key instead of `node`. I agree that this is a better solution than an array.

kubicle:
Very good! So just tell me what you want me to change in the PR and I will do it. Only removing @decorative_comment (and not return it to callers)? Anything else?

whitequark:
* Remove `@decorative_comment`
 * Don't change the interface; name the new function something like `associate_locations` and make sure the old function returns the data in the same format. The mapping would be straightforward.
 * Add tests!

kubicle:
Hello! 
Could you please take a look at the last commit above and tell me if this is what you meant? 
About the tests:
I saw TestSourceCommentAssociator is breaking and could modify it to pass. In short, the "stray" comment would now be associated with previous node,  and the intermediate comment would be with the node before it, I guess.
Thanks

lib/parser/source/comment/associator.rb
```diff
@@ -80,74 +81,101 @@ def initialize(ast, comments)
...
-        @comment_num = 0
+        @comment_num = -1
+        advance_comment
 
         advance_through_directives if @skip_directives
 
         process(nil, @ast)
 
-        @mapping
+        return @mapping
```
whitequark: Just `@mapping`, please

lib/parser/source/comment/associator.rb
```diff
@@ -80,74 +81,101 @@ def initialize(ast, comments)
...
+      end
+
+      # #associate is broken for nodes which have the same content.
+      # e.g. 2 identical lines in the code will see their comments "merged".
+      # Using #associate_locations prevents this. The returned hash uses
+      # node.location as a key to retrieve the comments for this node.
+      def associate_locations
+        @map_using_node = false
```
whitequark: You also need to set `@map_using_node = true` on top of `associate`, otherwise calling `associate` after `associate_locations` will break the API contract.

kubicle:
Oh, and I can add a test for associate_locations of course!

lib/parser/source/comment/associator.rb
```diff
@@ -80,74 +81,101 @@ def initialize(ast, comments)
...
+        return false if !@current_comment
+        comment_loc = @current_comment.location.expression
+        parent_loc = parent.location.expression
+        return comment_loc.end_pos <= parent_loc.end_pos
       end
 
-      def associate_and_advance_comment(node)
-        @mapping[node] << current_comment
+      def associate_and_advance_comment(prev_node, node)
+        if prev_node and node
```
whitequark: Use `&&`

lib/parser/source/comment/associator.rb
```diff
@@ -80,74 +81,101 @@ def initialize(ast, comments)
...
+          n = (@current_comment.location.line == prev_node.location.line) ? prev_node : node
```
whitequark: A more descriptive name for `n`, please

whitequark:
Yes, please

lib/parser/source/comment/associator.rb
```diff
@@ -80,74 +81,101 @@ def initialize(ast, comments)
...
+        return @mapping
+      end
+
+      # #associate is broken for nodes which have the same content.
+      # e.g. 2 identical lines in the code will see their comments "merged".
+      # Using #associate_locations prevents this. The returned hash uses
+      # node.location as a key to retrieve the comments for this node.
+      def associate_locations
+        @map_using_node = false
+        return associate
```
whitequark: Ditto

lib/parser/source/comment/associator.rb
```diff
@@ -80,74 +81,101 @@ def initialize(ast, comments)
...
-
-            process(prev_node, child)
+        if node.children.length > 0
+          node.children.each do |child|
+            if child.is_a?(AST::Node) && child.loc && child.loc.expression
+              process(prev_node, child)
+              prev_node = child
+            end
+          end
+          while current_comment_at_end?(node, nil)
```
whitequark: Can this become an infinite loop in some conditions?

lib/parser/source/comment/associator.rb
```diff
@@ -80,74 +81,101 @@ def initialize(ast, comments)
...
-        next_loc    = next_node.location.expression
+        if next_node
+          next_loc = next_node.location.expression
+          return false if comment_loc.end_pos > next_loc.begin_pos
+        end
+        if prev_node
+          prev_loc = prev_node.location.expression
+          return false if comment_loc.begin_pos < prev_loc.end_pos
+        end
+        return true
```
whitequark: Ditto

lib/parser/source/comment/associator.rb
```diff
@@ -80,74 +81,101 @@ def initialize(ast, comments)
...
+      end
 
-        if prev_node.nil?
-          comment_loc.end_pos <= next_loc.begin_pos
-        else
-          prev_loc  = prev_node.location.expression
+      def current_comment_decorates?(prev_node)
+        return false if !@current_comment
+        return @current_comment.location.line == prev_node.location.line
```
whitequark: Ditto

lib/parser/source/comment/associator.rb
```diff
@@ -80,74 +81,101 @@ def initialize(ast, comments)
...
+      end
 
-          comment_loc.begin_pos >= prev_loc.end_pos &&
-                comment_loc.end_pos <= next_loc.begin_pos
-        end
+      def current_comment_at_end?(parent, last_node)
+        return false if !@current_comment
+        comment_loc = @current_comment.location.expression
+        parent_loc = parent.location.expression
+        return comment_loc.end_pos <= parent_loc.end_pos
```
whitequark: Ditto

kubicle:
Working on your changes... 
It looks like I can simplify a bit and get rid of the "current_comment_decorates?" method.

whitequark:
Sure

kubicle:
@whitequark let me know how you like this one~

lib/parser/source/comment/associator.rb
```diff
@@ -79,75 +79,115 @@ def initialize(ast, comments)
...
+        processNode(node)
+        
+        if node.children.length > 0
+          node.children.each do |child|
+            next unless child.is_a?(AST::Node) && child.loc && child.loc.expression
+            visit(child)
           end
-
-          prev_node, next_node = nil, upper_node
+          processTrailingComments(node)
```
whitequark: `process_trailing_comments`

lib/parser/source/comment/associator.rb
```diff
@@ -79,75 +79,115 @@ def initialize(ast, comments)
...
         end
+      end
 
-        node.children.each do |child|
-          if child.is_a?(AST::Node) && child.loc && child.loc.expression
-            prev_node, next_node = next_node, child
-
-            process(prev_node, child)
-          end
+      def processNode(node)
```
whitequark: `process_node`

test/test_source_comment_associator.rb
```diff
@@ -2,18 +2,26 @@
...
+  def associate(code, map_using_location=false)
     parser = Parser::Ruby18.new
 
     buffer = Parser::Source::Buffer.new('(comments)')
     buffer.source = code
 
     ast, comments = parser.parse_with_comments(buffer)
-    associations  = Parser::Source::Comment.associate(ast, comments)
+    if map_using_location
+      associations  = Parser::Source::Comment::Associator.new(ast, comments).associate_locations
```
whitequark: I think this is a good sign that `Parser::Source::Comment.associate` needs a keyword argument `using_locations: false`.

test/test_source_comment_associator.rb
```diff
@@ -91,7 +187,7 @@ def foo
 end
     END
 
-    assert_equal 0, associations.size
+    assert_equal 1, associations.size
```
whitequark: Maybe add a test which verifies that the stray comment is associated to the right node.

lib/parser/source/comment/associator.rb
```diff
@@ -79,75 +79,115 @@ def initialize(ast, comments)
...
+          associate_and_advance_comment(@prev_node, node)
         end
+        @prev_node = node
       end
 
-      def current_comment
-        @comments[@comment_num]
+      def processTrailingComments(parent)
+        while current_comment_decorates?(@prev_node)
+            associate_and_advance_comment(@prev_node, nil)
```
whitequark: Indentation.

whitequark:
Sans a few minor issues, this looks very good.

kubicle:
Hey hey,  what about now?

lib/parser/source/comment/associator.rb
```diff
@@ -76,78 +76,111 @@ def initialize(ast, comments)
...
+      # NB: #associate is broken for nodes which have the same content.
+      # e.g. 2 identical lines in the code will see their comments "merged".
+      # Using map_using_locations=true prevents this. The returned hash uses
+      # node.location as a key to retrieve the comments for this node.
+      #
       # @return [Hash(Parser::AST::Node, Array(Parser::Source::Comment))]
+      #      or [Hash(Parser::Source::Map, Array(Parser::Source::Comment))]
       #
-      def associate
+      def associate(map_using_locations = false)
```
whitequark: This is not really what I meant... methods returning different types depending on an argument are horribly designed API.

Sigh. I'll just fix it myself

whitequark:
Please see the final API version on master.

Thanks for your work!

PR #195, opened 2015-05-18
Fix end comment association

### LINKED ISSUE #194: Improve trailing comment association.
### Reasoning

Consider this code:

``` ruby
class C
  def foo # def C#foo
    puts 'foo' # puts 'foo'
  end # def C#foo end

  def bar # def C#bar
    puts 'bar' # puts 'bar'
  end # def C#bar end
  
  #totally unrelated comment
end
```

It associates comments in the following way:

```
(class
  (const nil :C) nil
  (begin
    (def :foo
      (args)
      (send nil :puts
        (str "foo")))
    (def :bar
      (args)
      (send nil :puts
        (str "bar")))))
>>> 1: 0:# class C before
(def :foo
  (args)
  (send nil :puts
    (str "foo")))
>>> 3:10:# def C#foo
(str "foo")
>>> 4:15:# puts 'foo'
(def :bar
  (args)
  (send nil :puts
    (str "bar")))
>>> 5: 6:# def C#foo end
>>> 7:10:# def C#bar
(str "bar")
>>> 8:15:# puts 'bar'
(begin
  (def :foo
    (args)
    (send nil :puts
      (str "foo")))
  (def :bar
    (args)
    (send nil :puts
      (str "bar"))))
>>> 9: 6:# def C#bar end
>>>11: 2:#totally unrelated comment
```

See bigger example: [source](https://gist.github.com/RGBD/0fc6e0377fdefbbf721b#file-test-rb) and [output](https://gist.github.com/RGBD/0fc6e0377fdefbbf721b#file-default-log).
Also, it partly hides bug in [rubocop](https://github.com/bbatsov/rubocop).

### My opinion

I think, this is a bit counter-intuitive. Also now
Parser::Source::Comment::Associator#accociate contains a workaround for
trailing (akasparse in tests) comments, which might, or might not be associated
with enclosing or preceding block (see gist).

### What can be done

I propose another principle of comment association.

#### Features

- It is unambiguous
- It is scalable, association for the whole tree is exactly the same as
  united association for every node (on sparse comments later)
- It has the same behaviour for preceding comments, so that any documentation
  generator using this gem will keep working (correct me if I'm wrong).
- It is configurable, you can deside, whether you need these or that kind of
  comments

#### Definitions

Suppose previous node has been processed.

- Comment is **preceding**, if it starts before the beginning of current node.
  **preceding** comments are supposed to be documenting.
- Comment is **sparse**, if it is inside the body of current node, and after
  the last child node. **sparse** comments' purpose is ambiguous, they could be
  ignored.
- Comment is **decorating**, if it starts at the current node ending line.
	**decorating** comments comment expression before them, or block endings.

### Summary

I hope you will consider my proposal and check out my implementation in (#195)

RGBD:
Fixes #194.
Only a sketch. Please, review and respond.

### DIFF
--- a/lib/parser/source/map.rb
+++ b/lib/parser/source/map.rb
@@ -98,6 +98,8 @@ module Parser
         @expression.line
       end
 
+      alias_method :first_line, :line
+
       ##
       # A shortcut for `self.expression.column`.
       # @return [Integer]
@@ -107,19 +109,19 @@ module Parser
       end
 
       ##
-      # A shortcut for `self.expression.line`.
+      # A shortcut for `self.expression.last_line`.
       # @return [Integer]
       #
-      def end_line
-        @expression.end_line
+      def last_line
+        @expression.last_line
       end
 
       ##
-      # A shortcut for `self.expression.column`.
+      # A shortcut for `self.expression.last_column`.
       # @return [Integer]
       #
-      def end_column
-        @expression.end_column
+      def last_column
+        @expression.last_column
       end
 
       ##
--- a/lib/parser/source/map.rb
+++ b/lib/parser/source/map.rb
@@ -107,6 +107,22 @@ module Parser
       end
 
       ##
+      # A shortcut for `self.expression.line`.
+      # @return [Integer]
+      #
+      def end_line
+        @expression.end_line
+      end
+
+      ##
+      # A shortcut for `self.expression.column`.
+      # @return [Integer]
+      #
+      def end_column
+        @expression.end_column
+      end
+
+      ##
       # @api private
       #
       def with_expression(expression_l)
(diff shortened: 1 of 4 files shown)


lib/parser/source/map.rb
```diff
@@ -107,6 +107,22 @@ def column
       end
 
       ##
+      # A shortcut for `self.expression.line`.
```
whitequark: Doc is inconsistent.

lib/parser/source/map.rb
```diff
@@ -107,6 +107,22 @@ def column
...
+      # @return [Integer]
+      #
+      def end_line
```
whitequark: I would prefer `last_line`, and also a `first_line` as an alias to `line`.

whitequark:
Rebased and merged

PR #199, opened 2015-06-25
Fix the warning for Ruby 2.1 (latest is 2.1.6)

rousisk:
Hi,

It seems like https://github.com/whitequark/parser/commit/e93b9080f58fe57c217b0c3f16cb61ed1ad45d73 has incorrectly bumped the latest 2.1 Ruby check to 2.1.7. 

Since I can't find any reference to such minor I reverted it back to 2.1.6.

### DIFF
--- a/lib/parser/current.rb
+++ b/lib/parser/current.rb
@@ -33,8 +33,8 @@ module Parser
     CurrentRuby = Ruby20
 
   when /^2\.1\./
-    if RUBY_VERSION != '2.1.7'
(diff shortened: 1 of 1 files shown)


whitequark:
Why are you opening a bug against an unreleased parser version?

rousisk:
I somehow ended up with 2.3.0.pre.2 locally, ran into this warning and assumed that there was a typo in the commit. Completely misunderstood the purpose of the commit - sorry :)

PR #201, opened 2015-08-07
Reduce object count

rrosenblum:
I was running some memory profiles on RuboCop and noticed that the highest memory allocation was coming from `parser-2.2.2.5/lib/parser/source/buffer.rb`. This change reduces the memory allocation by about 30%.


marshall-lee:
Cool, but why not even `$/`?

marshall-lee:
Ah, okay, I got it.

PR #211, opened 2015-10-30
Remove unneeded indirection in Parser::AST::Processor

alexdowad:
I'm not sure if there is a good reason for these methods to exist, but I sure can't see any. Removing the unneeded indirection should increase performance a little bit.

I've checked RuboCop and it doesn't use the removed methods.

### DIFF
--- a/lib/parser/ast/processor.rb
+++ b/lib/parser/ast/processor.rb
@@ -24,6 +24,7 @@ module Parser
         node
       end
 
+      # @private
       def process_variable_node(node)
         on_var(node)
       end
@@ -43,6 +44,7 @@ module Parser
         ])
       end
 
+      # @private
       def process_var_asgn_node(node)
         on_vasgn(node)
       end
@@ -92,6 +94,7 @@ module Parser
         ])
       end
 
+      # @private
       def process_argument_node(node)
         on_argument(node)
       end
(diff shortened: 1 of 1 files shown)


whitequark:
The pattern feels odd, but there might have been some reason for it--I can't remember. Something to do with the behavior of `alias`?

Either way, it's now a part of our public API... it's even marked `@api public`.

whitequark:
Have you quantified the speedup?

alexdowad:
On first read, the code looked so obviously redundant that I didn't bother at first. But since you mentioned `alias`, I know why you did this.

When you `alias` (for example) `a` as `b`, and later override `a`, `b` will not "change". Aliasing *copies* a method's implementation to a new entry in the class' method table. Changing the copied method's implementation does nothing to the copy.

By redirecting the call through an additional method, you can override `on_vasgn`, and `on_lvasgn`, `on_ivasgn`, etc. will all "change".

It's unfortunate that you didn't mark `process_variable_node` et al as "private", though. Such is life.

bbatsov:
It should be check whether this is actually used in the wild first. Just the fact that something is public doesn't mean that clients are actually using it.

whitequark:
Yep, that was the reason. In light of this I don't even feel they should be removed.

alexdowad:
@whitequark Would you like them to be marked as `#@private`? (If anybody is using them, this won't break their code.)

whitequark:
Why? They provide useful behavior.

whitequark:
Oh, I see what you're suggesting. The fact that there is indirection is indeed an implementation detail, and the `process_*` methods should be marked `@private`.

alexdowad:
> They provide useful behavior.

They are useful as an implementation detail, but not to users of your library.

Due to the behavior of `alias`, which I explained above, overriding `process_variable_assignment` (for example) will do all of... nothing.

You can *call* them, but why? Just call `on_vasgn` directly.

whitequark:
Even better idea: at the very end, override them to print a `warning` that they're deprecated as a public API and will be removed in 2.4 or 3.0.

alexdowad:
Will do.

alexdowad:
Updated PR.

PR #220, opened 2015-11-15
* lexer.rl: reduce String literal object allocations

walf443:
This patch reduce String literal object allocations.

String object expanded many times on Lexer#advance.

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -89,7 +89,7 @@ class Parser::Lexer
 
   BLANK_STRING = ''.freeze
   ESCAPED_NEXT_LINE = "\\\n".freeze
-  REGEXP_META_CHARACTERS = "\\$()*+.<>?[]^{|}".freeze
+  REGEXP_META_CHARACTERS = Regexp.union(*"\\$()*+.<>?[]^{|}".chars).freeze
   UNDERSCORE_STRING = '_'.freeze
 
   attr_reader   :source_buffer
@@ -875,7 +875,7 @@ class Parser::Lexer
         # or closing delimiter, it is an escape sequence for that
         # particular character. Write it without the backslash.
 
-        if literal.regexp? && REGEXP_META_CHARACTERS.include?(escaped_char)
+        if literal.regexp? && REGEXP_META_CHARACTERS.match(escaped_char)
           # Regular expressions should include escaped delimiters in their
           # escaped form, except when the escaped character is
           # a closing delimiter but not a regexp metacharacter.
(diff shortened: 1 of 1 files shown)


lib/parser/lexer.rl
```diff
@@ -87,6 +87,11 @@ class Parser::Lexer
     'v' => "\v", '\\' => "\\"
   }
 
+  BLANK_STRING = ''.freeze
+  ESCAPED_NEXT_LINE = "\\\n".freeze
+  REGEXP_META_CHARACTERS = "\\$()*+.<>?[]^{|}".freeze
```
whitequark: Can you look at replacing `REGEXP_META_CHARACTERS.include?` with a regular expression match? It might be faster to do that.

mbj: `REGEXP_META_CHARACTERS = Regexp.union(*"your_chars_string".chars)` should do the trick.

whitequark: Yes, I mean, I'm asking @walf443 to measure the impact.

walf443:
I measure it with @ko1 's allocation_tracer gem.

```
diff --git bin/ruby-parse bin/ruby-parse
index e55c2f4..7f34f91 100755
--- bin/ruby-parse
+++ bin/ruby-parse
@@ -3,4 +3,14 @@
 $LOAD_PATH.unshift(File.expand_path('../../lib', __FILE__))
 require 'parser/runner/ruby_parse'

+require 'pp'
+require 'allocation_tracer'
+
+trace_result = ObjectSpace::AllocationTracer.trace{
 Parser::Runner::RubyParse.go(ARGV)
+}.select {|key| key[0].include?('parser/lexer.rb') }
+.each_with_object({}) {|(key,val),j| j[key] = val[0] }
+trace_result.sort_by {|key,val| key[1] }.each do |key,val|
+  puts "#{key[0]}:#{key[1]} :#{val}"
+end
+p trace_result.inject(0) {|sum,(key,val)| sum += val; sum }
```

```
time ./bin/ruby-parse test/**/*.rb > before.txt
time ./bin/ruby-parse test/**/*.rb > after.txt
```

https://gist.github.com/4c6cd382120a210dc174

before is 601885 and after is 601254.
It's small impact than i expected.

whitequark:
@walf443 I meant measure the impact of the change described in [this comment](https://github.com/whitequark/parser/pull/220#discussion_r44873341).

walf443:
I measure it followings:

```
time ruby-prof -f after1 ./bin/ruby-parse test/**/*.rb
```

```
after1 after2
| 0.00      0.000     0.000     0.000     0.000        1   String#include? | 0.00      0.000     0.000     0.000     0.000       25   Regexp#match |
```

Though it's rare case to match that case. it's no need to chage it.
Literal string object is allocated while defining methods.

walf443:
> Literal string object is allocated while defining methods.

It's wrong. it will be changed in MRI 2.3.0.

PR #230, opened 2015-12-13
Enhance Range and Rewriter

alexdowad:
I just ran into a couple problems when working on some RuboCop code which prompted me to make these enhancements.

It started with a RC cop which deletes some unneeded things. In some circumstances, it may try to autocorrect overlapping ranges of code. This causes `parser` to vomit up a `ClobberingError`.

The thing is that there is no conflict at all between the overlapping edits. They are just all deleting code anyways. So this PR amends `Rewriter` so that it is smart enough to notice when there is no conflict between overlapping deletions *or* replacements, and merge them when possible.

Another thing: some of those deletions span newlines. So at first, I couldn't even see that the problem was a `ClobberingError`, because `Diagnostic#render` calls `column_range`, and that fails if the `Range` spans more than one line.

To avoid such annoyances, make `Diagnostic` aware of the fact that ranges may span more than one line, and specifically make it only attempt to display (and highlight) the first line in the erroneous range.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,7 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in parser.gemspec
 gemspec
+
+# Workaround for bug in Bundler on JRuby
+# See https://github.com/bundler/bundler/issues/4157
+gem 'ast', '>= 1.1', '< 3.0'
--- a/Rakefile
+++ b/Rakefile
@@ -30,7 +30,7 @@ GENERATED_FILES = %w(lib/parser/lexer.rb
 
 CLEAN.include(GENERATED_FILES)
 
-desc 'Generate the Ragel lexer and Bison parser.'
+desc 'Generate the Ragel lexer and Racc parser.'
 task :generate => GENERATED_FILES do
   Rake::Task[:ragel_check].invoke
   GENERATED_FILES.each do |filename|
@@ -45,7 +45,7 @@ end
 
 task :regenerate => [:clean, :generate]
 
-desc 'Generate the Ragel lexer and Bison parser in release mode.'
+desc 'Generate the Ragel lexer and Racc parser in release mode.'
 task :generate_release => [:clean_env, :regenerate]
 
 task :clean_env do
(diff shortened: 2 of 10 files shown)


alexdowad:
Very puzzling failure on JRuby: `no such file to load -- ast`.

alexdowad:
Something is wrong with Bundler on JRuby. `bundle install` doesn't install `ast`.

alexdowad:
Hmm. I traced through `bundler` on JRuby, found what looks like the reason it's failing, and tried patching it. With my patched `bundler`, things are working. BUT! That doesn't explain why it *does* work on MRI.

alexdowad:
I've found the problem.

whitequark:
Other than the exception message nit, I have one problem with the PR: it means diagnostics can now lie about ranges. This is unacceptable (yes, less acceptable than crashing). An ideal solution would be to print the first and last line in the range, with correct highlighting and some indication that the middle of the span is skipped, e.g.:

```
foo.rb:1:2-3:5: error: multi-span range
foo.rb:1: def foo
foo.rb:1:     ~~~...
foo.rb:3: end
foo.rb:3: ~~~
```

alexdowad:
If you put `spec.platforms = ['ruby', 'jruby']` in the gemspec (just need to confirm whether those should be strings or symbols), that will make the specs run on JRuby. But it would be nice if this problem can be fixed right in Bundler, so others don't run into it.

alexdowad:
> Other than the exception message nit, I have one problem with the PR: it means diagnostics can now lie about ranges. This is unacceptable (yes, less acceptable than crashing).

Very well. That makes a lot of sense.

bbatsov:
I like the idea of showing more context about the range.

alexdowad:
Your specs are failing on JRuby because of [this](https://github.com/bundler/bundler/issues/4157).

alexdowad:
OK, I have fixed things up according to @whitequark's (helpful) feedback.

...And slipped a couple more changes in.

alexdowad:
What @whitequark wants, @whitequark gets. Updated.

alexdowad:
@whitequark, any chance of releasing an updated gem?

whitequark:
2.3.0.pre.5

alexdowad:
Thanks @whitequark.

PR #232, opened 2015-12-17
* bump warnings to 2.1.8 and 2.2.4

spickermann:
Ruby versions 2.1.8 and 2.2.4 were released yesterday:

  * https://www.ruby-lang.org/en/news/2015/12/16/ruby-2-1-8-released/
  * https://www.ruby-lang.org/en/news/2015/12/16/ruby-2-2-4-released/

### DIFF
--- a/lib/parser/current.rb
+++ b/lib/parser/current.rb
@@ -34,16 +34,16 @@ module Parser
     CurrentRuby = Ruby20
 
   when /^2\.1\./
-    if RUBY_VERSION != '2.1.7'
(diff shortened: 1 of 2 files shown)


whitequark:
Since we do in fact have a syntax change in 2.2.4, I will close this and adjust everything when the change is reflected in Parser.

jdrnetworking:
@whitequark would you be open a PR to address the 2.1.8 warning only?

whitequark:
@jdrnetworking https://github.com/whitequark/parser/blob/master/lib/parser/current.rb#L39-L55

PR #235, opened 2015-12-29
update .travis.yml

walf443:
added ruby 2.3.0.

Unfortunately removing patch version does  not seem to  worked well.
https://travis-ci.org/whitequark/parser/jobs/98856946#L111


whitequark:
I don't understand--the build you're referring to is green

walf443:
That link specify rvm 2.2. But It does not run with 2.2.3(or 2.2.4). It run ruby 2.2.0.
So It's better to use head version of patch unfortunately.

whitequark:
Ah true

PR #236, opened 2016-01-01
Add Source::Buffer#last_line

alexdowad:
Buffer already has `#first_line`. `#last_line` is symmetrical with
`#first_line`, and will be generally useful. (Specifically, it will be helpful
with some work I am doing on RuboCop.)


alexdowad:
Thanks @whitequark.

Issue #228, opened 2015-12-08
<<~SQUIGGLY_HEREDOC in Ruby MRI 2.3.x


whitequark:
See upstream commit https://github.com/ruby/ruby/commit/9a28a29b870b5f45d370bc8f16c431b435f0bbb3 and Ruby bug https://bugs.ruby-lang.org/issues/9098

mbj:
Will this also include:

`<<~'SQUIGGLY_HEREDOC'` that disables interpolation?

mbj:
Why, I'm asking. I do not see that one covered by the upstream tests, but it could work by "accident" (and may be buggy).

whitequark:
lol `x<<~FOO` is already valid syntax actually. https://bugs.ruby-lang.org/issues/9098#note-24

This is such a clusterfuck.

whitequark:
@mbj Yes, `<<~'X'` would be supported.

whitequark:
I'm withholding implementation of this in case it gets reverted in MRI.

chastell:
(FWIW, squiggly heredoc was kept in the recently-released Ruby 2.3.0.)

bbatsov:
Hmm, I didn't see it in the release notes. Are you sure?

backus:
@bbatsov https://github.com/ruby/ruby/blob/v2_3_0/NEWS#L53-L57

whitequark:
Ugh.

bbatsov:
Hmm, I wonder why they omitted such a big change here - https://www.ruby-lang.org/en/news/2015/12/25/ruby-2-3-0-released/

Anyways, good to know!

troessner:
>> Hmm, I wonder why they omitted such a big change here - https://www.ruby-lang.org/en/news/2015/12/25/ruby-2-3-0-released/

Human error?

bbatsov:
Likely so.

bbatsov:
@whitequark is the this the final thing that has to be added before 2.3 final is released?

whitequark:
Yes.

whitequark:
*sob*

@mbj, here's the answer to your question: https://github.com/ruby/ruby/commit/b1796737cf097ff7dc68b12ebb6d86a9528c95be.
This is going to be a compatibility nightmare, wouldn't it?

lucascaton:
Hi there. When will 2.3 final be released? Thanks! :smile:

whitequark:
[...]

PR #237, opened 2016-01-07
Add Source::Buffer#lines

alexdowad:
@whitequark Please have a look at this and see if you want it. Commit log message:

Almost any application which processes Ruby source code will want to access
specific lines. `Buffer#source_line` is provided for that purpose; however,
sometimes one wants an `Enumerable` object, which `#map`, `#select`, and so
on can be called on.

`Parser::Source::Buffer` already maintains such an array, but it wasn't exposed.
This caused RuboCop to maintain its own array of lines for each `Buffer`,
at a not-insignificant performance cost.

Exposing the `@lines` array to client code means that it and its contents must
all be frozen to prevent unwanted mutation. This means that accessing source
lines using `#lines` is more efficient than `#source_line`, since there is no
extra copying.

### DIFF
--- a/lib/parser/source/buffer.rb
+++ b/lib/parser/source/buffer.rb
@@ -191,6 +191,26 @@ module Parser
       end
 
       ##
+      # Return an `Array` of source code lines.
+      #
+      # @return [Array<String>]
+      #
+      def source_lines
+        @lines ||= begin
+          # If a file ends with a newline, the EOF token will appear
+          # to be one line further than the end of file.
+          lines = @source.lines.to_a.push('')
+
+          lines.each do |line|
+            line.chomp!(NEW_LINE)
+            line.freeze
+          end
+
+          lines.freeze
+        end
+      end
+
+      ##
       # Extract line `lineno` from source, taking `first_line` into account.
       #
       # @param  [Integer] lineno
@@ -198,16 +218,7 @@ module Parser
       # @raise  [IndexError] if `lineno` is out of bounds
       #
       def source_line(lineno)
-        unless @lines
-          @lines = @source.lines.to_a
-          @lines.each { |line| line.chomp!(NEW_LINE) }
-
-          # If a file ends with a newline, the EOF token will appear
-          # to be one line further than the end of file.
-          @lines << ""
-        end
-
-        @lines.fetch(lineno - @first_line).dup
+        source_lines.fetch(lineno - @first_line).dup
       end
 
       ##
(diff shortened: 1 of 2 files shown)


whitequark:
What about returning an enumerator instead?

alexdowad:
> What about returning an enumerator instead?

That could work, though some clients will end up converting the `Enumerable` to an `Array` using `#to_a`, which takes us back to making unnecessary copies again.

To give some more background, this PR originated from some profiling done on RuboCop. I identified `RuboCop::ProcessedSource#lines` as a slight (not huge, but noticeable) performance drag:

```ruby
    # Returns the source lines, line break characters removed, excluding a
    # possible __END__ and everything that comes after.
    def lines
      @lines ||= begin
        all_lines = raw_source.lines.map(&:chomp)
        last_token_line = tokens.any? ? tokens.last.pos.line : all_lines.size
        result = []
        all_lines.each_with_index do |line, ix|
          break if ix >= last_token_line && line == '__END__'
          result << line
        end
        result
      end
    end
```

...which seems redundant, as `Parser::Source::Buffer` already has a comparable array of lines, which we also use via `Source::Buffer#source_line`.

Another question, if you like this PR: do you think `#source_lines` would be better than `#lines`?

whitequark:
Returning an `Enumerable` has little difference for callers who want an `Array` but a lot of difference for callers who want an `Enumerable`, so we should return that instead. And call it `source_lines`, yes.

alexdowad:
> Returning an Enumerable has little difference for callers who want an Array but a lot of difference for callers who want an Enumerable, so we should return that instead.

Well, an `Array` is also an `Enumerable`. So we are returning an `Enumerable` here.

whitequark:
Er, sorry, `Enumerator`. Don't allocate anything upfront.

alexdowad:
> Er, sorry, Enumerator. Don't allocate anything upfront.

The trouble with that is that this patch is largely about performance; about allowing clients to share the data structure which `Parser` already maintains internally, rather than building their own copy.

When processing thousands of files, some of which may be large, not building 2 copies of the same structure is a measurable win. Returning an `Enumerator` here, when we have something more, erases that win. Because the user will be forced to convert it to an `Array`, which results in a copy.

whitequark:
Ah yes, I now see: I misread the patch and misunderstood what you suggest. LGTM then, after name change.

alexdowad:
Fixed up the PR.

Can you confirm it is OK if `@lines` only has a trailing blank string *when* the source ended with a newline? It seems to make a lot more sense to me.

whitequark:
I'm quite certain there's a test for that, so if it passes, all is OK.

alexdowad:
Thanks @whitequark.

And the next question always is... when is the next incremental release?

whitequark:
Not until I fix #228.

PR #239, opened 2016-01-08
Simplify lexer code which unnecessarily uses emit_table

alexdowad:
In a number of cases, a table lookup is used to find the token to emit, although
there is only one possible token which could be emitted from that point. It
makes the code clearer when the reader can see the token which will be emitted,
rather than having to go and consult the table.

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -1763,7 +1763,7 @@ class Parser::Lexer
           @lambda_stack.pop
           emit(:tLAMBEG)
         else
-          emit_table(PUNCTUATION_BEGIN)
+          emit(:tLBRACE)
         end
         fbreak;
       };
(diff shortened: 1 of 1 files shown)


alexdowad:
This "coveralls" is a pain. I have just exchanged line for line, and the lines which are executed are completely unchanged. Maybe the issue is that some entries in `PUNCTUATION` are unused now?

By the way, would you like unneeded entries to be trimmed from `PUNCTUATION` and `PUNCTUATION_BEGIN`?

whitequark:
Don't mind coveralls, it's mostly there to ensure that the grammar coverage can be seen... it should be at 100% ideally, but error bubbling is not handled at the moment.

whitequark:
`PUNCTUATION`/`PUNCTUATION_BEGIN`/etc are there for reference as much as their use in code, so they stay.

whitequark:
Ah btw, I think coveralls will sometimes appear to reject your changes while not all Travis bots have finished because different Ruby implementations report coverage in different ways.

PR #241, opened 2016-01-11
Simplify simplify

alexdowad:


lib/parser/lexer.rl
```diff
@@ -1769,16 +1769,21 @@ class Parser::Lexer
...
+
       # a()
       e_lparen
-      => { emit_table(PUNCTUATION_BEGIN)
+      => { emit(:tLPAREN)
            fbreak; };
 
-      # a(+b)
-      punctuation_begin
-      => { emit_table(PUNCTUATION_BEGIN)
```
whitequark: What's the point of this?

whitequark:
OK for `emit_table` change, but you'll have to back inlining of random lexer rules with benchmarks.

lib/parser/lexer.rl
alexdowad: A reader (like me) can see right away which constructs will be caught by this rule, and which tokens will be emitted. You don't have to jump back and forth between looking at the code here and looking at the punctuation tables. Also, `PUNCTUATION_BEGIN` is a big table, but only a few entries are relevant here.

But if you don't agree, that is fine.

whitequark: The rule names and behavior are made for benefit of someone who is already familiar with `parse.y`, since the most changes, and most important changes, stem from changes in upstream `parse.y`.

alexdowad: In other words... "you're outta luck, bub".

alexdowad: Would you still like the `emit_table` tweak?

whitequark: Sure.

alexdowad:
Updated, please have a look.

lib/parser/lexer.rl
```diff
@@ -1983,9 +1986,7 @@ class Parser::Lexer
         if version?(18)
           emit(:tIDENTIFIER)
 
-          if !@static_env.nil? && @static_env.declared?(tok)
-            fnext expr_end;
-          else
```
whitequark: Huh??

whitequark: Ah yes, we're in expr_end already.

PR #242, opened 2016-01-12
More tweaking on the lexer

alexdowad:
Please have a look and see if you want any of these changes.

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -196,8 +196,7 @@ class Parser::Lexer
         @source_pts = @source.unpack('C*')
       end
 
-      if (@source_pts.size > 1_000_000 && @has_encode) ||
-         @force_utf32
+      if @has_encode && (@source_pts.size > 1_000_000 || @force_utf32)
         # A heuristic: if the buffer is larger than 1M, then
         # store it in UTF-32 and convert the tokens as they're
         # going out. If it's smaller, the conversion overhead
@@ -210,7 +209,7 @@ class Parser::Lexer
         #
         # Patches accepted.
         @source = @source.encode(Encoding::UTF_32LE)
-        @need_encode = @has_encode && @encoding != Encoding::UTF_32LE
+        @need_encode = @encoding != Encoding::UTF_32LE
       end
 
       if @source_pts[0] == 0xfeff
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -186,11 +186,11 @@ class Parser::Lexer
       @source = @source_buffer.source
       @need_encode = false
 
-      if defined?(Encoding)
+      if @has_encode
         @encoding   = @source.encoding
       end
 
-      if defined?(Encoding) && @source.encoding == Encoding::UTF_8
+      if @has_encode && @source.encoding == Encoding::UTF_8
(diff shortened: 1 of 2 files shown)


whitequark:
> Converting source to UTF-32LE doesn't gain any performance
No, it does. You simply did not use the test corpus on which it is visible. Specifically, it improves parsing speed several times over on very large files with very few tokens. There is even a comment in the lexer saying this.

> Remove redundant calls to String#chr
You broke tests.

> Ruby does not accept ?\u{1 2}, we don't either
I'll need to compare this with what parse.y does.

> pe, etc. index into @source_pts, not @source
OK

> Avoid unnecessarily allocating strings
OK

> Use 'tok' instead of '@source[@ts...@te]'
OK

alexdowad:
> No, it does. You simply did not use the test corpus on which it is visible. Specifically, it improves parsing speed several times over on very large files with very few tokens.

Do such files exist "in the wild"? I'm trying to think of an example...

whitequark:
Sure, I discovered those while running parser on the entire rubygems corpus.

alexdowad:
> Sure, I discovered those while running parser on the entire rubygems corpus.

OK, I will look at that.

alexdowad:
> Sure, I discovered those while running parser on the entire rubygems corpus.

Wait a minute. You're not referring to Rubygems itself, but to the source of various gems, right?

whitequark:
Yes.

alexdowad:
> I'll need to compare this with what parse.y does.

Here's what parse.y does:

It has a function called `parser_tokadd_utf8` which is used to lex `\u` escapes. That function takes a `string_literal` argument which tells it... whether we are in a string literal. When lexing `?\u...`, that argument is false.

When the `string_literal` argument is false, `parser_tokadd_utf8` doesn't attempt to lex more than one codepoint inside `\u{...}`. It will just consume the first codepoint, optionally consume a closing '}', then return. Then the lexer continues on its merry way, sees the next codepoint, and thinks it is just an ordinary int literal. Great, an int literal! OK.

If the parser doesn't choke on the int literal, the lexer will keep marching on, not knowing that it is still inside `\u{...}`, and see `}`. Depending on the context, this may be lexed in a couple of different ways, but it will almost invariably cause the parser to choke.

alexdowad:
Removed the UTF-32LE encoding-related commit; added a couple others which merely civilize the existing code, rather than nuking it.

alexdowad:
I've found that `parser` does not handle `?\u0001` and `?\u{100}` correctly. I'll try to fix those first, and then `?\u{1 2}`.

alexdowad:
> I've found that parser does not handle ?\u0001 and ?\u{100} correctly.

False alarm, I just needed to use `setup_lexer 19` in the tests.

alexdowad:
Tried a different approach for `?\u{...}` with multiple codepoints, please have a look.

whitequark:
The 1.8 build is broken.

alexdowad:
> The 1.8 build is broken.

This is interesting. What do you do when people run the 1.9 parser on 1.8? Is that allowed?

I can see that you have guarded your `\u` escape tests with `if RUBY_VERSION >= '1.9'`.

whitequark:
No, it's not allowed.

alexdowad:
And we're green!

PR #243, opened 2016-01-14
2.3: optional superclass rule

cremno:
See https://github.com/ruby/ruby/commit/d311b7d0eb1e973c8d143f03e66ab72588f57a2a.

### DIFF
--- a/lib/parser/ruby23.y
+++ b/lib/parser/ruby23.y
@@ -1942,11 +1942,7 @@ keyword_variable: kNIL
                       result = @builder.back_ref(val[0])
                     }
 
-      superclass: term
-                    {
-                      result = nil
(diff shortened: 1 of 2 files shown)


whitequark:
Huh, can't believe I missed that.

bbatsov:
So, now you can write code like `class C end`? I wonder how no one noticed this.

whitequark:
I'm especially surprised, given that I'm subscribed to parse.y via RSS.

whitequark:
@cremno Thank you for this high quality PR. (I think you might be the second person who actually read CONTRIBUTING.md.)

whitequark:
Pushed 2.3.0.1

PR #244, opened 2016-01-14
Readme tweaks

bbatsov:
A few small readme tweaks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ $ gem install parser
 
 Parse a chunk of code:
 
-```
+```ruby
 require 'parser/current'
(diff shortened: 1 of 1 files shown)


whitequark:
> Use only fenced code blocks in the README

This breaks rdoc.info, which is why it wasn't done in the first place. rdoc.info doesn't have <code>```</code>, only `~~~`.

bbatsov:
> This breaks rdoc.info, which is why it wasn't done in the first place. rdoc.info doesn't have ```, only ~~~.

Hmm, but there were already such blocks in the README.

whitequark:
Looks like it learned how to parse those.

PR #245, opened 2016-01-15
Tweaking tweaking

alexdowad:


whitequark:
Thanks! I really value this work--most of the code you're fixing is the result of iterative approximation of MRI behavior and it really should be refactored, yet at the same time I'm so used to its quirks that I have not noticed that at all.

alexdowad:
And thank you!

troessner:
And thank you both :laughing: 
(Without `parser` `Reek` would be significantly harder to maintain since the alternatives are....dire)

whitequark:
@troessner Did you port Reek to Parser sometime? I must have missed that

troessner:
Yes, we switched 1.25 years ago.

bbatsov:
Nice! Parser hegemony! :-)

PR #246, opened 2016-01-15
] is a bracket, not a parenthesis

alexdowad:


whitequark:
Let me add you as a collaborator so you could commit obvious typos yourself. For logic changes please still do submit PRs, those are tricky with Ruby...

alexdowad:
> Let me add you as a collaborator so you could commit obvious typos yourself. For logic changes please still do submit PRs, those are tricky with Ruby...

Will do. Thanks.

PR #247, opened 2016-01-17
Ruby has no such thing as backslash-delimited strings

alexdowad:
Hey, @whitequark, please check and see if I'm totally off base here.

If I am right about this, then I think the only reason to store error lambdas in `@escape` rather than just calling `diagnostic` directly has just disappeared.

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -688,11 +688,8 @@ class Parser::Lexer
       codepoint = codepoint_str.to_i(16)
 
       if codepoint >= 0x110000
-        @escape = lambda do
-          diagnostic :error, :unicode_point_too_large, nil,
-                     range(codepoint_s, codepoint_s + codepoint_str.length)
-        end
-
+        diagnostic :error, :unicode_point_too_large, nil,
+                   range(codepoint_s, codepoint_s + codepoint_str.length)
         break
       end
 
@@ -707,9 +704,7 @@ class Parser::Lexer
   }
 
   action invalid_complex_escape {
-    @escape = lambda do
-      diagnostic :fatal, :invalid_escape
-    end
+    diagnostic :fatal, :invalid_escape
   }
 
   action slash_c_char {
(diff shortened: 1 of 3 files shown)


SilverPhoenix99:
Hi. I'm not sure if this is what you mean, but `eval('%\\a\\')` results in `"a"`.

alexdowad:
> I'm not sure if this is what you mean, but eval('%\\a\\') results in "a"

Indeed, that's exactly what I mean. Looks like we need to add a test to the test suite for that.

I need to file a bug against Pry... it doesn't understand this syntax.

alexdowad:
Hmm. It looks like master's handling of backslash-delimited strings (which **do** exist after all, I thought they didn't) is also buggy.

alexdowad:
Updated PR. @whitequark, tell me if I'm getting even further off base here.

alexdowad:
There are some cases which the new code doesn't handle right either; I need to add still more tests.

alexdowad:
Updated PR again. Still more tests added, took a different approach which I think (maybe? perhaps?) should handle all backslash-delimited strings correctly.

whitequark:
So the reason I did the `@escape` thing was that I thought there were cases where the MRI lexer, encountering a thing that looks like an escape sequence but is not a valid escape sequence, backtracks and avoids raising an error. (Backslash-delimited percent-strings are a side effect, I don't remember ever realizing that the lambda allowed those to parse successfully. I'm also surprised I didn't add any tests for them, since they've definitely failed to parse while traversing the Rubygems corpus... hrm.)

However, I've just spent a bunch of time reading parse.y and trying out things, and actually the MRI lexer never backtracks like that. I'm not sure what gave me the idea that it does, initially.

whitequark:
It's kind of stupid to have a separate pair of states just for the backslash-delimited strings literally no one ever used in the history of Ruby except people writing lexer tests (which is how I found these), but whatever.

alexdowad:
> It's kind of stupid to have a separate pair of states just for the backslash-delimited strings literally no one ever used in the history of Ruby except people writing lexer tests

Well, I find it hard to disagree with you here. But I don't think it's any less stupid to complicate the "main flow" of the escape-handling code just so it can handle backslash-delimited strings.

It would be nice if that "main flow" could be written in a straightforward way, and things like backslash delimiters would just be handled naturally, without needing special conditionals or anything like that. But I can't see right now how that could be possible.

Maybe I should have spent more time looking at how `parse.y` handles backslash delimiters.

whitequark:
> Maybe I should have spent more time looking at how parse.y handles backslash delimiters.

By not having a structure to the lexer.

alexdowad:
> By not having a structure to the lexer.

OK -- eat one character at a time, first see if it is a closing delimiter, *then* consider whether it is an escape or not.

Hmm.

PR #248, opened 2016-01-18
It's time to sizzle

alexdowad:
The object of this (not yet complete) PR is to turn `parser` from this:

tricycle-small

...into this:

nissan_skyline

I have ported `parser`'s lexer, which is far hungrier for CPU cycles than any other part, to C. Specifically, it's a Ragel file which compiles to a Ruby C extension.

No profiling and tuning whatsoever has been done on the extension yet. (I just got all the tests passing about 10 minutes ago.) But already, it lexes `lib/parser/lexer.rb` almost 8 times faster than the existing pure-Ruby lexer. Hopefully, it can be made to go significantly faster than that again. (10 would be a nice number... OK, no question about it, it will have to be made at least 10 times faster.)

Besides getting the extension all tuned up and ready to race, I need to clean up the code formatting, work on coding style and readability, do more thorough testing and verification to ensure correctness, and so on. If any interested parties are willing to help with some of this work, please pitch in.

One important thing which I have **not** worked out is how to build platform-specific gems, so that gem users will get the C extension on MRI, but pure Ruby on JRuby/RBX/etc. An alternative would be to release it as a different gem.

All feedback is appreciated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ lib/parser/ruby22.rb
 lib/parser/ruby23.rb
 lib/parser/macruby.rb
 lib/parser/rubymotion.rb
+ext/lexer/lexer.c
+lib/parser/lexer.so
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,7 @@
 require 'bundler/gem_tasks'
 require 'rake/testtask'
 require 'rake/clean'
+require 'rake/extensiontask'
 
 task :default => [:test]
 
@@ -138,8 +139,22 @@ task :changelog do
   sh('git commit CHANGELOG.md -m "Update changelog." || true')
 end
 
+Rake::ExtensionTask.new 'lexer' do |ext|
+  ext.lib_dir = 'lib/parser'
+end
+task :compile => 'ext/lexer/lexer.c'
+
 rule '.rb' => '.rl' do |t|
   sh "ragel -F1 -R #{t.source} -o #{t.name}"
+end
+
+rule '.c' => '.rl' do |t|
+  sh "ragel -F1 #{t.source} -o #{t.name}"
+
+  # Ragel likes to use int variables where a #define would do
+  src = File.read(t.name)
+  src.gsub!(/^static const int (\w+) = (\d+);/, '#define \1 \2')
+  File.open(t.name, 'w') { |f| f.write(src) }
 end
 
 rule '.rb' => '.y' do |t|
--- a/ext/lexer/extconf.rb
+++ b/ext/lexer/extconf.rb
@@ -1,0 +1,10 @@
+require 'mkmf'
+
+if RUBY_VERSION < '1.9'
+  raise "Parser's smokin' hot native-code lexer can only be built on Ruby " \
+        "1.9+. Sorry!"
+end
+
+$CFLAGS << ' -Wall -Werror -Wno-declaration-after-statement '
+$CFLAGS << ' --std=c99 -march=native -mtune=native -O2 '
+create_makefile 'lexer'
--- a/ext/lexer/lexer.h
+++ b/ext/lexer/lexer.h
@@ -1,0 +1,369 @@
+typedef enum {
+	SINGLE_QUOTE,      /* ' */
+	DOUBLE_QUOTE,      /* " */
+	PERCENT_Q,         /* %q */
+	BIG_PERCENT_Q,     /* %Q */
+	LSHFT_SINGLE_QUOT, /* <<' */
+	LSHFT_DOUBLE_QUOT, /* <<" */
+	BARE_PERCENT,      /* % */
+
+	PERCENT_W,         /* %w */
+	BIG_PERCENT_W,     /* %W */
+
+	PERCENT_I,         /* %i */
+	BIG_PERCENT_I,     /* %I */
+
+	SYM_SINGLE_QUOT,   /* :' */
+	SYM_DOUBLE_QUOT,   /* :" */
+	PERCENT_S,         /* %s */
+
+	SLASH,             /* / for regexp */
+	PERCENT_R,         /* %r */
+
+	PERCENT_X,         /* %x for xstr */
+	BACKTICK,          /* ` */
+	LSHFT_BACKTICK,    /* <<` */
+
+	INVALID
+} str_type;
+
+struct lexer_state;
+
+typedef struct literal {
+	struct lexer_state *lexer;
+	VALUE buffer;
+	int buffer_s;
+	int buffer_e;
+
+	int nesting;
+
+	str_type str_type;
+	VALUE start_tok;
+	int interpolate;
+
+	VALUE start_delim;
+	VALUE end_delim;
+	VALUE delimiter;
+
+	int heredoc_e;
+	int str_s;
+	int herebody_s;
+
+	int indent;
+	int label_allowed;
+	int interp_braces;
+	int space_emitted;
+	int monolithic;
+
+	int dedent_body;
+	int dedent_level;
+} literal;
+
+define_stack_type(lit_stack, literal, {0});
+
+typedef struct lexer_state {
+  uint cs;               /* DFA state */
+  uint p;                /* stream position */
+  uint pe;               /* end-of-stream position */
+
+  int *cs_stack;
+  int cs_stack_top;
+  int cs_stack_size;
+
+  VALUE token_queue;
+  VALUE static_env;
+
+  VALUE source_buffer;
+  VALUE source;          /* source code string */
+  VALUE source_pts;      /* source as a codepoint array */
+  VALUE encoding;
+
+  VALUE tokens;
+  VALUE comments;
+
+  stack_state cond;
+  stack_state cmdarg;
+  ss_stack cond_stack;   /* for saving 'cond' */
+  ss_stack cmdarg_stack; /* for saving 'cmdarg' */
+
+  VALUE lambda_stack;
+  int paren_nest;
+  lit_stack literal_stack;
+
+  int version;
+  int in_kwarg;
+  int force_utf32;
+
+  VALUE diagnostics;
+
+  int newline_s;         /* position of last newline encountered */
+  int eq_begin_s;
+  int herebody_s;
+  int escape_s;
+
+  int dedent_level;
+
+  VALUE escape;
+} lexer_state;
+
+static void lexer_mark(void*);
+static void lexer_dealloc(void*);
+static VALUE lexer_reset(int, VALUE*, VALUE);
+
+static void literal_init(literal*, lexer_state*, VALUE, VALUE, int, int, int, int, int);
+static str_type literal_string_to_str_type(VALUE);
+static VALUE literal_str_type_to_string(str_type);
+static void literal_set_start_tok_and_interpolate(literal*, str_type);
+static VALUE literal_get_start_delim(VALUE);
+static VALUE literal_get_end_delim(VALUE);
+static int  literal_munge_escape_p(literal*, VALUE);
+static int  literal_nest_and_close(literal*, VALUE, int, int, VALUE);
+static void literal_emit_start_tok(literal*);
+static void literal_start_interp_brace(literal*);
+static int  literal_end_interp_brace_and_close(literal*);
+static void literal_extend_string(literal*, VALUE, int, int);
+static void literal_flush_string(literal*);
+static void literal_extend_content(literal*);
+static void literal_extend_space(literal*, int, int);
+static int  literal_words_p(literal*);
+static void literal_infer_indent_level(literal*, VALUE);
+
+static void emit_token(lexer_state*, VALUE, VALUE, int, int);
+static void emit_comment(lexer_state*, int, int);
+static void emit_do(lexer_state*, int, int, int);
+
+static VALUE tok(lexer_state*, int, int);
+static VALUE range(lexer_state*, int, int);
+static void diagnostic(lexer_state*, VALUE, VALUE, VALUE, VALUE, VALUE);
+static int get_codepoint(lexer_state*, int);
+static int arg_or_cmdarg(int);
+static int is_nthref(VALUE);
+static int is_backref(VALUE);
+static int is_capitalized(VALUE);
+static int is_regexp_metachar(VALUE);
+static int eof_codepoint(int);
+static VALUE find_unknown_options(VALUE);
+static int bad_cvar_name(VALUE);
+static int bad_ivar_name(VALUE);
+static int find_8_or_9(VALUE str);
+static void emit_int(lexer_state*, VALUE, int, int);
+static void emit_rational(lexer_state*, VALUE, int, int);
+static void emit_complex(lexer_state*, VALUE, int, int);
+static void emit_complex_rational(lexer_state*, VALUE, int, int);
+static void emit_float(lexer_state*, VALUE, int, int);
+static void emit_complex_float(lexer_state*, VALUE, int, int);
+static int push_literal(lexer_state*, VALUE, VALUE, int, int, int, int, int);
+static int pop_literal(lexer_state*);
+static VALUE array_last(VALUE);
+static VALUE unescape_char(char);
+static VALUE escape_char(VALUE);
+static void lex_unicode_points(lexer_state*, int);
+
+#define emit(type) emit_token(state, type, tok(state, ts, te), ts, te)
+
+#define def_lexer_attr_reader(name) \
+  static VALUE lexer_get_ ## name(VALUE self) { \
+    lexer_state *state; \
+    Data_Get_Struct(self, lexer_state, state); \
+    return state->name; }
+
+#define def_lexer_attr_writer(name) \
+  static VALUE lexer_set_ ## name(VALUE self, VALUE val) { \
+    lexer_state *state; \
+    Data_Get_Struct(self, lexer_state, state); \
+    return state->name = val; } \
+
+#define def_lexer_attribute(name) \
+    def_lexer_attr_reader(name) \
+    def_lexer_attr_writer(name)
+
+#define Qzero INT2NUM(0)
+
+#define init_symbol(name) \
+    name = ID2SYM(rb_intern(#name)); \
+    rb_gc_register_address(&name)
+
+VALUE k__ENCODING__;
+VALUE k__FILE__;
+VALUE k__LINE__;
+VALUE kALIAS;
+VALUE kAND;
+VALUE kBEGIN;
+VALUE klBEGIN;
+VALUE kBREAK;
+VALUE kCASE;
+VALUE kCLASS;
+VALUE kDEF;
+VALUE kDEFINED;
+VALUE kDO;
+VALUE kDO_BLOCK;
+VALUE kDO_COND;
+VALUE kDO_LAMBDA;
+VALUE kELSE;
+VALUE kELSIF;
+VALUE kEND;
+VALUE klEND;
+VALUE kENSURE;
+VALUE kFALSE;
+VALUE kFOR;
+VALUE kIF;
+VALUE kIF_MOD;
+VALUE kIN;
+VALUE kMODULE;
+VALUE kNEXT;
+VALUE kNIL;
+VALUE kNOT;
+VALUE kOR;
+VALUE kREDO;
+VALUE kRESCUE;
+VALUE kRESCUE_MOD;
+VALUE kRETRY;
+VALUE kRETURN;
+VALUE kSELF;
+VALUE kSUPER;
+VALUE kTHEN;
+VALUE kTRUE;
+VALUE kUNDEF;
+VALUE kUNLESS;
+VALUE kUNLESS_MOD;
+VALUE kUNTIL;
+VALUE kUNTIL_MOD;
+VALUE kWHEN;
+VALUE kWHILE;
+VALUE kWHILE_MOD;
+VALUE kYIELD;
+
+VALUE tAMPER;
+VALUE tAMPER2;
+VALUE tANDDOT;
+VALUE tANDOP;
+VALUE tAREF;
+VALUE tASET;
+VALUE tASSOC;
+VALUE tBACK_REF;
+VALUE tBACK_REF2;
+VALUE tBANG;
+VALUE tCARET;
+VALUE tCHARACTER;
+VALUE tCMP;
+VALUE tCOLON;
+VALUE tCOLON2;
+VALUE tCOLON3;
+VALUE tCOMMA;
+VALUE tCOMMENT;
+VALUE tCONSTANT;
+VALUE tCVAR;
+VALUE tDIVIDE;
+VALUE tDOT;
+VALUE tDOT2;
+VALUE tDOT3;
+VALUE tDSTAR;
+VALUE tEH;
+VALUE tEQ;
+VALUE tEQL;
+VALUE tEQQ;
+VALUE tFID;
+VALUE tFLOAT;
+VALUE tGEQ;
+VALUE tGT;
+VALUE tGVAR;
+VALUE tIDENTIFIER;
+VALUE tIMAGINARY;
+VALUE tINTEGER;
+VALUE tIVAR;
+VALUE tLABEL;
+VALUE tLABEL_END;
+VALUE tLAMBDA;
+VALUE tLAMBEG;
+VALUE tLBRACE;
+VALUE tLBRACE_ARG;
+VALUE tLBRACK;
+VALUE tLBRACK2;
+VALUE tLCURLY;
+VALUE tLEQ;
+VALUE tLPAREN;
+VALUE tLPAREN_ARG;
+VALUE tLPAREN2;
+VALUE tLSHFT;
+VALUE tLT;
+VALUE tMATCH;
+VALUE tMINUS;
+VALUE tNEQ;
+VALUE tNL;
+VALUE tNMATCH;
+VALUE tNTH_REF;
+VALUE tOP_ASGN;
+VALUE tOROP;
+VALUE tPERCENT;
+VALUE tPIPE;
+VALUE tPLUS;
+VALUE tPOW;
+VALUE tQWORDS_BEG;
+VALUE tQSYMBOLS_BEG;
+VALUE tRATIONAL;
+VALUE tRBRACK;
+VALUE tRCURLY;
+VALUE tREGEXP_BEG;
+VALUE tREGEXP_OPT;
+VALUE tRPAREN;
+VALUE tRSHFT;
+VALUE tSEMI;
+VALUE tSPACE;
+VALUE tSTAR;
+VALUE tSTAR2;
+VALUE tSTRING;
+VALUE tSTRING_BEG;
+VALUE tSTRING_CONTENT;
+VALUE tSTRING_DBEG;
+VALUE tSTRING_DEND;
+VALUE tSTRING_DVAR;
+VALUE tSTRING_END;
+VALUE tSYMBEG;
+VALUE tSYMBOL;
+VALUE tSYMBOLS_BEG;
+VALUE tTILDE;
+VALUE tUMINUS;
+VALUE tUMINUS_NUM;
+VALUE tUPLUS;
+VALUE tWORDS_BEG;
+VALUE tXSTRING_BEG;
+
+VALUE comment_klass;
+VALUE diagnostic_klass;
+VALUE range_klass;
+
+VALUE severity_error;
+VALUE fatal;
+VALUE warning;
+
+VALUE ambiguous_literal;
+VALUE ambiguous_prefix;
+VALUE bare_backslash;
+VALUE character;
+VALUE cvar_name;
+VALUE embedded_document;
+VALUE empty_numeric;
+VALUE escape_eof;
+VALUE incomplete_escape;
+VALUE invalid_escape;
+VALUE invalid_escape_use;
+VALUE invalid_hex_escape;
+VALUE invalid_octal;
+VALUE invalid_unicode_escape;
+VALUE ivar_name;
+VALUE no_dot_digit_literal;
+VALUE prefix;
+VALUE regexp_options;
+VALUE string_eof;
+VALUE trailing_in_number;
+VALUE unexpected;
+VALUE unexpected_percent_str;
+VALUE unicode_point_too_large;
+VALUE unterminated_unicode;
+
+VALUE empty_array;
+VALUE blank_string;
+VALUE escaped_next_line;
+VALUE utf8_encoding;
+VALUE cr_then_anything_to_eol;
+VALUE crs_to_eol;
--- a/ext/lexer/lexer.rl
+++ b/ext/lexer/lexer.rl
@@ -1,0 +1,2847 @@
+#include <ruby.h>
+#include <ruby/encoding.h>
+
+#include <stdint.h>
+#include <stdio.h>
+
+#include "stack.h"
+#include "stack_state.h"
+#include "lexer.h"
+
+%%machine lex;
+%%write data;
+
+static VALUE lexer_alloc(VALUE klass)
+{
+  lexer_state *state = xmalloc(sizeof(lexer_state));
+
+  state->cs = state->p = state->pe = 0;
+  state->paren_nest = 0;
+
+  state->cs_stack      = xmalloc(4 * sizeof(int));
+  state->cs_stack_top  = 0;
+  state->cs_stack_size = 4;
+
+  state->source_buffer = Qnil;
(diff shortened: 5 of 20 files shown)


bbatsov:
I totally didn't see this one coming. :-) Very interesting endeavour!

mbj:
I like the idea of adding an optional C extension to speed the parser up. The only thing I worry about added maintenance overhead as lexing is now "more dimensional". Added maintenance overhead could cause this project to stall mid/longterm, as there is AFAIK no sponsor currently.

whitequark:
Yeah, I'm not very keen on maintaining a C lexer as well. What about handling untrusted input, too? The Ruby version is clearly safe.

whitequark:
Disregard my earlier nitpicking, it's irrelevant.

I've looked further at this change. I am not convinced that a C extension that duplicates the entire lexer is a better solution to the problem of "I want faster lexing" than getting a decent Ruby implementation instead of MRI.

Regardless of that, I will not merge this until I will conclusively see that the lexer does not exhibit undefined behavior under stress-testing; running it under asan+msan+afl for a sufficiently long time will convince me, but nothing less. People are using Parser in SaaS setting, it will be extremely irresponsible to put an RCE into their service.

whitequark:
So, for starters, can you demonstrate that your C implementation on MRI is faster than warmed-up pure-Ruby Parser on JRuby on HotSpot with Truffle?

mbj:
@whitequark, @alexdowad Maybe the best option (if the C parser is such a win its worth the time): Is having a pluggable lexer interface.

So the C lexer could be maintained in a different gem reducing the maintenance burden for @whitequark?

alexdowad:
> So, for starters, can you demonstrate that your C implementation on MRI is faster than warmed-up pure-Ruby Parser on JRuby on HotSpot with Truffle?

Haven't tried yet, so I don't know if it is faster or not. It will be interesting to find out. Not tonight, though.

whitequark:
@mbj The lexer interface is not becoming stable. If someone is willing to maintain a gem that replicates the Ruby interface as a drop-in without unnecessarily uglifying the Ruby API (the stack states are nowhere near a critical path, just do `rb_call`), sure, that can be arranged.

alexdowad:
> So the C lexer could be maintained in a different gem reducing the maintenance burden for @whitequark?

I think the first order of business, as already pointed out by @whitequark, is to confirm whether the performance benefits are really compelling or not. If the answer is yes, then...

Second order of business is to clean up the code. Third order of business is to tune it to maximum perf. Fourth is to verify it is really correct, using all available tools.

When it is as sweet as a hot fudge sundae with cherry on top, then the question is how to get it out into the users' hot little hands. If @whitequark wants it in `parser`, that's fine. If not, I wouldn't bother with things like pluggable lexers. Better to just fork the gem and make it available that way.

whitequark:
I agree.

alexdowad:
> Same goes for switch--you can replace it with if..else.

Technically, what you say is true. There is more to say on that point, but I'm not inclined to argue it right now; there are bigger fish to fry.

alexdowad:
Some progress here:

- C lexer is `valgrind`-clean
- ...And has been tweaked up to 2x its original speed. Don't ask me what it took to get there; you don't want to know. @whitequark will probably go into cardiac arrest if he reads the code.

Just pulled the latest JRuby `master`, did a fresh build with `mvn clean install`, and tried the resulting JRuby+Truffle. It's very puzzling just how slow it is. It takes about 20 minutes to run the same (trivial) benchmark which MRI + pure-Ruby lexer completes in 20 seconds (and MRI + C lexer in less than 2 seconds).

???

whitequark:
cc @headius--does JRuby do something silly there like bail out on a large method?

whitequark:
cc @headius --does JRuby do something silly there like bail out on a large method?

headius:
JRuby+Truffle when run on a normal JVM (not one that has the Graal compiler) will run *very* slow. If you want to get realistic results for this, you should be running normal JRuby. JRuby+Truffle is experimental, does not support Java extensions, is not yet feature-complete, and will NOT run fast unless run on a Graal-capable JVM.

whitequark:
@alexdowad alright then. Please check regular JRuby on regular HotSpot and, to get some idea of what happens when/if Graal lands, JRuby+Truffle on a Graal-enabled HotSpot too.

alexdowad:
> unless run on a Graal-capable JVM.

So how does one get a Graal-capable JVM? We are just assessing whether it would be worthwhile for this gem to add a C extension, and would like to know just how fast pure Ruby can go when the latest JRuby hotness lands.

headius:
FYI, if (when) we port the ext to Java for JRuby, parser would need to release a -java platform version that includes it. There's no way to choose a different dependency at install time, so it all has to be done via gem platforms.

I don't think the port would take more than a couple hours.

alexdowad:
> Please check regular JRuby on regular HotSpot

It's merely 60 times slower than MRI + C lexer.

headius:
> So how does one get a Graal-capable JVM?

http://www.oracle.com/technetwork/oracle-labs/program-languages/downloads/index.html

Note this is not the open-source version, but apparently the closed-source Graal JDK has some super secret optimizations that make JRuby+Truffle run better.

> We are just assessing whether it would be worthwhile for this gem to add a C extension, and would like to know just how fast pure Ruby can go when the latest JRuby hotness lands.

JRuby+Truffle is probably still a year or more away from being production ready, and has other aspects (startup time, memory requirements) that mean JRuby proper will be the JRuby of choice for quite some time. If there's a C ext for MRI, it would be worth having a Java ext for JRuby.

headius:
Incidentally, what's the benchmark you're running? I'd like to try it on JRuby+Hotspot and JRuby+Truffle.

whitequark:
@headius No, the idea is to *avoid* having a C extension for MRI and a Java extension for JRuby. One lexer is a maintenance nightmare already, I cannot imagine having to support three.

Specifically, if some JRuby sekret magick can run the existing pure-Ruby lexer on par with the MRI C extension, then I will direct anyone unsatisfied with performance of Parser to just go install JRuby. If not... yeah, we have to look at extensions. *sigh*

alexdowad:
Sorry, I was about to post my trivial benchmark but got distracted:

```ruby
$:.unshift(File.join(File.dirname(__FILE__), 'lib'))
require 'parser'
require 'benchmark'

class Diagnostics
  def process(_)
  end
end

path = 'lib/parser/lexer.rb'
src = File.read(path)

puts "Encoding: #{src.encoding}"

times = 10

puts "Lexing #{path} #{times} times"

result = Benchmark.measure do
  times.times do
    lexer = Parser::Lexer.new(23)
    lexer.diagnostics = Diagnostics.new
    buffer = Parser::Source::Buffer.new(path)
    buffer.source = src
    lexer.source_buffer = buffer
    loop do
      token = lexer.advance
      break unless token[0]
    end
  end
end

puts "ELAPSED:"
puts result
```

alexdowad:
As you can see, it just lexes the lexer 10 times.

alexdowad:
> One lexer is a maintenance nightmare already, I cannot imagine having to support three.

To be clear about my intentions, I am not planning to just dump my white-hot blazingly-fast lexer on some poor, hapless gem maintainer and walk away from it. If it's broken, I intend to fix it.

whitequark:
That still creates problems. I would like to figure out if we can avoid those entirely before looking at how we can partially mitigate them. Security audit alone will take a disproportionate amount of time.

I'm not outright opposed to merging your C lexer, but I would still like to avoid that if we can get an equal increase in user experience via some other way.

whitequark:
What % of rubocop's runtime is spent in the lexer, anyway?

headius:
I'll run some numbers.

alexdowad:
> What % of rubocop's runtime is spent in the lexer, anyway?

On one auto-correction benchmark, I measured something like 49%.

headius:
Is it typical to run the parser with diagnostics?

alexdowad:
> Is it typical to run the parser with diagnostics?

You can just remove that bit if you like. But all that means is that the script will error out with an uncaught exception if there is a lexer error. There shouldn't be any lexer errors, so it doesn't really matter either way.

whitequark:
@alexdowad So, we need to take into account that no change to the lexer we can make will speed up rubocop by a factor of more than 2x. How representative of a typical workload is that benchmark? Will your near-2x improvement in lexing speed result in a qualitative improvement of user experience on a typical Rubocop workload, or merely a modest quantitative improvement? (@bbatsov might be able to answer that?) Given that most files don't change a lot, would it make more sense to cache ASTs between Rubocop runs, keyed by the file hashsum? That could save the analysis time too, not just parsing time.

whitequark:
sorry, near-2x improvement in Rubocop processing speed.

headius:
Given the results I'm seeing (it's taking a long time to gather numbers) I'm pretty convinced this thing is bottlenecked on allocation. CPU never rises above perhaps 25% of my whole system, telling me it's memory pipeline standing in the way of perf. Some of that is inherent to Ruby, but a lot of it could probably be improved. I would be surprised if we couldn't get 2x by doing some allocation profiling and fixing.

alexdowad:
> no change to the lexer we can make will speed up rubocop by a factor of more than 2x.

Absolutely. But if the lexer is not changed, no other optimization will ever speed RC up by more than 2x, even if *everything* else is optimized right down to zero.

If this bottleneck is killed, that doesn't mean I stop there. Then I kill the next bottleneck, and then the next one, until RC runs so fast that it knocks you out of your seat every time you lint your code.

whitequark:
@headius 2x wrt current JRuby speed or 2x wrt current MRI speed?

alexdowad:
> CPU never rises above perhaps 25% of my whole system

On JRuby? Running pure-Ruby `parser` on MRI, I peg one CPU core at 100%.

whitequark:
@alexdowad My suggestion of caching ASTs or even analysis results will speed up RC regardless of how fast the lexer or even RC itself is.

headius:
@whitequark Potentially both, if the allocations are for the same reasons...but the tooling I'd use is for JVM and will show JRuby internals that are doing too much allocation as well.

@alexdowad Yes, but MRI does its GC in the same thread. JVM GC is spread across all cores, and seems to be taking up a fair amount of time.

alexdowad:
> My suggestion of caching ASTs or even analysis results

We already cache analysis results. Sorry I didn't respond to that bit.

whitequark:
@alexdowad So the slowdown incurred by the lexer is the difference between near-instantaneous results and single second wait times? I find these kinds of performance improvements more compelling if so, since it is a qualitative change indeed.

alexdowad:
> So the slowdown incurred by the lexer is the difference between near-instantaneous results and single second wait times?

@whitequark From my investigation thus far, I believe that a fast lexer, without any other changes, will make a qualitative difference in the experience of those using RC. It will not make results near-instantaneous. BUT, here's the thing: if the lexer is not made dramatically faster, no other optimizations will make results near-instantaneous. If the lexer is fast, then other optimizations *do* have the potential to make results near-instantaneous.

That is my conclusion thus far, but I'm not married to it and would love to be proved wrong. Wrongness-provers of the world, please step up and show me what I'm missing!

whitequark:
Alright, count me mildly in favor of merging, all other things equal.

alexdowad:
> Yes, but MRI does its GC in the same thread. JVM GC is spread across all cores, and seems to be taking up a fair amount of time.

MRI runs this particular benchmark about 5x faster than JRuby, so it's hard to make a meaningful comparison.

From my analysis thus far, I believe that the lexer "core" -- the main loop generated by Ragel -- is the bottleneck for the pure-Ruby lexer. Again, I would like to be proved wrong.

headius:
Ok, here's some results. Still seems allocation-bound to me. Ragel generates weird code.

JRuby master on OpenJDK 1.8.0_60 process size around 250MB):
```
Encoding: UTF-8
Lexing lib/parser/lexer.rb 10 times
 26.390000   0.250000  26.640000 ( 21.251906)
 22.560000   0.110000  22.670000 ( 21.454913)
 21.640000   0.090000  21.730000 ( 21.079679)
 21.120000   0.090000  21.210000 ( 20.652302)
 21.950000   0.090000  22.040000 ( 21.358344)
 21.640000   0.090000  21.730000 ( 21.073522)
 21.650000   0.080000  21.730000 ( 20.916628)
 22.500000   0.130000  22.630000 ( 21.859048)
 21.350000   0.080000  21.430000 ( 20.783937)
 21.190000   0.080000  21.270000 ( 20.568819)
ELAPSED:
222.040000   1.090000 223.130000 (211.014425)
```

JRuby+Truffle on Graal does not fare well. There's a LOT of debugging output that seems to indicate optimizations are not working right:
```
Encoding: UTF-8
Lexing lib/parser/lexer.rb 10 times
 194.560000   1.320000  195.880000 ( 89.395241)
 187.070000   0.960000  188.030000 ( 88.458658)
 169.980000   0.700000  170.680000 ( 84.974119)
 174.700000   0.860000  175.560000 ( 86.473328)
 214.900000   1.600000  216.500000 ( 88.799379)
 217.920000   1.630000  219.550000 ( 91.442958)
 187.070000   0.960000  188.030000 ( 88.458658)
 169.980000   0.700000  170.680000 ( 84.974119)
 166.110000   0.770000  166.880000 ( 85.202306)
 164.830000   0.800000  165.630000 ( 84.636863)
ELAPSED:
 1815.560000   9.930000  1825.490000 ( 866.311150)
```

I'm going to do an allocation profile.

whitequark:
@alexdowad Have you looked at optimizing the code that lexes strings? Currently it does so character by character, which is mortifyingly slow (several Ruby calls per character!) and allocates tons of garbage. I planned to fix that years ago but never quite found time for it.

alexdowad:
> count me mildly in favor of merging

Don't speak too quickly, until after you see some of the crazy things I am doing for perf. Make sure to have a barf bag handy.

> Have you looked at optimizing the code that lexes strings?

Good idea. This is orthogonal from the question of whether or not to add a C extension; even if we do, some people will still rely on the pure-Ruby lexer, so it is always worthwhile to make it faster.

headius:
FWIW, I'm also in favor of having an ext for the hottest inner loops of the parser. Ruby makes parsing easy but definitely not fast.

alexdowad:
> I'm also in favor of having an ext for the hottest inner loops of the parser.

Generating the lexer with Ragel makes it tricky to use an extension for only the inner loops. We could have an extension which would implement all the lexer actions by calling into Ruby-land, though.

headius:
A quick allocation profile here shows most of the allocation is Object[], which appear to be from JRuby's interpreter allocating temporary variable space. This tells me that we're probably not actually jitting some code here.

@whitequark To answer your earlier question in a different way...yes, JRuby does sometimes bail out and not JIT methods that are too big -- mostly because the JVM has method size limits and even within those limits there's method sizes that Hotspot will refuse to JIT.

Turning on JIT logging, I see:
```
2016-01-20T06:20:12.897-06:00: JITCompiler: Could not compile; passes run: []:Lexer.Parser::Lexer.advance at /Users/headius/projects/parser/lib/parser/lexer.rb:10789 because of: "Could not compile org.jruby.internal.runtime.methods.MixedModeIRMethod@1d119efb; instruction count 19285 exceeds threshold of 2000"
```

lexer.rb at line 10789 is the `advance` method. So the lexer's advance is not actually jitting and is running interpreted in JRuby. It's too big.

I'll fiddle with it and see if I can come up with anything.

headius:
Yeah, we're not going to be able to compile this method.

I bumped up the size limit from 2000 instructions to 20000 instructions, and as a result the JIT compiler needed a 2GB heap to work with. Once I gave it that, it bailed out because the resulting bytecode was too large for the JVM to load (exceeded class file limitations).

As long as this method is running interpreted, it's going to be 5-20x slower than it could be.

whitequark:
@headius It's possible to reduce the size of the generated lexer. The current mode should be `ragel -F1`; try `-F0`, `-T1` or at last `-T0`. It should cut down on code duplication very significantly.

whitequark:
I wonder if we should try newer ragel, too.

headius:
The biggest problem is all the inlined when bodies. If they were separate methods, or even procs, we'd be able to do a MUCH better job of compiling this. With it all inlined, it's simply enormous.

I'll try the other ragel flags and see if it improves.

whitequark:
Yes, changing the ragel flags should help with that. With `-T1` already I think every semantic action should appear only once in the output.

whitequark:
Beware that you may have to add or remove some of the `_lex_*` aliases in lexer.rl:293 when changing the generation flags.

whitequark:
Making the bodies of semantic actions procs is a non-starter with Ragel 6 but should be easily doable with Ragel 7, so that's something to investigate, although I'm not overly excited about shipping my own Ragel for CI and such either.

headius:
Even with -T0 the advance method is around 5000 lines and over 10000 instructions. It's still too big for the JRuby JIT to handle it. I was able to get a bit closer by turning on invokedynamic, but the generated code still exceeds the 16-bit maximum size of a JVM method body.

It may be possible to post-process the output, I'm not sure. Giant methods are just bad for optimization in modern tiered runtimes.

whitequark:
Post-processing of tool output is out of question... Ragel 7 with (effectively) a custom generator for JRuby might help us gain some ground back, but that's non-trivial too. Hm. Not sure what to do, yet.

alexdowad:
> Not sure what to do, yet.

@headius Is there good reason to believe that if `advance` could be made small enough for JRuby to optimize it, and Truffle gets really, really real, that it could beat the current MRI performance by at least 4-5x?

If not, we are still looking at using an extension (or extensions) to get good performance.

headius:
In JRuby 1.7 we went to extra lengths to compile large methods like this. Specifically, when a method exceeds 500 lines, we break it up into 500 line chunks and chain them together. This has not been implemented for JRuby 9k, because such methods are usually rare. We may want to do it.

Here's JRuby 1.7 (which is able to compile Lexer.advance by chaining) running the benchmark:
```
Encoding: UTF-8
Lexing lib/parser/lexer.rb 10 times
 20.640000   0.390000  21.030000 ( 15.482000)
 10.230000   0.060000  10.290000 (  9.425000)
  9.740000   0.060000   9.800000 (  9.708000)
  9.830000   0.060000   9.890000 (  9.694000)
  9.860000   0.050000   9.910000 (  9.843000)
  9.890000   0.050000   9.940000 (  9.680000)
  9.620000   0.040000   9.660000 (  9.582000)
 10.050000   0.050000  10.100000 (  9.667000)
  9.550000   0.030000   9.580000 (  9.524000)
  9.560000   0.040000   9.600000 (  9.520000)
ELAPSED:
109.710000   0.860000 110.570000 (102.305000)
```

Chained method bodies do have to deoptimize many of the best optimizations, so I'm not surprised we only get about 2x improvement.

This is still 4x slower than MRI, though. Something's still not right.

headius:
@alexdowad Both are possible, yes. Truffle has shown performance improvements of 50-100x over MRI. JRuby itself usually beats MRI by 2-5x or more. This particular codebase showcases some things we are obviously not optimizing well.

alexdowad:
Hmm. So do we go ahead with developing extensions, or do we focus on JRuby+Truffle to save us? @whitequark, your opinion?

whitequark:
@alexdowad I highly prefer the JRuby path. Can you take a look at using Ragel 7 to factor out the semantic actions as closures? Your C mad skillz will prove quite handy there...

headius:
The Ragel 7 direction could lead to improvements in other projects too. I believe Oga uses Ragel for its lexer too, though XML is obviously much simpler.

headius:
My thoughts:

* It will take a fair bit of work to get the Ragel output to be sane for JRuby's JIT. It's probably worth it.
* The C ext would not be difficult to also implement for JRuby since it's still being generated by Ragel. It may be worth doing it in the short term but pushing it to a separate gem (e.g. if you install 'parser-ext' it will be used, but the main parser remains pure-Ruby).
* It's possible @chrisseaton may be able to tweak a few things to get the JRuby+Truffle output to look better. @chrisseaton: I just bundle installed with JRuby proper, `rake build` with JRuby proper, and then added racc/lib and ast/lib to -I for running with Truffle. LOTS of warnings about OSR failing, over and over again.

We JRuby folks are willing to put some cycles in here to get things running faster.

whitequark:
> It will take a fair bit of work to get the Ragel output to be sane for JRuby's JIT. It's probably worth it.

One thing that Ragel 7 adds is code generation backends that are completely decoupled from the state machine compiler. This is one reason I suggest this at all--it should be doable without delving too much into the obscure guts of Ragel.

headius:
@whitequark That sounds nice. I once attempted to make a JVM bytecode backend for Ragel, so it could use direct jumps rather than jump tables. The generator code was so intertwined with the state machine compiler it made my eyes bleed.

alexdowad:
> Can you take a look at using Ragel 7 to factor out the semantic actions as closures?

You are suggesting writing a custom backend, right?

Can you give a brief sketch of what you would like the generated code to look like?

chrisseaton:
@alexdowad thanks for the interest in Truffle, but I think this benchmark is unlikely to be very fast on JRuby+Truffle yet. We are only just starting to implement our big-ideas for string optimisation that we hope will make temptlating engines and benchmarks such as this fast (for example https://github.com/jruby/jruby/pull/3605).

It's probably best to concentrate on normal JRuby unless you are interested in research-level work.

whitequark:
> You are suggesting writing a custom backend, right?

I think this can be a minor modification of the existing Ruby backend.

> Can you give a brief sketch of what you would like the generated code to look like?

This is best answered by @headius.

whitequark:
@chrisseaton I'm not sure ropes will help--we (ideally; there's some overly silly code for now) allocate very few strings. Rather we have a state machine, dispatching over numeric codepoints of one large string, compiled to an extremely large method, and it's not being JITted...

chrisseaton:
@whitequark ah sorry I skimmed the issue too lightly.

We may be able to do a little better than JRuby due to the OSR that @headius already mentioned, but yeah there's a limit to how big a method anyone can JIT.

SilverPhoenix99:
Hi, all.

I arrived very late to the conversation, but anyway, here's my 2 cents...

One option is to split the lexer into parts, where one of them is only for string lexing, which could be isolated into it's own (smaller) method. Of course this is just an idea, and it may not reduce the method sizes enough for jit.

Another thing about strings, is that you can create a peek() method that returns a special character for the string delimiter, if cs is in one of the string states. Since ragel accepts signed alphabets you could use a negative character as a special mark. For example, if the special mark is -2, then something like:

```
STRING_DELIM = -2

def peek
  return STRING_DELIM if @cs = :STRING_EXPR && @data[@p] == literal.delimiter
  # ...
end
```

And in Ragel:

```
getkey peek;

# ...

string_delimiter = -2;

STRING_EXPR := |*
  string_delimiter => do_string_end;

  '#{' => do_string_dbeg;
  '#@' identifier => do_string_dvar;

  '\\' string_delimiter;
  any;
*|;
```

Note that this reduces performance since getkey is now a method, hence the idea to isolate the string lexing into it's own state. Still it shouldn't be a bottleneck.

chrisseaton:
> One option is to split the lexer into parts

Right, but that's a job for the VM really, not the program authors.

alexdowad:
> Another thing about strings, is that you can create a peek() method that returns a special character for the string delimiter, if cs is in one of the string states.

Hmm. This might be a good idea. Need to think about it more.

whitequark:
I think that a good enough solution for the lexer that doesn't non-string-related common paths is to specialize it for the most popular types of strings. A bit dirty but extremely effective.

alexdowad:
I think my conclusion is that I will:

1. Fix up the pure-Ruby lexer to make it lex strings in chunks rather than 1 character at a time, then
2. Finish my C lexer (it's most of the way there, anyways) and see how useful (or not) it becomes, then
3. Look at making Ragel generate output which is better for JIT.

whitequark:
Sounds good.

headius:
@alexdowad Ping me when you're satisfied with the C lexer. I'd like to try a Java version, and it shouldn't be much work to duplicate yours.

alexdowad:
> Ping me when you're satisfied with the C lexer.

Will do. I don't know about the "won't be much work" part...

headius:
Oh, I didn't see the .rl for the C ext. Yeah, MRI C API garbage all over :-(

alexdowad:
> Yeah, MRI C API garbage all over

You haven't seen the optimized version. It's far worse. I do all kinds of fun things, like poking into the `RArray` struct which MRI uses for arrays, and tweaking the flags manually...

alexdowad:
Using the C lexer with RuboCop, the speedup obtained when inspecting (rather than correcting) code is rather disappointing. Just about 10%.

I need to do more benchmarking to confirm whether there will be a really compelling speedup when correcting code or not. If not, there is no need to merge the C lexer.

whitequark:
Why is the correcting path so slow?

alexdowad:
> Why is the correcting path so slow?

O(n^2) runtime. I should try to tame that first.

headius:
I was looking over lexer.rl and have a few questions.

* There's a number of productions (if that's the right term) that inline large chunks of code that could easily be a method call. They'd help greatly in reducing the size of the Lexer#advance method. Is there a reason for this?
* Are there any other flags for ragel than the -T0 etc flags? Does it always inline this logic or is it possible to make Ragel do more method-based refactoring of the resulting code?

IMHO the code Ragel produces for Ruby is *heinous*. There's code like this inline multiple places:

```ruby
escape = {" " => '\s', "\r" => '\r', "\n" => '\n', "\t" => '\t',
           "\v" => '\v', "\f" => '\f'}[@source[@ts + 1]]
```

You read right, that's creating a new Hash plus 12 new Strings every time it is encountered, when it could create none of those by referencing a constant.

Or this:

```ruby
                  if current_literal.munge_escape? escaped_char
                    # If this particular literal uses this character as an opening
                    # or closing delimiter, it is an escape sequence for that
                    # particular character. Write it without the backslash.

                    if current_literal.regexp? && REGEXP_META_CHARACTERS.match(escaped_char)
                      # Regular expressions should include escaped delimiters in their
                      # escaped form, except when the escaped character is
                      # a closing delimiter but not a regexp metacharacter.
                      #
                      # The backslash itself cannot be used as a closing delimiter
                      # at the same time as an escape symbol, but it is always munged,
                      # so this branch also executes for the non-closing-delimiter case
                      # for the backslash.
                      current_literal.extend_string(tok, @ts, @te)
                    else
                      current_literal.extend_string(escaped_char, @ts, @te)
                    end
                  else
                    # It does not. So this is an actual escape sequence, yay!
                    if current_literal.regexp?
                      # Regular expressions should include escape sequences in their
                      # escaped form. On the other hand, escaped newlines are removed.
                      current_literal.extend_string(tok.gsub(ESCAPED_NEXT_LINE, BLANK_STRING), @ts, @te)
                    else
                      current_literal.extend_string(@escape || tok, @ts, @te)
                    end
                  end
```

This is inline in the lexer several places, doing exactly the same thing in each place. It does not set and local variables, and only reads one. Why inline this?

I think there's still an impression of Ruby that method calls are very expensive. They're not cheap or free, as in an optimizing runtime (e.g. JRuby/JVM) but they're nowhere near as expensive as they used to be...especially if they just have simple argument forms (no optionals or restargs).

I have a feeling there's a lot we could do to clean up the existing lexer without going to C.

whitequark:
> There's a number of productions (if that's the right term) that inline large chunks of code that could easily be a method call. They'd help greatly in reducing the size of the Lexer#advance method. Is there a reason for this?

Semantic actions. I don't think there is any particular reason, except that this Ruby lexer is quite possibly the largest Ruby+Ragel program ever written.

> Are there any other flags for ragel than the -T0 etc flags?

Not ones useful for this.

> Does it always inline this logic or is it possible to make Ragel do more method-based refactoring of the resulting code?

It does not have anything more finely grained than the range of code you've seen with flags from `-T0` to `-F1`.

> You read right, that's creating a new Hash plus 12 new Strings every time it is encountered, when it could create none of those by referencing a constant.

I wrote that chunk of code, it's in a semantic action! Ragel itself generates nothing of the sort.

> This is inline in the lexer several places, doing exactly the same thing in each place. It does not set and local variables, and only reads one. Why inline this?

This should only happen on `-F0` and higher. On `-T0` or `-T1` you should see every semantic action exactly once in the output... perhaps you're looking at stale output? Try `rake regenerate`.

headius:
Yesterday I spent some time doing hand-refactoring of the lexer, and this was the result of about an hour's work: https://gist.github.com/headius/cda9350b751449379318

This has been formatted for readability, so it appears to be more lines, but this version of the lexer is much closer to actually compiling in JRuby (i.e. it has one tenth as many IR instructions, so we don't boot it out anymore). And I didn't do a particularly good job of reducing duplication; many of the methods I pulled out could be further deduplicated by passing some literals/tokens in.

In any case...I don't suggest that we hand-refactor the lexer or anything absurd like that, but if we could  modify the source or get Ragel to do some "best practice" restructuring, I think we'd end up with a more robust and optimizable lexer.

alexdowad:
@headius We can (and will) definitely optimize the lexer further. I am just seeing if RC's performance problems can be solved without a C lexer; waiting for @whitequark to review 9aaa705, since that is holding me up on fixing RC's autocorrection.

headius:
> I wrote that chunk of code, it's in a semantic action! Ragel itself generates nothing of the sort.

Yes, I know you wrote that...I'm just pointing out that this is a huge wart on the lexer. Anything that encounters this is creating literally dozens of objects on JRuby and MRI alike just to pick out one character from a hash lookup.

Some auditing of what Ragel is actually creating would be very valuable.

> This should only happen on -F0 and higher. On -T0 or -T1 you should see every semantic action exactly once in the output... perhaps you're looking at stale output? Try rake regenerate.

That's great, but are you saying you'd be willing to use -T0 or -T1 for shipped parser gem?

And even in those modes, there's code inline in the lexer that could be callouts to methods or constants. I'm sure we could shrink the overall size and complexity of the lexer by many times.

headius:
> And even in those modes, there's code inline in the lexer that could be callouts to methods or constants. I'm sure we could shrink the overall size and complexity of the lexer by many times.

...and the resulting code was still extremely large...too large for us to compile.

whitequark:
I see, if we have to extract it into methods or closures anyway, then of course it only makes sense to use `-F1`.

alexdowad:
> Yes, I know you wrote that...I'm just pointing out that this is a huge wart on the lexer

That code is almost never executed. It is only invoked to handle a specific parser error involving the misuse of `?` character constants. I don't think it's worthwhile to extract the hash into a constant.

enebo:
@alexdowad but the general problem all JIT's will have is seeing a large method.  So the cost of the hash itself is not important from performance but it is important from a space perspective.  const lookup will be less bytecode than the bytecodes for literal construction and also be faster (which is in this case is not important).

alexdowad:
> const lookup will be less bytecode than the bytecodes for literal construction

Gotcha. OK, when I work on making this lexer JIT-friendly, I will look at that.

enebo:
@alexdowad as a mild followup in Java VMs but I suspect all VMs with profiling; lots of small methods means profiled optimizations makes decisions on which code paths are important and they aggressively inline the small methods which are used the most.  A static compiler on the other hand cannot predict importance on usage which is why I think some of the generation scripts aggressively duplicate code (e.g. making inlining decisions when it cannot make them based on profiling).  You probably know that but I am just pointing out if every body in a case/when was a method a JIT would still end up making it fast (assuming it can natively compile the method -- which is the problem for JRuby/Java here).  It just might warm up slower.

enebo:
Oh I should point out the problem with my previous statement though...On MRI it has no JIT so what is optimal code for JRuby may end up being non-optimal on MRI.  So if there is to be one codebase for the lexer then we probably want some hybrid where lexer is small enough to be native code but in which known hottest paths for the most common tokens of Ruby are inlined manually via lexer generation.

whitequark:
I have decided against merging a C extension. It adds maintenance burden and does not help JRuby and Rubinius (since those runtimes severely pessimize when they have to stay compatible with this API). I will not in good conscience enable it by default, since it is hard to prove it memory-safe and harder still to maintain it so. Thus, it provides too little gain for too much work.

The only ways to improve performance that I find acceptable are those that 1) use a memory-safe language for the lexer 2) benefit all runtimes indiscriminately 3) allow to use just a single lexer source.

  * Patching Ragel to emit a more JIT-friendly form of code works. 
  * Rewriting the lexer in Rust (or what you have) independent of the MRI C API, and then binding it via FFI or a C extension or whatever, works, although we'll have to build binary gems then. 

Anything else... nope.

PR #249, opened 2016-01-20
Moving forward

alexdowad:
Please have a look.

On `lib/parser/lexer.rb`, which hardly contains a single string literal, this lexer (with special states for single-quoted and double-quoted strings) is about 5% slower. But on `lib/parser/ruby18.rb`, which has LOTS of single-quoted string literals, it is somewhat less than 2x faster.

My highly scientific "lex a file 10 times" benchmark measures 4.6 secs on `ruby18.rb`, down from 8.7 secs on `master`.

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -841,7 +841,6 @@ class Parser::Lexer
     # tLABEL_END is only possible in non-cond context on >= 2.2
     if @version >= 22 && !@cond.active?
       lookahead = @source[@te...@te+2]
-      lookahead = lookahead.encode(@encoding) if @need_encode
     end
 
     current_literal = literal
--- a/test/test_lexer.rb
+++ b/test/test_lexer.rb
@@ -299,6 +299,16 @@ class TestLexer < Minitest::Test
                    :tLABEL_END,       "'")
   end
 
+  def test_label_with_utf32_encoded_source__22
+    setup_lexer 22
+    @lex.force_utf32 = true
+    assert_scanned("{'a':",
+                   :tLBRACE,          '{',
+                   :tSTRING_BEG,      "'",
+                   :tSTRING_CONTENT,  'a',
+                   :tLABEL_END,       "'")
+  end
+
   def test_label_colon2__22
     setup_lexer 22
 


alexdowad:
Hey @whitequark, about the handling of string literals here...

An alternative would be to write our own "main loop" for lexing string literals, and not use Ragel's state machines for that. When the beginning of a string is identified, call into a lexer action which lexes the string content, updates `@p` (Ragel has a built-in command to do this, can't remember it right now), then returns from `advance`.

Use a regexp which eats all string content up to an escape or delimiter. Use `regexp.match(@source, @p)` to start it at the right place.

Bad idea?

whitequark:
@alexdowad That is one way to handle string literals, but I'm concerned that it may duplicate too much code. You still have to handle heredocs, escapes, interpolations and other weirdness...

alexdowad:
> That is one way to handle string literals, but I'm concerned that it may duplicate too much code.

Hmm.

Optimizing single and double-quoted strings the way this PR does it doesn't seem to be bringing the hoped-for performance benefits. Yes, on files (like Racc grammars) with vast numbers of long single or double-quoted strings, the numbers look great. But on an "average" Ruby codebase, with moderate numbers of string literals, it's just a blip. And what is troubling is that on files with *no* string literals, it is actually slower.

In the case where there are no string literals, it's not executing any extra code, but perhaps factors like extra code size/data size and resulting loss of cache locality could be involved. That's just speculation, but anyways the fact is that it is slower.

I believe that the approach using regexp has the potential to be faster than this one, and could optimize more types of string literals without bloating code size (by picking a different regexp based on the string type). But there is no doubt that the mix of approaches will make the code harder to understand.

I don't think that heredocs, etc. will present any special problem. The regexp-based code would just do what the existing `plain_string`, `interp_string`, etc. lexer states do.

Anyways, I guess the question right now is whether you want to merge d8dd3bc and 6fa2118 or not.

whitequark:
Hm, I am quite unsure, if it pessimizes the common case.

alexdowad:
How about the other commits here? Perhaps we merge them first and then think more about the string literals?

alexdowad:
I just discovered a problem with `Source::Rewriter` while working on RC. Please have a look at the last commit here.

alexdowad:
Doing more work on `Source::Rewriter`, will push soon.

alexdowad:
Hey @whitequark, I have removed the commits which added special lexer states for single-quoted and double-quoted strings. I have spent quite a bit of time trying to get `Source::Rewriter` more robust. Most of it was just figuring out what the rules should be for determining which updates conflict and which don't.

There are a lot of different ways to skin this cat, but many of them lead to strange, inconsistent behavior in corner cases. So far, what I have here is the most logical, consistent set of rules I have come up with. I have added a rather lengthy comment explaining the logic.

alexdowad:
To reduce size of PR, merged commits which only affect comments or which only add missing unit tests. All commits which affect actual library code are still waiting here.

alexdowad:
Size of this PR is now much reduced.

whitequark:
LGTM on Source::Rewriter. Not sure why encoding changes are in the same PR, but anyhow, I would like to see the addition of a UTF-32-forced run of the parser testsuite before merging any of those.

alexdowad:
> LGTM on Source::Rewriter.

I'll split it out into a separate PR.

alexdowad:
> I would like to see the addition of a UTF-32-forced run of the parser testsuite before merging any of those.

It turns out this was very prudent. The UTF-32-forced parser tests revealed that a5bdc4b was not sound.

PR #251, opened 2016-01-24
- Add :csend to Parser::Meta::NODE_TYPES

mbj:
Also fixes a duplicate `send` member.

### DIFF
--- a/lib/parser/meta.rb
+++ b/lib/parser/meta.rb
@@ -16,7 +16,7 @@ module Parser
         match_with_lvasgn match_current_line
         module class sclass def defs undef alias args
         cbase arg optarg restarg blockarg block_pass args def kwarg kwoptarg
-        kwrestarg send super zsuper yield block send
+        kwrestarg send csend super zsuper yield block
         and not or if when case while until while_post
         until_post for break next redo return resbody
         kwbegin begin retry preexe postexe iflipflop eflipflop


mbj:
@whitequark This passes all tests but changes the coverage as measured by simplecov. I've no idea on how to configure the coverage expectation, nor why this constant literal change triggers a metric difference at all.

alexdowad:
> I've no idea on ... why this constant literal change triggers a metric difference at all.

Because Coveralls is junk?

alexdowad:
Thank you for catching this!

mbj:
> Because Coveralls is junk?

One possible cause is that the constant literal case increases the overall LOC, that gets measured against the explicit covered LOC. But this is not the case here.

whitequark:
I don't really care about spurious Coveralls changes, they are primarily used to measure the coverage of grammars.

mbj:
@whitequark Can we get a release with that one? It would fix a special case on my side.

alexdowad:
> Can we get a release with that one? It would fix a special case on my side.

I would recommend we finish #249 as well -- better if it makes it into the next minor release.

whitequark:
#249 is a lot of code I won't have time to review for a while.

mbj:
> #249 is a lot of code I won't have time to review for a while.

Also #249 is not a semantic change. It should not block a fix release IMO.

alexdowad:
> #249 is a lot of code I won't have time to review for a while.

Very well.

mbj:
> Also #249 is not a semantic change. It should not block a fix release IMO.

Correction it is a semantic change. But not a bugfix.

whitequark:
`2.3.0.2`

mbj:
thx

Issue #252, opened 2016-01-25
Crash when parsing symbols


mbj:
The following file (reduced from: https://github.com/ruby/spec/blob/master/core/symbol/casecmp_spec.rb) crashes parser:

```ruby
# -*- encoding: us-ascii -*-

p :"\xC3"
```

Backtrace:

```
/home/mbj/.gem/ruby/2.3.0/gems/parser-2.3.0.2/lib/parser/builders/default.rb:172:in `to_sym': invalid encoding symbol (EncodingError)
        from /home/mbj/.gem/ruby/2.3.0/gems/parser-2.3.0.2/lib/parser/builders/default.rb:172:in `symbol_compose'
        from ruby23.y:1839:in `_reduce_467'
        from /home/mbj/.rubies/ruby-2.3.0/lib/ruby/2.3.0/racc/parser.rb:259:in `_racc_do_parse_c'
        from /home/mbj/.rubies/ruby-2.3.0/lib/ruby/2.3.0/racc/parser.rb:259:in `do_parse'
        from /home/mbj/.gem/ruby/2.3.0/gems/parser-2.3.0.2/lib/parser/base.rb:162:in `parse'
        from /home/mbj/.gem/ruby/2.3.0/gems/parser-2.3.0.2/lib/parser/runner/ruby_parse.rb:132:in `process'
        from /home/mbj/.gem/ruby/2.3.0/gems/parser-2.3.0.2/lib/parser/runner.rb:202:in `process_buffer'
        from /home/mbj/.gem/ruby/2.3.0/gems/parser-2.3.0.2/lib/parser/runner.rb:195:in `block in process_files'
        from /home/mbj/.gem/ruby/2.3.0/gems/parser-2.3.0.2/lib/parser/runner.rb:181:in `each'
        from /home/mbj/.gem/ruby/2.3.0/gems/parser-2.3.0.2/lib/parser/runner.rb:181:in `process_files'
        from /home/mbj/.gem/ruby/2.3.0/gems/parser-2.3.0.2/lib/parser/runner.rb:159:in `block in process_all_input'
        from /home/mbj/.rubies/ruby-2.3.0/lib/ruby/2.3.0/benchmark.rb:293:in `measure'
        from /home/mbj/.gem/ruby/2.3.0/gems/parser-2.3.0.2/lib/parser/runner.rb:157:in `process_all_input'
[...]

PR #254, opened 2016-01-25
Lexer emits string/symbol literals in source file encoding

alexdowad:
Previously, even if you had an encoding comment like:

    # encoding: us-ascii

...The string/symbol literals created by the lexer would be UTF-8 encoded.

### DIFF
--- a/lib/parser/source/buffer.rb
+++ b/lib/parser/source/buffer.rb
@@ -95,9 +95,7 @@ module Parser
         elsif detected_encoding == Encoding::BINARY
           input
         else
-          input.
-            force_encoding(detected_encoding).
-            encode(Encoding::UTF_8)
+          input.force_encoding(detected_encoding)
         end
       end
 
--- a/test/test_parser.rb
+++ b/test/test_parser.rb
@@ -4976,10 +4976,11 @@ class TestParser < Minitest::Test
     end
 
     def test_magic_encoding_comment
+      funny_russian_sym = "проверка".encode(Encoding::KOI8_R).to_sym
       assert_parses(
         s(:begin,
-          s(:lvasgn, :"проверка", s(:int, 42)),
-          s(:send, nil, :puts, s(:lvar, :"проверка"))),
+          s(:lvasgn, funny_russian_sym, s(:int, 42)),
+          s(:send, nil, :puts, s(:lvar, funny_russian_sym))),
         %Q{# coding:koi8-r
            \xd0\xd2\xcf\xd7\xc5\xd2\xcb\xc1 = 42
            puts \xd0\xd2\xcf\xd7\xc5\xd2\xcb\xc1}.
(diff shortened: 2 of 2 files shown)


test/test_parser.rb
```diff
@@ -4976,10 +4976,11 @@ def test_bom
     end
 
     def test_magic_encoding_comment
+      funny_russian_sym = "проверка".encode(Encoding::KOI8_R).to_sym
       assert_parses(
         s(:begin,
-          s(:lvasgn, :"проверка", s(:int, 42)),
-          s(:send, nil, :puts, s(:lvar, :"проверка"))),
+          s(:lvasgn, funny_russian_sym, s(:int, 42)),
+          s(:send, nil, :puts, s(:lvar, funny_russian_sym))),
```
whitequark: ?

test/test_parser.rb
```diff
@@ -5336,4 +5337,21 @@ def test_bug_regex_verification
...
+
+  def test_bug_non_utf8_symbol_literal
+    # "\xC7" is actually invalid in US-ASCII, but Ruby allows the conversion of
+    # this invalid string to a symbol, regardless if you do so using
+    # `String#to_sym` or if you have :"\xC7" appear as a literal in a US-ASCII
+    # encoded source file.
+    #
+    # If it was UTF-8, we would get an EncodingError.
+
+    bad_symbol = "\xC7".force_encoding(Encoding::US_ASCII).to_sym
```
whitequark: This is actually incorrect, because:
```
# -*- encoding: us-ascii -*-

p :"\xC3".to_s # "\xC3"
p :"\xC3".to_s.valid_encoding? # true
p :"\xC3".to_s.encoding # #<Encoding:ASCII-8BIT>
```

lib/parser/source/buffer.rb
```diff
@@ -95,9 +95,7 @@ def self.reencode_string(input)
         elsif detected_encoding == Encoding::BINARY
           input
         else
-          input.
-            force_encoding(detected_encoding).
-            encode(Encoding::UTF_8)
+          input.force_encoding(detected_encoding)
```
whitequark: This will fail if a file with a non-ASCII-compatible encoding is ever parsed. Or something with ShiftJIS, IIRC.

test/test_parser.rb
alexdowad: You are right!

test/test_parser.rb
alexdowad: The test file it itself encoded in UTF-8. The only way I know to get a symbol in a different encoding is to convert a string over manually and then `to_sym` it.

lib/parser/source/buffer.rb
alexdowad: The PR has been updated. But there are other questions to resolve first; see my comments in the GH issue.

test/test_parser.rb
whitequark: Oh right, I missed the reencode into KOI8-R.

whitequark:
Closed per discussion in #252.

PR #256, opened 2016-01-28
Make Buffer#line_begins fly

alexdowad:
The existing code uses `String#each_char`, which allocates a new string object for **each** character in the source file. (This is the #2 source of object allocations in RuboCop.) It also uses a lot of (slow) Ruby method calls.

Switching to `String#index` increases the performance of `Source::Buffer#line_begins` by a factor of more than 20.

### DIFF
--- a/lib/parser/source/buffer.rb
+++ b/lib/parser/source/buffer.rb
@@ -253,14 +253,11 @@ module Parser
 
       def line_begins
         unless @line_begins
-          @line_begins, index = [ [ 0, 0 ] ], 1
+          @line_begins, index = [ [ 0, 0 ] ], 0
 
-          @source.each_char do |char|
-            if char == NEW_LINE
-              @line_begins.unshift [ @line_begins.length, index ]
-            end
-
+          while index = @source.index("\n".freeze, index)
             index += 1
+            @line_begins.unshift [ @line_begins.length, index ]
           end
         end
 


lib/parser/source/buffer.rb
```diff
@@ -253,14 +253,11 @@ def last_line
...
         unless @line_begins
-          @line_begins, index = [ [ 0, 0 ] ], 1
-
-          @source.each_char do |char|
-            if char == NEW_LINE
-              @line_begins.unshift [ @line_begins.length, index ]
-            end
+          @line_begins, index = [ [ 0, 0 ] ], 0
 
+          while index = @source.index("\n".freeze, index)
```
alexdowad: `"string literal".freeze` compiles to a special opcode in MRI, which uses a single, shared object. It doesn't actually call `Object#freeze` every time.

whitequark: We have a lot of those sprinkled all over the codebase--we need to fix them either to this new form or the old form. @bbatsov, how much do we care about the versions of Ruby without this optimization?

alexdowad:
Wow, I didn't expect this. On JRuby, `"string".freeze` is consistently faster than a constant lookup.

alexdowad:
On RBX, a constant lookup is faster.

alexdowad:
From MRI 2.1+, there is no difference. On MRI <= 2.0, a constant lookup is faster. (Likely this optimization was introduced in 2.1.)

whitequark:
@alexdowad Please rip out const lookups then.

alexdowad:
> Please rip out const lookups then.

Have a look at the `use_str_literal_freeze` branch in this repo. Looks good?

whitequark:
Yep

alexdowad:
Merged.

alexdowad:
A few more selected literals converted to use shared objects... see `avoid_more_allocations` branch. You like?

@whitequark Do you think that the strings which are emitted as part of lexer tokens should always be mutable, or would it be acceptable for them to be frozen as well? This would be a small change to the observable behavior of the library, but would also help to reduce object allocations.

PR #259, opened 2016-01-28
Source::Rewriter handles clobbering correctly for insertions

alexdowad:
Previously, insertions ("replacements" of a zero-length range) would never
clobber or be clobbered by anything, even if they landed right in the middle
of a replaced or deleted range of text. As can easily be seen, this makes a
mess and should give rise to a clobbering error.

The tricky thing is that an insertion can legitimately appear at the very
beginning or very end of a replaced or deleted range. If it is in the middle,
though, that is definitely wrong and should cause an error. (Unless it appears
in the middle of a replaced range, and the replacement is longer than the
replaced text by the length of the insertion or more, and the replacement has
the same effect as the insertion.)

Additionally, two insertions at the very same point should also clobber each
other. Why? Well, how do we know which inserted text should come before the
other? We don't. And if the inserted texts are identical, how do we know if
the intention was to add one or two copies of that text? Again, we don't. It's
safer to just flag this as clobbering.

Additionally, there were some cases where the previous code would merge edits
which should really have raised a clobbering error. It is now more robust all
around, and close to 30 new unit tests have been added to verify its behavior.

### DIFF
--- a/lib/parser/source/rewriter.rb
+++ b/lib/parser/source/rewriter.rb
@@ -5,7 +5,8 @@ module Parser
     # {Rewriter} performs the heavy lifting in the source rewriting process.
     # It schedules code updates to be performed in the correct order and
     # verifies that no two updates _clobber_ each other, that is, attempt to
-    # modify the same part of code.
+    # modify the same section of code. (However, if two updates modify the
+    # same section in exactly the same way, they are merged.)
     #
     # If it is detected that one update clobbers another one, an `:error` and
     # a `:note` diagnostics describing both updates are generated and passed to
@@ -37,6 +38,7 @@ module Parser
         @source_buffer = source_buffer
         @queue         = []
         @clobber       = 0
+        @insertions    = 0 # clobbered zero-length positions; index 0 is the far left
       end
 
       ##
@@ -100,9 +102,7 @@ module Parser
         adjustment = 0
         source     = @source_buffer.source.dup
 
-        sorted_queue = @queue.sort_by.with_index do |action, index|
-          [action.range.begin_pos, index]
-        end
+        sorted_queue = @queue.sort_by { |action| action.range.begin_pos }
 
         sorted_queue.each do |action|
           begin_pos = action.range.begin_pos + adjustment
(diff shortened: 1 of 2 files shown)


alexdowad:
New minor release?

JoshCheek:
Hi, I use the feature of inserting at the same spot multiple times, approximately like this:

```ruby
require 'parser/ruby23'

buffer   = Parser::Source::Buffer.new("example").tap { |b| b.source = '2' }
rewriter = Parser::Source::Rewriter.new(buffer)

range    = Parser::Source::Range.new(buffer, 0, 0)
rewriter.insert_before(range, "(")
rewriter.insert_before(range, "[")

range    = Parser::Source::Range.new(buffer, 1, 1)
rewriter.insert_after(range, "]")
rewriter.insert_after(range, ")")

rewriter.process # => "([2])"
```

This is for [SeeingIsBelieving](https://github.com/JoshCheek/seeing_is_believing), which has to wrap every line of source code ([example](https://github.com/JoshCheek/seeing_is_believing/blob/00d0651b4ae2adce9e15c832aa503c69ef74f75a/lib/seeing_is_believing/wrap_expressions.rb#L22-L42)). Is there a suggested solution? I'll try playing around with it in the interim.

> Additionally, two insertions at the very same point should also clobber each
> other. Why? Well, how do we know which inserted text should come before the
> other? 

I determined it either experimentally or by reading the source (it was a couple years ago), then wrote my code to apply insertions in the order that I observed.

whitequark:
I feel like ultimately @alexdowad's fixes nudge Parser in a direction that provides more correctness. I certainly never put enough thought about the behavior of multiple insertions at the same point (or else it would be documented proper), and you are relying on an implementation detail that's spilling through. The problem though is that your use case is also valid, and to work around it you'll have to keep some kind of intermediate map around... which is basically what the Rewriter already is.

Maybe make the old behavior explicit and opt-in? `rewriter.insert_before_multi` that inserts before the position before all other inserted chunks, and `rewriter.insert_after_multi` that inserts after the position after all other inserted chunks. Should be easy enough to implement...

JoshCheek:
That would be nice. Currently, it looks like this can probably support my use cases:

```
require 'parser/ruby23'

buffer   = Parser::Source::Buffer.new("example").tap { |b| b.source = '12' }
rewriter = Parser::Source::Rewriter.new(buffer)

range_first  = Parser::Source::Range.new(buffer, 0, 1)
range_second = Parser::Source::Range.new(buffer, 1, 2)
range_full   = Parser::Source::Range.new(buffer, 0, 2)
rewriter.insert_before(range_full, "<")
rewriter.insert_before(range_first, "(")
rewriter.insert_after(range_first, ")")
rewriter.insert_before(range_second, "[")
rewriter.insert_after(range_second, "]")
rewriter.insert_after(range_full, ">")

rewriter.process # => "<(1)[2]>"
```

I'll see how hard it is to retain/create the full range wherever I do insertions. Hopefully not too hard, but there are some edge cases I try to handle (eg `__END__`) that could be difficult.

alexdowad:
> I feel like ultimately @alexdowad's fixes nudge Parser in a direction that provides more correctness.

I agree. Quite a bit of thought went into this issue when redesigning `Source::Rewriter`.

If someone writes straight-line code (like @JoshCheek's) which performs multiple insertions at the same point, presumably they know what they are doing. This is not true when you have multiple independent transforms applied to a source text, and 2 of them happen to insert at the same point. The fact that one insertion happened to be performed before the other means nothing. The only safe thing to do is to raise a clobbering error.

This is not a hypothetical problem. In RuboCop, speculatively performing independent transforms simultaneously (and backing out if there is a clobbering error) makes enormous speedups possible for auto-correction. But before the `Rewriter` overhaul, it would mangle your code beyond recognition.

If we want to optionally allow multiple insertions at the same point, `insert_before` and `insert_after` will have to take on more meaning than they currently have. Right now they both insert the same type of `Action` in the action queue, but that will not work any more.

Why? Think of a situation where someone does:

```ruby
rewriter.insert_before(range, "(")
rewriter.insert_after(range, ")")
```

Followed by:

```ruby
rewriter.insert_before(range, "[")
rewriter.insert_after(range, "]")
```

This is a perfectly reasonable thing to do. First insert a balanced pair of parens, then insert a balanced pair of brackets. But currently this would lead to:

```([ ... code here )]```

Since the user said `insert_before(..., "(")` followed by `insert_before(..., "["), presumably the opening bracket should come *before* the opening paren. But again, since they said `insert_after(..., ")")` followed by `insert_after(..., "]")`, again, logically the closing bracket should come *after* the closing paren.

IN SHORT:

If you want to allow multiple insertions at the same point, inserting "after" must be distinguished from inserting "before". This is a bit different from the old behavior, but is more logical and consistent.

JoshCheek:
I'm not really sure how to work with the current rewriter, after playing for a bit, I realized I do need to insert at the same spot multiple times. Eg I need to be able to wrap each line and also the whole document. But if the whole document is just one line, then the range is the same for both, which raises a clobbering error. I suppose I could write after each insertion, or write an algorithm that does it more intelligently, but it could be expensive, and at the very least, is a certain amount of effort to regain functionality I used to have.

I also need to be able to parse invalid files so I can strip comments I've added, even mid-edit ([example](https://github.com/JoshCheek/seeing_is_believing/blob/00d0651b4ae2adce9e15c832aa503c69ef74f75a/features/flags.feature#L224-L228)) That's probably a separate issue on a separate commit, but thought I'd mention it here, as the conjunction of these two currently implies I need to restrict my parser dependency to 2.2.2.6 I'll have to look at it more closely, though, I haven't given enough love to this project recently, so I'm re-discovering previous knowledge. 

You can see an example of how I use it while teaching [here](https://vimeo.com/154120595), I also use it quite often, for myself, over a REPL.

whitequark:
@JoshCheek It should be trivial to add the `insert_{before,after}_multi` methods that I talk about (make them just like `insert_{before,after}` but without the clobbers), and I'll accept a PR of course.

JoshCheek:
FWIW, the reordering of insertions wound up being a more trouble than I initially anticipated. Have to do a weird 3-phase insertion now: https://github.com/JoshCheek/seeing_is_believing/compare/252911d...e73fced#diff-12

whitequark:
That looks about right to me?

PR #261, opened 2016-01-29
Use more shared strings in lexer, including those emitted with tokens

alexdowad:


alexdowad:
When Ruby 3.0 is here we can drop a whole bunch of `.freeze` calls...

lib/parser/lexer.rl
```diff
@@ -1668,11 +1668,11 @@ class Parser::Lexer
 
         indent      = !$1.empty? || !$2.empty?
         dedent_body = !$2.empty?
-        type        =  '<<' + ($3.empty? ? '"' : $3)
+        type        =  '<<'.freeze + ($3.empty? ? '"'.freeze : $3)
```
whitequark: I think the result needs to be frozen, not constituents

alexdowad:
Updated based on comment from @whitequark.

lib/parser/lexer.rl
```diff
@@ -1668,20 +1668,17 @@ class Parser::Lexer
...
           p = @ts + 1
           fnext expr_beg; fbreak;
         else
           fnext *push_literal(type, delimiter, @ts, heredoc_e, indent, dedent_body);
 
-          if @herebody_s.nil?
-            @herebody_s = new_herebody_s
-          end
-
+          @herebody_s ||= new_herebody_s
```
whitequark: These are not generally equivalent... I guess it's fine here

whitequark:
You broke the build on 2.3. How about you test the changes beforehand?

alexdowad:
> How about you test the changes beforehand?

@whitequark, I do test changes before opening a PR. However, it would be best to wait until the build passes before merging, since I don't test on all 11 supported interpreters.

This is what happened on Travis CI, Ruby 2.3.0:

```sh
$ git fetch origin +refs/pull/261/merge:
fatal: Couldn't find remote ref refs/pull/261/merge
Unexpected end of command stream
The command "eval git fetch origin +refs/pull/261/merge:" failed. Retrying, 2 of 3.
fatal: Couldn't find remote ref refs/pull/261/merge
Unexpected end of command stream
The command "eval git fetch origin +refs/pull/261/merge:" failed. Retrying, 3 of 3.
fatal: Couldn't find remote ref refs/pull/261/merge
Unexpected end of command stream
The command "eval git fetch origin +refs/pull/261/merge:" failed 3 times.
The command "git fetch origin +refs/pull/261/merge:" failed and exited with 128 during .
```

As you can see, the failure had nothing to do with the code in the PR.

These kind of intermittent failures on Travis CI seem to be getting more and more frequent on various projects I contribute to. Perhaps they are having trouble scaling up to increasing demand?

whitequark:
@alexdowad Uhhhh, no. Please see this log: https://travis-ci.org/whitequark/parser/jobs/105838195#L216

I wouldn't complain if it was something obscure like Ruby 1.8, but in this case it is the very Ruby version that is poised to benefit from your changes that was broken.

alexdowad:
Thank you for the link. The one I clicked on before displayed something different, don't know why.

Sorry for this mistake. I have opened a PR which fixes the problem.

PR #263, opened 2016-02-01
Test emitted ranges

alexdowad:
When working on optimization of `Range#line`, I looked at what tests we have for ranges emitted by the lexer and was surprised to discover that the test suite hardly checks anything at all as re: correctness of the lexed ranges.

Perhaps unsurprisingly, when I tightened the test suite up to check the begin and end of each emitted range, it led me to discover a (relatively minor) bug.

@whitequark, see if you like the way the ranges are tested here or if you would prefer something else.

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -2027,7 +2027,7 @@ class Parser::Lexer
         end
 
         if version?(18, 19, 20)
-          emit(:tINTEGER, digits.to_i(@num_base))
+          emit(:tINTEGER, digits.to_i(@num_base), @ts, @num_suffix_s)
           p = @num_suffix_s - 1
         else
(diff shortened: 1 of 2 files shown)


whitequark:
Can we just test those in the parser tests? Lexer tests aren't even close to exhaustive anyway.

alexdowad:
> Can we just test those in the parser tests?

Oh, I see... you have those things like `^^ begin` and `~~~~~ expression` and so on.

I think the value of checking in the lexer tests as well is shown by the bug which I discovered this way. But it's up to you.

whitequark:
Hrm, this is actually not possible to test in `test_parser.rb` because none of those result in valid parsetrees. Conversely, does it matter a lot then?..

whitequark:
Well, I guess it doesn't hurt.

PR #264, opened 2016-02-02
Cache line/column numbers for fast lookup in Source::Buffer

### LINKED ISSUE #260: Cool Range#line down
`Range#line` is a very significant hotspot for RuboCop (about 10% of runtime when inspecting code).

Probably the best way to cool it down would be for the lexer to track line numbers and pass a line number to `Range.new` when initializing ranges. Since the lexer already treats newlines specially in most cases, code could be added to increment a line number count.

Another option would be to cache `Range#line`. The problem here is that ranges are frozen after initialization.

Thoughts?

alexdowad:
This halves the time which RuboCop spends in `Range#line`.

...Which will probably not be enough, eventually. But for now, it's good. That's why I annotated the added methods as `@api private` -- we may come up with something better later.

(It turns out that tracking the line numbers in the lexer is even more difficult than I thought. The handling of newlines would have to be completely reworked, and I have better things to work on.)

Closes #260.

### DIFF
--- a/lib/parser/source/buffer.rb
+++ b/lib/parser/source/buffer.rb
@@ -106,6 +106,10 @@ module Parser
 
         @lines       = nil
         @line_begins = nil
+
+        # Cache for fast lookup
+        @line_for_position = {}
+        @col_for_position  = {}
       end
 
       ##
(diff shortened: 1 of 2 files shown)


lib/parser/source/buffer.rb
```diff
@@ -189,6 +193,34 @@ def decompose_position(position)
...
+
+      ##
+      # Convert a character index into the source to a column number.
+      #
+      # @param  [Integer] position
+      # @return [Integer] column
+      # @api private
+      #
+      def column_for_position(position)
+        @col_for_pos[position] ||= begin
```
whitequark: `@column_for_position`

whitequark:
LGTM otherwise

alexdowad:
Updated. I've only tested on MRI 2.2.2 and 2.3.0, merge at your own risk!

whitequark:
Travis passed on 1.8.

PR #265, opened 2016-02-02
Testing UTF-32LE

alexdowad:
In one of my other PRs, @whitequark suggested that the parser tests should be run with the `force_utf32` flag set on the lexer. This amends the test suite to do just that; actually, it runs the parser tests both with and without `force_utf32`.

This immediately revealed that some of my optimizations to avoid string allocations were not sound, as @whitequark earlier said. The ones which break the test suite have been backed out.

Additionally, another bug was revealed: parsing escapes can break the lexer when `force_utf32` is on. The reason is that hash lookup in `ESCAPES` doesn't work when the source has been transcoded to UTF-32LE, because the keys of `ESCAPES` are not in that encoding. Then `ESCAPES.fetch(char, char)` returns the escaped char (which could be `"n"` or anything else), and it is appended to the current literal buffer, which fails with an incompatible encoding error.

Please have a look and see if you think the fix is sound.

(A bug was just reported against RC the other day which looks like this exact same bug... https://github.com/bbatsov/rubocop/issues/2751)

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -82,10 +82,10 @@ class Parser::Lexer
   # %
 
   ESCAPES = {
-    'a' => "\a", 'b'  => "\b", 'e'  => "\e", 'f' => "\f",
-    'n' => "\n", 'r'  => "\r", 's'  => "\s", 't' => "\t",
-    'v' => "\v", '\\' => "\\"
-  }
+    ?a.ord => "\a", ?b.ord  => "\b", ?e.ord => "\e", ?f.ord => "\f",
+    ?n.ord => "\n", ?r.ord  => "\r", ?s.ord => "\s", ?t.ord => "\t",
+    ?v.ord => "\v", ?\\.ord => "\\"
+  }.freeze
 
   REGEXP_META_CHARACTERS = Regexp.union(*"\\$()*+.<>?[]^{|}".chars).freeze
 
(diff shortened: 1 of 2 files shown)


alexdowad:
I should mention another (strange) thing. The test failures related to `ESCAPES` *only* occurred on the dedented heredoc tests for 2.3. Both of the failing tests used interpolated, dedented heredocs (`<<~"END"`). Why should escape handling be any different for them than for any other interpolated string?? I haven't figured that out.

lib/parser/lexer.rl
```diff
@@ -82,10 +82,11 @@ class Parser::Lexer
   # %
 
   ESCAPES = {
-    'a' => "\a", 'b'  => "\b", 'e'  => "\e", 'f' => "\f",
-    'n' => "\n", 'r'  => "\r", 's'  => "\s", 't' => "\t",
-    'v' => "\v", '\\' => "\\"
-  }
+    'a'[0].ord  => "\a", 'b'[0].ord => "\b", 'e'[0].ord => "\e",
```
whitequark: `?a`

alexdowad:
Updated.

alexdowad:
@whitequark So what's your call on this one? 9d0c97a definitely fixes a bug. I'm not sure if there may be a better fix possible, but if so, we can always replace this fix with a better one when it is found.

alexdowad:
Travis is also green.

whitequark:
I was busy with an international move.

PR #269, opened 2016-02-06
Improve performance parsing non-ASCII input.

### LINKED ISSUE #268: Parser is extremely slow for files containing *any* unicode
I ran into a bit of a nasty performance bug with the parser gem while trying to understand why `rubocop` was so slow on some of my source files. I'm genuinely unsure if there's a good solution, but I wanted to document it in an issue in case you had any thoughts.

`parser` relies heavily on taking slices of the source buffer at character offsets, most notably in `Parser::Lexer#tok` and `Parser::Source::Range#source`. However, if an input file contains *any* non-ASCII characters, Ruby implements character slices by walking the string codepoint-at-a-time, since it has to treat multibyte utf-8 sequences as a single character.

One one 1500-line source file, `ruby-parse FILE > /dev/null` takes 2.5s; If I remove the two lines containing unicode characters, it immediately speeds up to 700ms.

If I look at the performance using Linux's `perf` tool, we can see that about 75% of the time (on the unicode version) is spent on simple string operations:
```
 48.33%  ruby-parse  libruby.so.2.1.0    [.] str_strlen
 27.20%  ruby-parse  libruby.so.2.1.0    [.] str_utf8_nth
```

In addition to being slow in absolute terms, this behavior makes parser quadratic in file length, since it will be repeatedly seeking to location `n` in the file, for ranges of `n` ranging from 0 to `len(file)`, as it parses.

nelhage:
If the input contains high-bit codepoints, always transcode to UTF-32,
to ensure O(1) string slicing. This prevents quadratic behavior, because
string indexing and slicing is O(n) on UTF-8-encoded strings. (An
internal Ruby optimization preserves constant-time access if the string
is UTF-8-encoded but happes to only contain ASCII characters.

fixes #268

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -211,8 +211,14 @@ class Parser::Lexer
         # symbols, then storing the source in UTF-8 would be faster.
         #
         # Patches accepted.
-        @source = @source.encode(Encoding::UTF_32LE)
-        @need_encode = true
+        begin
+          @source = @source.encode(Encoding::UTF_32LE)
+          @need_encode = true
+        rescue EncodingError
+          # We may receive invalid UTF-8, which we will be able to
+          # re-encode. Just carry on in UTF-8, which will happily skip
+          # over invalid byte sequences and leave them as-is.
+        end
       end
 
       if @source_pts[0] == 0xfeff
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -198,7 +198,7 @@ class Parser::Lexer
       end
 
       if @has_encode &&
-        (@source_pts.size > 1_000_000 || @force_utf32) &&
+        (!@source.ascii_only? || @force_utf32) &&
         @encoding != Encoding::UTF_32LE
         # A heuristic: if the buffer is larger than 1M, then
         # store it in UTF-32 and convert the tokens as they're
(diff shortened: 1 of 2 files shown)


alexdowad:
Nice clean, concise commit log message.

nelhage:
Looking into failing tests.

nelhage:
(also I should update the comment)

nelhage:
Do you I squash the git history, or leave the bugfix commits around?

whitequark:
Squash please.

whitequark:
Hahahahaha I can't believe you're the person behind accidentallyquadratic

nelhage:
@whitequark This *keeps happening* to me! Planning to write this one up once a solution lands :)

nelhage:
Do either of you have opinions on the test case in fcedf47? The `instance_variable_get` is unpleasant, but by construction this patch should not affect the `Lexer` interface at all (other than performance), so I'm not sure how else to test it.

nelhage:
Is the 2.1.8 failure on Travis a known issue? It doesn't look related, at a glance.

whitequark:
> Do either of you have opinions on the test case in fcedf47?

I'm OK with it.

> Is the 2.1.8 failure on Travis a known issue?

Yes.

whitequark:
Thanks!

nelhage:
Ah, was gonna squash once I was sure it was passing, but thanks! That was an unexpectedly helpful and swift response to a thorny issue.

whitequark:
Whoops, assumed you squashed it but GitHub didn't update the PR. Oh well.

PR #270, opened 2016-02-06
Generalize the unicode performance fixes.

nelhage:
@whitequark how do you feel about this, or something like this? The last
PR essentially fixed the parser performance, but rubocop was still slow
becaue of repeated slicing of source ranges. I don't have an intuition
for whether the extra cost of indirecting through @source_buffer is too
expensive -- I realize the lexer is some of the hottest code here.

---

This is an attempt to generalize the performance gains of #269 to extend
to clients of the resulting AST.

The immediate problem is that many AST consumers (rubocop is the one I'm
currently interested in) are going to invoke
`Parser::Source::Range#source` on many of the AST nodes, and that method
has the same linear-time-slicing behavior as `Parser::Lexer#tok` did.

By pushing the re-encoding inside the `Parser::Source::Buffer`
abstraction, we make it also available to `Range`, and we simplify
`Lexer` while we're at it: A bunch of encoding bookkeeping is now hidden
inside the `Buffer#slice` abstraction.

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -19,7 +19,7 @@
 #    they're pointing to ("current"), plus one. `@ts` contains the index
 #    of the corresponding character. The code for extracting matched token is:
 #
-#       @source[@ts...@te]
+#       @source_buffer.slice(@ts...@te)
 #
 #  * If your input is `foooooooobar` and the rule is:
 #
@@ -109,8 +109,6 @@ class Parser::Lexer
     @tokens     = nil
     @comments   = nil
 
-    @has_encode = ''.respond_to?(:encode)
-
     reset
   end
 
@@ -128,10 +126,7 @@ class Parser::Lexer
 
     @force_utf32   = false # Set to true by some tests
 
-    @source        = nil # source string
     @source_pts    = nil # @source as a codepoint array
-    @encoding      = nil # target encoding for output strings
-    @need_encode   = nil
 
     @p             = 0   # stream position (saved manually in #advance)
     @ts            = nil # token start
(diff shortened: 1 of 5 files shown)


lib/parser/lexer.rl
```diff
@@ -334,22 +302,17 @@ class Parser::Lexer
...
     def encode_escape(ord)
       ord.chr
     end
+  end
 
-    def tok(s = @ts, e = @te)
-      @source[s...e]
-    end
+  def tok(s = @ts, e = @te)
+    @source_buffer.slice(s...e)
```
whitequark: I have an even better idea. How about `[@source_pts.slice(s...e)].pack('U*')` ?

alexdowad: I tried that; it's much slower than `String#[]`.

lib/parser/source/buffer.rb
```diff
@@ -178,6 +181,23 @@ def raw_source=(input)
         end
 
         @source = input.gsub("\r\n".freeze, "\n".freeze).freeze
+
+        if defined?(Encoding) &&
+           !@source.ascii_only? &&
+           @source.encoding != Encoding::UTF_32LE
+          begin
+            @slice_source = @source.encode(Encoding::UTF_32LE)
+          rescue EncodingError
```
whitequark: Too defensive. This shouldn't be silently swallowed here, in fact any encoding problem that may arise here should have been caught in `reencode_string`.

lib/parser/lexer.rl
whitequark: Ack.

lib/parser/source/buffer.rb
```diff
@@ -178,6 +181,23 @@ def raw_source=(input)
...
+          end
+        end
+      end
+
+      def slice(i)
+        if @slice_source
```
whitequark: `@slice_source.nil?`

whitequark: Also, not `i` but `range` or something, since this accepts not just indexes.

whitequark:
Other than comments above, LGTM.

lib/parser/source/buffer.rb
nelhage: If I leave that out, I get a failure in `test_lexer.rb`:

```
  1) Error:
TestLexer#test_bug_string_non_utf:
Encoding::UndefinedConversionError: "\xE9" to UTF-8 in conversion from ASCII-8BIT to UTF-8 to UTF-32LE
    /Users/nelhage/stripe/parser/lib/parser/source/buffer.rb:189:in `encode'
    /Users/nelhage/stripe/parser/lib/parser/source/buffer.rb:189:in `raw_source='
    /Users/nelhage/stripe/parser/lib/parser/source/buffer.rb:168:in `source='
    /Users/nelhage/stripe/parser/test/test_lexer.rb:84:in `assert_scanned'
    /Users/nelhage/stripe/parser/test/test_lexer.rb:3180:in `test_bug_string_non_utf'
```

I think the problem is that `ASCII-8BIT` can't be re-encoded to UTF-32; Should I add an explicit check for whether the input is `ASCII-8BIT`, instead?

whitequark:
Thanks! This was a very longstanding issue (performance bug, really).

Issue #273, opened 2016-02-13
parser blows up on syntax error


troessner:
```
 ( $ ) ruby-parse -v
ruby-parse based on parser version 2.3.0.5
```

```
  ( $ ) cat bad_syntax.rb
class Create
  def self.

  end
end
```

```
  ( $ ) ruby-parse bad_syntax.rb
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.4-compliant syntax, but you are running 2.2.3.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
Failed on: bad_syntax.rb
/Users/timo/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/parser-2.3.0.5/lib/parser/diagnostic.rb:121:in `[]=': 1...1 out of range (RangeError)
	from /Users/timo/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/parser-2.3.0.5/lib/parser/diagnostic.rb:121:in `render_line'
	from /Users/timo/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/parser-2.3.0.5/lib/parser/diagnostic.rb:96:in `render'
	from /Users/timo/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/parser-2.3.0.5/lib/parser/runner.rb:147:in `block in prepare_parser'
	from /Users/timo/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/parser-2.3.0.5/lib/parser/diagnostic/engine.rb:67:in `call'
	from /Users/timo/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/parser-2.3.0.5/lib/parser/diagnostic/engine.rb:67:in `process'
	from /Users/timo/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/parser-2.3.0.5/lib/parser/base.rb:251:in `on_error'
	from /Users/timo/.rbenv/versions/2.2.3/lib/ruby/2.2.0/racc/parser.rb:258:in `_racc_do_parse_c'
	from /Users/timo/.rbenv/versions/2.2.3/lib/ruby/2.2.0/racc/parser.rb:258:in `do_parse'
[...]

PR #276, opened 2016-02-25
Bugfix for problem mentioned in #273

alexdowad:


test/test_diagnostic.rb
```diff
@@ -69,4 +69,28 @@ def test_multiline_render
...
+    @buffer.source = source
+
+    location = Parser::Source::Range.new(@buffer, 33, 34)
+    diag = Parser::Diagnostic.new(:error, :unexpected_token, { :token => 'tNL' },
+                                  location)
+
+    assert_equal([
+      '(string):2:32-3:0: error: unexpected token tNL',
+      '(string):2:   foo: ->() # I forgot my brace',
+      '(string):2:                                ',
```
whitequark: There should really be a `^` under the newline.

alexdowad: One character past the end of `brace`, in other words?

Makes sense... I don't have time to look into this right now (unfortunately), but when I do, I can fix it.

alexdowad: A clarifying note: this is the actual output which is printed to the console right now by `ruby-parse`.

whitequark: I'll do it myself.

alexdowad:
Thanks @whitequark. By the way, I am wondering whether `"\n"` should really be considered to span 2 lines. (As well as `"some text\n"`.) If so, I think we should ensure that the multiline diagnostic format is not used in these situations.

whitequark:
I did that.

PR #277, opened 2016-02-26
Add Source::Rewriter APIs for multiple insertions at the same position

yujinakayama:
For https://github.com/whitequark/parser/pull/259#issuecomment-184514530

### DIFF
--- a/lib/parser/source/rewriter.rb
+++ b/lib/parser/source/rewriter.rb
@@ -39,6 +39,9 @@ module Parser
         @queue         = []
         @clobber       = 0
         @insertions    = 0 # clobbered zero-length positions; index 0 is the far left
+
+        @insert_before_multi_order = 0
+        @insert_after_multi_order = 0
       end
 
       ##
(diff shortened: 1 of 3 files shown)


alexdowad:
@yujinakayama It's nice that you are working on this. I think the algorithm used is not totally sound, but can't think of a correct one right now.

Another issue: It appears that with this patch, a "multi" insertion will *never* clobber, or be clobbered by, any other edit. That is not correct. A "multi" insertion should not clobber *other insertions*, but (for example) a deletion which spans both sides should still clobber it.

yujinakayama:
> Another issue: It appears that with this patch, a "multi" insertion will never clobber, or be clobbered by, any other edit. That is not correct. A "multi" insertion should not clobber other insertions, but (for example) a deletion which spans both sides should still clobber it.

The multi-insertion properly clobber or be clobbered by other non-insertion actions with the current implementation. I've added specs for those cases.

lib/parser/source/rewriter.rb
```diff
@@ -65,6 +68,28 @@ def insert_before(range, content)
...
 
       ##
+      # Inserts new code before the given source range by allowing other
+      # insertions at the same position.
+      # Note that an insertion with latter invocation comes _before_ earlier
+      # insertion at the same position in the rewritten source.
+      #
+      # @example Inserting '[('
+      #   rewriter.
+      #     insert_after_multi(range, '(').
```
whitequark: `insert_before_multi`

yujinakayama: Fixed

yujinakayama:
Can you cut a release for this please?

whitequark:
@yujinakayama I assume you've tested this with Rubocop?

@JoshCheek Can you please test these changes with your projects and tell if you find them suitable?

yujinakayama:
@whitequark I tested with Transpec, since it does almost the same approach (for analyzing runtime value) with @JoshCheek's SeeingIsBelieving and has issue with the latest parser without the multi-insertion API.

whitequark:
I'd like to get a bit more testing out of this API before committing to support it.

JoshCheek:
I'll try to switch over, but might not have time until next weekend, just super busy right now.

JoshCheek:
Mostly painless ([updates](https://github.com/JoshCheek/seeing_is_believing/commit/4d22850fe75200807bacbb75abe9a4e666ee94fc)). Though note that behaviour changed for inserting before. Not really an issue, just different.

```
[<A>] # old
<[A>] # new
```

whitequark:
You can invoke the lexer alone to extract the comments. That is a supported use case.

PR #288, opened 2016-05-05
Pass --no-line-convert to Racc to prevent wrapping generated parser code with module_eval.

iliabylich:
Required for https://github.com/opal/opal/pull/1465

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,8 @@ test/tmp
 test/version_tmp
 tmp
 *.output
+.ruby-version
+.ruby-gemset
 lib/parser/lexer.rb
 lib/parser/ruby18.rb
 lib/parser/ruby19.rb
--- a/Rakefile
+++ b/Rakefile
@@ -148,6 +148,7 @@ rule '.rb' => '.y' do |t|
            t.source,
            "-o", t.name
          ]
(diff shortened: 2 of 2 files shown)


whitequark:
Needs to happen only in debug mode.

iliabylich:
I thought it would be nice for opal to use a parser gem directly. But to do this we need parser gem to not wrap its methods with `module_eval`.  I can copy the grammar to the opal repo, but it requires us to sync it with the parser gem from time to time. Does racc (and parser) really need those `module_eval`-s? Does anyone use generated file names/method lines?

Let me explain a problem.
1. Opal uses parser to convert source code to ast
2. Opal implements methods like `eval/module_eval` through its library `opal-parser`
3 `opal-parser` is just a compiled version of opal's compiler
4. To compile `opal-parser` and `parser` we need `parser` to be as simple as possible
5. Right now compiled `parser/ruby23` contains a lot of `module_eval` invocations

Is it possible to remove `module_eval`-s from parser's parser without re-generating it locally?

whitequark:
Like I said, I want `module_eval`s to be generated in debug builds of parser. In release builds they aren't really necessary.

iliabylich:
@whitequark Done I guess

iliabylich:
Thank you!

PR #299, opened 2016-06-07
- Parser::AST::Processor: do not add nil to the end of required argument.

### LINKED ISSUE #298: Parser::AST::Processor modifies sexp
``` ruby
require 'ast'
require 'parser/ruby23'

sexp = Parser::Ruby23.parse("def m(a); end")
p sexp

processed = Parser::AST::Processor.new.process(sexp)
p processed
```
Output:
``` ruby
s(:def, :m,
  s(:args,
    s(:arg, :a)), nil)
s(:def, :m,
  s(:args,
    s(:arg, :a, nil)), nil)
```

`Processor` adds an extra `nil` to the end of any required argument. Should it be fixed? Wondering because it requires some breaking changes in `processor`

iliabylich:
Closes #298

### DIFF
--- a/lib/parser/ast/processor.rb
+++ b/lib/parser/ast/processor.rb
@@ -39,9 +39,13 @@ module Parser
       def on_vasgn(node)
         name, value_node = *node
 
-        node.updated(nil, [
-          name, process(value_node)
-        ])
+        if !value_node.nil?
+          node.updated(nil, [
+            name, process(value_node)
+          ])
+        else
+          node
+        end
       end
 
       # @private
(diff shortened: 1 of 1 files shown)


whitequark:
No, wait, this is not right. `on_argument` is supposed to be overridden, so you need to call it from both `on_optional_argument` and `on_required_argument`.

iliabylich:
But `on_required_argument` should be different from `on_optional_argument`. It should simply return `node`. Or am I missing something?

whitequark:
Maybe just alter `on_argument` to look at the number of children then...

iliabylich:
Optional arg nodes always have two children, right? Then I guess we can just check for `value_node` to be present. Updated PR.

lib/parser/ast/processor.rb
```diff
@@ -39,9 +39,13 @@ def process_variable_node(node)
       def on_vasgn(node)
         name, value_node = *node
 
-        node.updated(nil, [
-          name, process(value_node)
-        ])
+        if value_node
```
whitequark: `if !value_node.nil?`

IlyaBylich: It can be false? Could you provide an example of such case? I can't get it locally

whitequark: It can't. The code I suggest conveys intent more clearly.

iliabylich:
@whitequark Fixed

PR #300, opened 2016-06-08
- Add missing Ruby 2.3 and 2.4 to .yardopts.

robblanco:


whitequark:
This is not a user-visible fix so it doesn't need the changelog marks in the commit. (I've fixed that.)

PR #311, opened 2016-08-30
Remove unused variable: begin_pos

nateberkopec:
Discovered w/Ruby verbose mode


alexdowad:
Thanks for cleaning up my code! At some point, I moved the sorting into `merge_actions`, but forgot to remove this code from `merge_replacements`.

PR #313, opened 2016-09-16
Fix mlhs for block arguments with trailing commas

### LINKED ISSUE #312: Trailing commas for block arguments should be parsed as mlhs
Ruby lets you use a trailing comma in a block argument for destructuring

```ruby
people = { 'John' => 22, 'Jim' => 23 }

people.each do |name,|
  puts "Hello, #{name}!"
end

# >> Hello, John!
# >> Hello, Jim!
```

Removing the comma results in different behavior:

```ruby
people = { 'John' => 22, 'Jim' => 23 }

people.each do |name|
  puts "Hello, #{name}!"
end

# >> Hello, ["John", 22]!
# >> Hello, ["Jim", 23]!
```

Parser produces the same AST for both cases:

```sh
$ ruby-parse --version
ruby-parse based on parser version 2.3.1.2

$ ruby-parse -e 'foo.each { |bar,| bar }'
(block
  (send
    (send nil :foo) :each)
  (args
    (arg :bar))
  (lvar :bar))

$ ruby-parse -e 'foo.each { |bar| bar }'
(block
  (send
    (send nil :foo) :each)
  (args
    (arg :bar))
  (lvar :bar))
```

## Expected behavior

Parser correctly handles destructuring like this for assignments:

```sh
$ ruby-parse -e 'foo = bar'
(lvasgn :foo
  (send nil :bar))
$ ruby-parse -e 'foo, = bar'
(masgn
  (mlhs
    (lvasgn :foo))
  (send nil :bar))
```

and parser also supports other forms of destructuring for block arguments:

```sh
$ ruby-parse -e 'foo.each { |(bar)| bar }'
(block
  (send
    (send nil :foo) :each)
  (args
    (mlhs
      (arg :bar)))
  (lvar :bar))
```

so it seems like the appropriate fix would be a parse that wraps the block arguments in `(mlhs ...)` like so:

```sh
$ ruby-parse -e 'foo.each { |bar,| bar }'
(block
  (send
    (send nil :foo) :each)
  (args
    (mlhs
      (arg :bar)))
  (lvar :bar))
```

Thanks for your great work on parser! I would be happy to look into this bug later today when I have some extra time.

backus:
Fixes #312 and backports the parser change to all rubies including 1.8, ruby motion, and macruby. I installed 1.8 and verified that this syntax should be backported that far. I was also able to verify this syntax for macruby but I'm not sure how to install rubymotion.

### DIFF
--- a/lib/parser/macruby.y
+++ b/lib/parser/macruby.y
@@ -1350,6 +1350,9 @@ rule
                                   concat(val[3])
                     }
                 | f_arg tCOMMA
+                  {
+                    result = [@builder.multi_lhs(nil, val[0], nil)]
+                  }
                 | f_arg tCOMMA                       f_rest_arg tCOMMA f_arg opt_f_block_arg
                     {
                       result = val[0].
--- a/lib/parser/ruby18.y
+++ b/lib/parser/ruby18.y
@@ -1259,6 +1259,9 @@ rule
 
        block_var: block_par
                 | block_par tCOMMA
+                  {
+                    result = [@builder.multi_lhs(nil, val[0], nil)]
+                  }
                 | block_par tCOMMA tAMPER lhs
                     {
                       result =  val[0].
--- a/lib/parser/ruby19.y
+++ b/lib/parser/ruby19.y
@@ -1330,6 +1330,9 @@ rule
                                   concat(val[3])
                     }
                 | f_arg tCOMMA
+                  {
+                    result = [@builder.multi_lhs(nil, val[0], nil)]
+                  }
                 | f_arg tCOMMA                       f_rest_arg tCOMMA f_arg opt_f_block_arg
                     {
                       result = val[0].
--- a/lib/parser/ruby20.y
+++ b/lib/parser/ruby20.y
@@ -1385,6 +1385,9 @@ opt_block_args_tail:
                                   concat(val[3])
                     }
                 | f_arg tCOMMA
+                  {
+                    result = [@builder.multi_lhs(nil, val[0], nil)]
+                  }
                 | f_arg tCOMMA                       f_rest_arg tCOMMA f_arg opt_block_args_tail
                     {
                       result = val[0].
--- a/lib/parser/ruby21.y
+++ b/lib/parser/ruby21.y
@@ -1365,6 +1365,9 @@ opt_block_args_tail:
                                   concat(val[3])
                     }
                 | f_arg tCOMMA
+                  {
+                    result = [@builder.multi_lhs(nil, val[0], nil)]
+                  }
                 | f_arg tCOMMA                       f_rest_arg tCOMMA f_arg opt_block_args_tail
                     {
                       result = val[0].
--- a/lib/parser/ruby22.y
+++ b/lib/parser/ruby22.y
@@ -1364,6 +1364,9 @@ opt_block_args_tail:
                                   concat(val[3])
                     }
                 | f_arg tCOMMA
+                  {
+                    result = [@builder.multi_lhs(nil, val[0], nil)]
+                  }
                 | f_arg tCOMMA                       f_rest_arg tCOMMA f_arg opt_block_args_tail
                     {
                       result = val[0].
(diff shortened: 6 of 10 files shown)


backus:
I'm not sure what is going on in CI but it looks like an issue that popped up in other PRs before this too. I think the issue is that rubygems needs to be updated in the .travis.yml. I'll experiment on this branch.

backus:
Looks like this fixed the problem. I'll extract a separate PR to avoid cluttering up this fix

whitequark:
This PR has merge conflicts, but more importantly it's not the correct fix. Parser never records locations of `tCOMMA`s as a matter of design. On top of that, even if it did, that wouldn't go into `begin` or `end`, since those exist for paired delimiters like braces.

backus:
@whitequark okie doke I've rebased and removed the comma tracking.

whitequark:
This was a rather impressive fix, because this bug was, I think, even grandfathered from some ruby_parser ancestry--it goes back to ~2007.

Thank you a lot for the fix, as well as for doing the legwork on 1.8 and MacRuby! RubyMotion is more or less identical to MacRuby so it is of no concern.

backus:
Happy to help! Feel free to ping me if this does lead to any issues that need to be addressed.

mbj:
OT: This bug was AFAIK discovered via the mutant self check called "neutral mutation", mutant tries to insert the unmodified code, and when specs not pass this means an unintend mutation was inserted. When parser and thus unparser inserted the wrongly parsed AST this resulted in a semantic change that caused those tests to fail.

I mainly post this here for future reference when I have to convince other mutation testing tool authors these self checks help to fix infrastructure problems.

iliabylich:
@whitequark @backus  As far as I understand this PR replaces args + trailing comma to mlhs. But it's valid only for a single argument + trailing comma:
``` ruby
# One argument, then comma:
proc { |a,| [a] }.call(1, 2, 3)
# => [1]
proc { |(a)| [a] }.call(1, 2, 3)
# => [1]

# Two arguments, then comma:
proc { |a,b,| [a,b] }.call(1, 2, 3)
# => [1, 2]
proc { |(a,b,_)| [a,b] }.call(1, 2, 3)
# => [1, nil]
```

mlhs is a single argument in the arguments list, MRI doesn't try to wrap all arguments with array and pass it as mlhs value.

whitequark:
@iliabylich I'm confused. None of the examples you use correspond to #312; the following would:
```
[5] pry(main)> proc { |a,| [a] }.call([1, 2])                                                                                                     
=> [1]                                                                                                                                            
[6] pry(main)> proc { |a| [a] }.call([1, 2])                                                                                                     
=> [[1, 2]]                                                                                                                                       
[7] pry(main)> a, = [1, 2]; a
=> 1                                                                                                                                              
[8] pry(main)> a = [1, 2]; a                                                                                                                     
=> [1, 2]                                                                                                                                         
```

iliabylich:
@whitequark In this PR args with trailing comma were replaced by mlhs argument i.e.:
``` sh
$ ruby-parse -v
ruby-parse based on parser version 2.3.1.4

$ ruby-parse -e "p { |a,b,| }"
(block
  (send nil :p)
  (args
    (mlhs
      (arg :a)
      (arg :b))) nil)

$ ruby-parse -e "p { |(a,b)| }"
(block
  (send nil :p)
  (args
    (mlhs
      (arg :a)
      (arg :b))) nil)
```
Generated AST is equal, but the code and its behavior is not, my previous example shows it. Could you please compare *mlhs argument* and *args + trailing comma*?

whitequark:
*sigh*

iliabylich:
I guess changing [this line](https://github.com/whitequark/parser/blob/master/lib/parser/ruby24.y#L1366) to `f_arg_item tCOMMA` fixes this issue - is it correct?

whitequark:
Definitely not. That doesn't match the upstream parser.

whitequark:
@iliabylich I think the right fix is to simply branch on `val[0].length == 1` in every parser.

PR #314, opened 2016-09-16
Update rubygems version in .travis.yml

backus:
I've extracted this from #313 which was experiencing failures for all MRI ruby versions due to the old rubygems version. This change should fix the problem and also make it possible to move 1.9.3 out of `allowed_failures`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ matrix:
  allow_failures:
   - rvm: ruby-head
   - rvm: rbx-2
-  - rvm: 1.9.3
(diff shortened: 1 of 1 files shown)


whitequark:
> and also make it possible to move 1.9.3 out of allowed_failures.

Can you do that?

backus:
@whitequark Yeah I was waiting to push it actually so that travis wouldn't clone the second commit for any of the pending builds

backus:
@whitequark this is ready for review now

PR #321, opened 2016-09-30
Add `#expand` and `#contract` methods to `Parser::Source::Range`

Drenmi:
I fully understand the implications of extending the public API with new methods, but I thought I'd offer this for your consideration anyway. The need for these methods have arisen now and then as I've been working on [RuboCop](https://github.com/bbatsov/rubocop), where a new range sometimes needs to be constructed by expanding or contracting an existing one (often in order to include or exclude brackets.)

Because the need to fully qualify the constant, these methods offer ample opportunity for code reduction, and can clean out some of the direct references to `Parser` in higher level constructs (i.e. cops.) Consider:

```
# Before
Parser::Source::Range(node.source_range.source_buffer, node.begin_pos + 1, node.end_pos - 1)

# After
node.source_range.contract
```

This pull request also contains a small amount of opportunistic clean up; adding a unit test for `#resize`, and fixing a minor indentation mistake in `#join`. Feel free to pick and choose among the commits, should you want some of them, but not all.

🙇

### DIFF
--- a/lib/parser/source/range.rb
+++ b/lib/parser/source/range.rb
@@ -179,8 +179,8 @@ module Parser
       #
       def join(other)
         Range.new(@source_buffer,
-            [@begin_pos, other.begin_pos].min,
-            [@end_pos,   other.end_pos].max)
+          [@begin_pos, other.begin_pos].min,
+          [@end_pos,   other.end_pos].max)
       end
 
       ##
--- a/lib/parser/source/range.rb
+++ b/lib/parser/source/range.rb
@@ -174,6 +174,20 @@ module Parser
       end
 
       ##
+      # @param [integer] by
+      # @return [Range] a range expanded in both directions by `by`.
+      def expand(by = 1)
+        Range.new(@source_buffer, @begin_pos - by, @end_pos + by)
+      end
+
+      ##
+      # @param [integer] by
+      # @return [Range] a range contracted in both directions by `by`.
+      def contract(by = 1)
+        Range.new(@source_buffer, @begin_pos + by, @end_pos - by)
+      end
+
+      ##
       # @param [Range] other
       # @return [Range] smallest possible range spanning both this range and `other`.
       #
--- a/test/test_source_range.rb
+++ b/test/test_source_range.rb
@@ -135,4 +135,31 @@ class TestSourceRange < Minitest::Test
       sr.resize(-1)
     end
   end
+
+  def test_expand
+    sr1 = Parser::Source::Range.new(@buf, 2, 3)
+    sr2 = Parser::Source::Range.new(@buf, 1, 4)
+    assert_equal sr2, sr1.expand
+
+    sr1 = Parser::Source::Range.new(@buf, 2, 3)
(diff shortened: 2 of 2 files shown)


whitequark:
> * Add unit tests for `Parser::Source::Range#resize`

> * Fix bad indentation in `Parser::Source::Range#join`

These are not changes to API so they should not be marked with `*`.

whitequark:
The code is alright and I have no objections to this extension of API, in principle. However, it is not clear for me what is the use case. Why do you want to expand or contract ranges bidirectionally? Why is doing this by one character a common use case? I'm struggling to imagine.

bbatsov:
Well, @Drenmi already provided some examples from our codebase - often you need to build some new source range based on an existing range and the way we do it right now is quite ugly.

whitequark:
@bbatsov That's not really an example because it lacks motivation. Sure, fragment B is less verbose than fragment A but why would I want to write fragment A in the first place?

Drenmi:
@whitequark, @bbatsov: I tried to backtrack to find the cases where I needed this in the past, but this was quite a few months ago, so I think instead what I will do is go through our current uses of `Parser::Source::Range#new`, and see whether this API extension would be the most effective at reducing that, or whether it should be something else entirely. 😀 

If you want, we can close this while I do that. If this turns out to be the best option, I can reopen the pull request with more, concrete examples.

whitequark:
I don't mind keeping this open.

alexdowad:
In general, `Source::Range` objects are passed around and used quite a bit in Rubocop to represent some bit of source text which is being analyzed. We have helper methods which do things like "expanding a range to encompass whitespace on the left/right/both", "expanding a range to encompass parens on the left/right/both", "expanding a range to encompass commas on the left/right/both", and so on.

When rewriting source text to improve its style, we also use `Source::Range` objects to represent the code to replace, which is a natural fit with Parser's Rewriter.

Say for example we want to remove the last argument from a method call. Leaving a hanging comma would result in a syntax error, so we need to expand the replaced range to include the comma on the left (if there is one).

Manipulating ranges to make them larger/smaller/whatever is *usually* done when rewriting code to improve its style. There are a few other cases, but they are in the minority.

whitequark:
> We have helper methods which do things like "expanding a range to encompass whitespace on the left/right/both", "expanding a range to encompass parens on the left/right/both", "expanding a range to encompass commas on the left/right/both", and so on.

> Leaving a hanging comma would result in a syntax error, so we need to expand the replaced range to include the comma on the left (if there is one).

Sorry, I've been unclear. I absolutely understand that, which is precisely why I don't understand this PR. All of the examples you provide depend in some way on the semantics of the text, and most of them are asymmetric. The methods added in this PR disregard semantics (let's say what you want to do is to encompass parens on left and right; you can't just call `#expand`, you have to check if there really are parens immediately near) and are symmetric only.

whitequark:
Ping? Could you reply to my comments above?

Drenmi:
@whitequark: I'm no longer sure this is as useful as it first seemed. Will close for now and revisit when I have more time. Sorry. 😊

PR #322, opened 2016-10-07
Introduce (procarg0) node for tCOMMA in "|a,|".

### LINKED ISSUE #312: Trailing commas for block arguments should be parsed as mlhs

iliabylich:
Closes #312.

### DIFF
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -912,6 +912,23 @@ Format:
 
 Begin of the `expression` points to `&`.
 
+### Required single block argument (1.9)
+
+Parser emits this node when a block has only one required argument without trailing comma.
+
+Note: `proc {|a|}` and `proc {|a,|}` are semantically
+different, the second one acts more like mlhs (`p { |(a)| }`) and it
+expands passed array argument, i.e. `proc { |a,| a }.call([1])` returns `1`.
+
+Format:
+
+~~~
+(procarg0 :foo)
+"|foo|"
+  ~~~ expression
+  ~~~ name
+~~~
+
 ### Expression arguments
 
 Ruby 1.8 allows to use arbitrary expressions as block arguments,
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -24,9 +24,28 @@ module Parser
 
     class << self
       ##
+      # AST compatibility attribute; block arguments of `m { |a| }` are
+      # not semantically equivalent to block arguments of `m { |a,| }` or `m { |a, b| }`,
+      # all new code should set this attribute to true.
+      #
+      # If set to false (the default), arguments of `m { |a| }` are emitted as
+      # `s(:args, s(:arg, :a))`
+      #
+      # If set to true, arguments of `m { |a| }` are emitted as
+      # `s(:args, s(:procarg0, :a))
+      #
+      # @return [Boolean]
+      attr_accessor :emit_procarg0
+    end
+
+    @emit_procarg0 = false
+
+    class << self
+      ##
       # @api private
       def modernize
         @emit_lambda = true
+        @emit_procarg0 = true
       end
     end
 
@@ -637,6 +656,14 @@ module Parser
         arg_prefix_map(amper_t, name_t))
     end
 
+    def procarg0(arg)
+      if self.class.emit_procarg0
+        arg.updated(:procarg0)
+      else
+        arg
+      end
+    end
+
     # Ruby 1.8 block arguments
 
     def arg_expr(expr)
@@ -1112,7 +1139,7 @@ module Parser
         case this_arg.type
         when :arg, :optarg, :restarg, :blockarg,
              :kwarg, :kwoptarg, :kwrestarg,
-             :shadowarg
+             :shadowarg, :procarg0
 
           this_name, = *this_arg
 
--- a/test/test_parser.rb
+++ b/test/test_parser.rb
@@ -2350,8 +2350,14 @@ class TestParser < Minitest::Test
     # f_arg                                                      opt_f_block_arg
     # f_arg tCOMMA
     assert_parses_blockargs(
+      s(:args, s(:procarg0, :a)),
+      %q{|a|},
+      SINCE_1_9)
+
+    assert_parses_blockargs(
       s(:args, s(:arg, :a)),
-      %q{|a|})
+      %q{|a|},
+      %w(1.8))
 
     assert_parses_blockargs(
       s(:args, s(:arg, :a), s(:arg, :c)),
@@ -2363,7 +2369,7 @@ class TestParser < Minitest::Test
       %w(1.8))
 
     assert_parses_blockargs(
-      s(:args, s(:mlhs, s(:arg, :a))),
+      s(:args, s(:arg, :a)),
       %q{|a,|}
     )
 
@@ -2565,6 +2571,25 @@ class TestParser < Minitest::Test
         s(:blockarg, :b)),
       %q{|*r, p, &b|},
       SINCE_1_9)
+  end
+
+  def test_multiple_args_with_trailing_comma
+    assert_parses_blockargs(
+      s(:args,
+        s(:arg, :a),
+        s(:arg, :b)),
+      %q(|a, b,|)
+    )
+  end
+
+  def test_procarg0_legacy
+    Parser::Builders::Default.emit_procarg0 = false
+    assert_parses_blockargs(
+      s(:args,
+        s(:arg, :a)),
+      %q{|a|}
+    )
+    Parser::Builders::Default.emit_procarg0 = true
   end
 
   def test_block_kwarg_combinations
@@ -5385,13 +5410,28 @@ class TestParser < Minitest::Test
           s(:send,
             s(:int, 1), :tap),
           s(:args,
+            s(:procarg0, :n)),
+          s(:send, nil, :p,
+            s(:lvar, :n))),
+        s(:int, 0)),
+      %q{true ? 1.tap do |n| p n end : 0},
+      %q{},
+      SINCE_1_9)
+
+    assert_parses(
+      s(:if,
+        s(:true),
+        s(:block,
+          s(:send,
+            s(:int, 1), :tap),
+          s(:args,
             s(:arg, :n)),
           s(:send, nil, :p,
             s(:lvar, :n))),
         s(:int, 0)),
       %q{true ? 1.tap do |n| p n end : 0},
       %q{},
-      ALL_VERSIONS)
+      %w(1.8))
 
     assert_parses(
       s(:if,
@@ -5712,8 +5752,20 @@ class TestParser < Minitest::Test
         s(:send, nil, :a,
           s(:ivar, :@b)),
         s(:args,
+          s(:procarg0, :c)), nil),
+      %q{a @b do |c|;end},
+      %q{},
+      SINCE_1_9)
+
+    assert_parses(
+      s(:block,
+        s(:send, nil, :a,
+          s(:ivar, :@b)),
+        s(:args,
           s(:arg, :c)), nil),
-      %q{a @b do |c|;end})
+      %q{a @b do |c|;end},
+      %q{},
+      %w(1.8))
   end
 
   def test_bug_lambda_leakage
--- a/lib/parser/ast/processor.rb
+++ b/lib/parser/ast/processor.rb
@@ -121,6 +121,7 @@ module Parser
       alias on_kwarg          process_argument_node
       alias on_kwoptarg       process_argument_node
       alias on_kwrestarg      process_argument_node
+      alias on_procarg0       process_argument_node
 
       alias on_arg_expr       process_regular_node
       alias on_restarg_expr   process_regular_node
--- a/lib/parser/macruby.y
+++ b/lib/parser/macruby.y
@@ -1347,9 +1347,6 @@ rule
                                   concat(val[3])
                     }
                 | f_arg tCOMMA
-                  {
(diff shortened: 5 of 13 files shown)


whitequark:
I prefer the name `procarg0`. Also, the AST should look like `s(:args, s(:procarg0, :a))` for the `m { |a,| }` case since this is logically equivalent to a sort of mlhs.

Also, this is a somewhat exceptional case in that the comma is not just a separator but it also has semantic meaning. So, we need to put the source location of the comma into the source map. I think `Source::Map::Operator` would do here.

whitequark:
Oh, and you should update `doc/AST_FORMAT.md`.

iliabylich:
@whitequark Should this `procarg0` node be an operator? And if so, should I document it in the `Arguments` section of `doc/AST_FORMAT.md`?

whitequark:
> Should this procarg0 node be an operator?

It is an argument node, like `s(:kwarg)`. It should be documented in that section. The reason it uses `Source::Map::Operator` is distinct; the correspondence from AST nodes to map classes is ad-hoc and doesn't carry much semantic meaning (neither it is even stable, only the valid accessors are).

iliabylich:
@whitequark Done I guess.

doc/AST_FORMAT.md
```diff
@@ -997,6 +997,17 @@ Format:
...
 ~~~
 
+### "Excessed comma"
+
+Representation of the trailing comma after block arguments.
+
+~~~
+(procarg0)
+"|a,|"
+ ~~~~ expression
```
whitequark: Should be:
```
|a,|
 ~~ expression
  ~ operator
```

lib/parser/builders/default.rb
```diff
@@ -24,9 +24,28 @@ class << self
...
+      # `s(:args, s(:arg, :a))`
+      #
+      # If set to true, arguments of `m { |a,| }` are emitted as
+      # `s(:args, s(:arg, :a), s(:procarg0))
+      #
+      # @return [Boolean]
+      attr_accessor :emit_excessed_comma
+    end
+
+    @emit_excessed_comma = false
```
whitequark: `@emit_procarg0` and elsewhere.

test/test_parser.rb
```diff
@@ -3388,6 +3383,26 @@ def test_send_lambda_legacy
     Parser::Builders::Default.emit_lambda = true
   end
 
+  def test_excessed_comma
+    assert_parses(
+      s(:block, s(:send, nil, :m),
+        s(:args, s(:arg, :a), s(:procarg0)), nil),
+      %q{m { |a, | }},
+      %q{    ~~~~~ expression (args)
```
whitequark: needs a test for `expression (args.procarg0)`.

test/test_parser.rb
```diff
@@ -3388,6 +3383,26 @@ def test_send_lambda_legacy
...
+        |      ~ operator (args.procarg0)
+      }
+    )
+  end
+
+  def test_excessed_comma_legacy
+    Parser::Builders::Default.emit_excessed_comma = false
+    assert_parses_blockargs(
+      s(:args, s(:arg, :a)),
+      %q{|a,|}
```
whitequark: also needs a location test

doc/AST_FORMAT.md
iliabylich: But I'm getting this test failing locally:
``` ruby
assert_parses(
  s(:block, s(:send, nil, :m),
    s(:args, s(:arg, :a)), nil),
  %q{m { |a| }},
  %q{     ~ expression (args)
  }
)
```
While this one passes:
``` ruby
assert_parses(
  s(:block, s(:send, nil, :m),
    s(:args, s(:arg, :a)), nil),
  %q{m { |a| }},
  %q{    ~~~ expression (args)
  }
)
```

test/test_parser.rb
iliabylich: What should be tested here? There's no `procarg0` node.

doc/AST_FORMAT.md
whitequark: No no, the expression for `(args)` encompasses `|a|` whereas the expression for `(args.arg)` encompasses `a`.

iliabylich:
Finally got it, fixed :smile:

whitequark:
As described in https://github.com/whitequark/parser/issues/312#issuecomment-252407424, the fix in this PR is essentially inverted from the real semantics. Can you please make it so that:

  * `proc{|a|}` results in `(args (procarg0 :a))`;
  * everything else results in exactly the same code as before, i.e. `proc{|a,|}` becomes `(args (arg :a))`.

There should probably be some sort of flag passed to the builder. I suggest first altering `ruby24.y`, and once that passes review you can duplicate it elsewhere.

iliabylich:
@whitequark Is it possible to handle it on the grammar level? I mean, `lambda` and `proc` are just Kernel methods, they can be aliased to each other or completely removed. For me it sounds like a method-level grammar, the only block that has a new argument type is a block passed to a `proc` method and has **exactly** one argument without trailing comma.

whitequark:
> I mean, lambda and proc are just Kernel methods, they can be aliased to each other or completely removed.

This is wrong. Lambda/proc distinction, although visible through the Kernel methods, is fundamental to the language. There are not just two methods, but two distinct types of closures in Ruby;  `-> {}` is an alias for `lambda` (you used to be able to override this by replacing `Kernel#lambda`, and this is why we didn't originally have a `(lambda)` node, just a `(block (send (const :Kernel) :lambda) etc)`, but not anymore), whereas `do f(&b) b end; f {}` is an alias for `proc`.

```
> ->{}.lambda?
=> true
> def f(&b) b end; f {}.lambda?
=> false
```

The thing is that the parser has exactly one syntactic construct, a block, but depending on which runtime context this block appears in, it can become a proc and a lambda. Moreover these two are not interconvertible into each other:
```
[9] pry(main)> proc(&->{}).lambda?
=> true
[11] pry(main)> lambda(&f{}).lambda?
=> false
```

Anyway, the bottom line is that when we parse the block--when we are *about to pass* a block to a function--we have no idea if the arguments should be proc-like or lambda-like, and so this information has to be preserved in the AST.

whitequark:
> This is wrong.

Scratch this, I may have just misunderstood you.

whitequark:
Let me try to rephrase the above the other way, and maybe more concisely: it is not possible to handle the distinction *purely* on the AST/grammar level, however a marker on AST/grammar level is *necessary* for the runtime support code to decide on how the arguments should be parsed, in case the block is passed to `proc`.

iliabylich:
@whitequark Consider the following example:
``` ruby
module Kernel
  alias proc2 proc
  alias lambda2 lambda
end

[
  lambda { |a,| [a] },
  lambda2 { |a,| [a] },

  proc { |a,| [a] },
  proc2 { |a,| [a] }
].each do |pr|
  p pr.call([1])
end
```

I'm trying to say that we can't simply stick to a method called `proc`. `proc` is just a method name that has some behavior and this method can potentially have absolutely any name.

whitequark:
> I'm trying to say that we can't simply stick to a method called proc. proc is just a method name that has some behavior and this method can potentially have absolutely any name.

Oh! I agree of course, I wrote my earlier comment misleadingly. In my earlier comment...

> proc{|a|} results in (args (procarg0 :a));
> everything else results in exactly the same code as before, i.e. proc{|a,|} becomes (args (arg :a)).

by `proc` I mean "any expression", not the `proc` function specifically.

iliabylich:
Alright, updated only 2.4, other versions are temporary excluded from the build/test suite.

@whitequark Could you review it please? The code is ugly but at least is seems to be correct now.

whitequark:
I think you can get away with making just one change, in the `block_param: f_arg opt_block_args_tail` rule. That is, when val[1] is empty and val[0] has one element, branch to `@builder.procarg0` else do like before.

iliabylich:
@whitequark Great idea, done. Any other comments before applying same rules on other versions?
By the way, I guess `AST::Processor` also requires one more alias for a new argument type - is it true?

whitequark:
> Any other comments before applying same rules on other versions?

Looks good to me.

> By the way, I guess AST::Processor also requires one more alias for a new argument type - is it true?

Yep.

iliabylich:
What should happen with `(arg_expr)` from 1.8? One more node?
As far as I understand it should be:

``` ruby
proc { |@a| }

# Before
(block
  (send nil :proc)
  (args
    (arg-expr
      (ivasgn :@a))) nil)

# After
(block
  (send nil :proc)
  (args
    (procarg0-expr
      (ivasgn :@a))) nil)
```

whitequark:
Nothing.
```
[10] pry(main)> proc{|a,|[a]}.call(1)
=> [1]
[11] pry(main)> proc{|a,|[a]}.call([1])
=> [[1]]
[12] pry(main)> proc{|a|[a]}.call(1)                                                                                                           
=> [1]                                                                                                                                            
[13] pry(main)> proc{|a|[a]}.call([1])                                                                                                            
=> [[1]]                                                                                                                                          
[14] pry(main)> RUBY_VERSION                                                                                                                      
=> "1.8.7"                                                                                                                                        
```

iliabylich:
@whitequark Was about to ask you for a final review, but then realized that this fix doesn't handle single optional argument:
``` ruby
2.3.0 :001 > proc { |a=1| a }.call(1)
 => 1 
2.3.0 :002 > proc { |a=1| a }.call([1])
 => [1] 
2.3.0 :003 > proc { |a| a }.call(1)
 => 1 
2.3.0 :004 > proc { |a| a }.call([1])
 => [1]
```

I believe it can't be solved without introducing 1 more node. `(optprocarg0`) sounds horrible.

How about closing this issue? I'm sure that it can be handled by the library that uses `parser` using source maps. [This](https://github.com/opal/opal/blob/master/lib/opal/nodes/iter.rb#L167-L172) is how opal handles it atm.

I just think that it's a bad idea to break compatibility and introduce some new nodes to handle such a small issue.

whitequark:
> How about closing this issue? I'm sure that it can be handled by the library that uses parser using source maps. This is how opal handles it atm.

No. This is utter nonsense. Parser's AST is designed such that source maps are *never* required to interpret them semantically. `optprocarg0` looks just fine to me.

> I just think that it's a bad idea to break compatibility and introduce some new nodes to handle such a small issue.

We don't break compatibility because this is opt-in.

iliabylich:
@whitequark Added `optprocarg0` node, could you check it please?

doc/AST_FORMAT.md
```diff
@@ -864,6 +864,19 @@ Format:
  ~~~ name
 ~~~
 
+### Required single block argument
```
whitequark: Just put the version requirement in the title, like `### Required single block argument (1.9)`. This is the convention used elsewhere.

doc/AST_FORMAT.md
```diff
@@ -864,6 +864,19 @@ Format:
...
+
+Only for parser > 1.8.
+
+Format:
+
+~~~
+(procarg0 :foo)
+|foo|
```
whitequark: This should be in quotes, like `"|foo|"`.

doc/AST_FORMAT.md
```diff
@@ -876,6 +889,17 @@ Format:
  ~~~ name
 ~~~
 
+### Optional single block argument
```
whitequark: I think both of these should go just before "Expression arguments" and gain a small note about their difference from `|a,|`.

iliabylich:
Hold on, MRI doesn't let you to pass trailing comma after optional block argument. Do we really need a `optprocarg0` node? There are no places where you need to identify a trailing comma after optional argument and make it acting like mlhs *because it's just impossible*.

whitequark:
Then nope!

iliabylich:
Updated PR

iliabylich:
Thank you for your patience :smile:

PR #295, opened 2016-06-02
* parser/current: update 2.2 warning to 2.2.5

eagletmt:
https://www.ruby-lang.org/en/news/2016/04/26/ruby-2-2-5-released/

PR #329, opened 2016-12-01
* parser/current: update 2.2 warning to 2.2.6

jormon:
https://www.ruby-lang.org/en/news/2016/11/15/ruby-2-2-6-released/


jormon:
just copied the intent of #295

PR #331, opened 2016-12-07
Travis: remove 1.8.7

olleolleolle:
Travis' builds fail on there not being a binary to install.

This PR removes 1.8.7, rbx-2 from the build matrix.


whitequark:
Uh, and why are you removing rbx and jruby-18mode?

whitequark:
I've removed just MRI 1.8.7 in master.

PR #348, opened 2017-05-22
Fix ruby_parser comparison markup

ddfreyne:
screen shot 2017-05-22 at 20 33 59

Without the blank line, the markup is broken.


ddfreyne:
This one’s incomplete, as there’s a bunch of other places where the problem occurs, I just noticed!

PR #354, opened 2017-06-22
Get tests passing with frozen-string-literals enabled.

pat:
These changes ensure that all string literals can be frozen (as per the optional feature in MRI 2.3 and onwards). I would recommend adding the following to your `.travis.yml` file to ensure regressions aren't introduced:

```yml
before_script:
- if (ruby -e "exit RUBY_VERSION.to_f >= 2.4"); then export RUBYOPT="--enable-frozen-string-literal"; fi; echo $RUBYOPT
```

This will add the flag when the tests are run on MRI 2.4 or newer (while the feature was introduced in 2.3, it doesn't seem to work reliably until 2.4).

### DIFF
--- a/lib/parser/lexer/literal.rb
+++ b/lib/parser/lexer/literal.rb
@@ -237,7 +237,7 @@ module Parser
     end
 
     def clear_buffer
-      @buffer = ''
+      @buffer = ''.dup
 
       # Prime the buffer with lexer encoding; otherwise,
       # concatenation will produce varying results.
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ rvm:
  - 2.0.0
  - 2.2.5
  - 2.3.1
+ - 2.4.1
  - ruby-head
  - jruby-18mode
  - jruby-19mode
@@ -18,5 +19,7 @@ before_install:
  - bundle --version
  - gem update --system 2.6.6
  - gem --version
+before_script:
+- if (ruby -e "exit RUBY_VERSION.to_f >= 2.4"); then export RUBYOPT="--enable-frozen-string-literal"; fi; echo $RUBYOPT
 script:
  - bundle exec rake test_cov
--- a/lib/parser/source/buffer.rb
+++ b/lib/parser/source/buffer.rb
@@ -246,7 +246,7 @@ module Parser
       def source_lines
         @lines ||= begin
           lines = @source.lines.to_a
-          lines << '' if @source.end_with?("\n".freeze)
+          lines << ''.dup if @source.end_with?("\n".freeze)
 
           lines.each do |line|
             line.chomp!("\n".freeze)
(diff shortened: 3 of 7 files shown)


lib/parser/lexer/literal.rb
```diff
@@ -237,7 +237,7 @@ def coerce_encoding(string)
     end
 
     def clear_buffer
-      @buffer = ''
+      @buffer = ''.dup
```
segiddins: Why not use `String.new` for these to avoid the extra allocation?

pat: I've recently discovered via @koic that Ruby isn't quite consistent with `String.new`. `String.new.encoding` is `ASCII-8BIT`, whereas `String.new('').encoding` (which doesn't avoid the extra allocation) is `UTF-8`.

I'm happy to change the PR to use `String.new('')` - I presume UTF-8 is the preferred encoding? - I've just found that some library maintainers (particularly Rails) prefer `dup`.

whitequark:
> I would recommend adding the following to your .travis.yml file to ensure regressions aren't introduced:

Please do so.

whitequark:
Fails:
```
/home/travis/.rvm/gems/ruby-2.4.1/gems/racc-1.4.14/lib/racc/statetransitiontable.rb:214:in `mkmapexp': can't modify frozen String (RuntimeError)
```

pat:
Yeah, I noticed that too :(

There are PRs (for both 1.4.x and master branches) for frozen string literals in racc pending, though they've been dormant for a little while. I'm going to try to help get their builds passing when I have a spare moment, but that's probably not in the next 24 hours or so.
https://github.com/tenderlove/racc/pull/80
https://github.com/tenderlove/racc/pull/81

pat:
Hi all, just coming back to this PR after several months, and with a couple of thoughts:

* the master branch of racc seems to be frozen-string-literal-friendly (but I've no idea if/when a new gem release will come out).
* These days I'm actually preferring the pragma comment `# frozen_string_literal: true` at the top of each file, rather than the `before_script` option I've added to the travis configuration. The catch is that it becomes something that needs to be remembered every time new files are introduced into the project - but removes the need for dependencies to be completely frozen-string-literal-friendly.

I'm happy to re-jig this PR to take the latter approach if that's desired? Would be great to get this merged in, rather than just waiting hopefully for a new racc release.

whitequark:
> but I've no idea if/when a new gem release will come out

Ping @tenderlove?

> I'm happy to re-jig this PR to take the latter approach if that's desired? Would be great to get this merged in, rather than just waiting hopefully for a new racc release.

Sure, we can even do both.

pat:
@whitequark I submitted a new PR with the pragma comments. This way at least things aren't reliant on dependencies (and I'm sure @tenderlove is either busy enough, or hopefully enjoying a bit of a break).

So, I figure this PR can be closed, but possibly revisited later if people are particularly keen to use the `RUBYOPT` env flag.

PR #355, opened 2017-06-26
Allows file paths to be used as well as file names for Ruby Rewrite.

HelenCampbell:
Previously Ruby Rewrite would only work when a file name was passed in. The addition of these changes now allows for file paths without affecting the old functionality.

### DIFF
--- a/lib/parser/runner/ruby_rewrite.rb
+++ b/lib/parser/runner/ruby_rewrite.rb
@@ -32,7 +32,10 @@ module Parser
     def load_and_discover(file)
       load file
 
-      const_name = file.
+      file_name = File.basename(file)
+      file_name[0] = file_name[0].capitalize
+
+      const_name = file_name.
         sub(/\.rb$/, '').
         gsub(/(^|_)([a-z])/) do |m|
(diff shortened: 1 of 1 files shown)


lib/parser/runner/ruby_rewrite.rb
```diff
@@ -32,7 +32,10 @@ def setup_option_parsing(opts)
     def load_and_discover(file)
       load file
 
-      const_name = file.
+      file_name = File.basename(file)
+      file_name[0] = file_name[0].capitalize
```
whitequark: I don't think this line is necessary, the gsub block below does the same.

pmcmaw: @whitequark when using it the gsub section below only seems to capitalize the letter after the underscore. 

Input: string_wrapper.rb
Actual Output: stringWrapper

When adding that line it now outputs the following:
Expected Output: StringWrapper

whitequark: This is not the case. Try it yourself in irb and you will see.

PR #361, opened 2017-08-01
Fix warnings

iliabylich:
1. Fixed warnings in `Parser::Source::Rewriter`.
2. Fixed warnings coming from minitest (`assert nil, value` -> `assert_nil value`, see https://github.com/seattlerb/minitest/issues/666)
3. Changed CI config, it can fix an issue with installing 1.9.2 on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: ruby
 rvm:
  - 1.9.2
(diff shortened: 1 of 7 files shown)


whitequark:
Any idea why the JRuby builds started failing? If it's unrelated I'll merge.

iliabylich:
@whitequark I've installed jruby locally (`jruby 9.1.11.0-SNAPSHOT`) and everything works fine (both `rake` and `rake regenerate`). I have no ideas what's wrong with CI.

It can be related to bumping OS version from `precise` to `trusty`, but anyway, `precise` is no longer supported and travis plans to start forcing `trusty` this September.

PR #293, opened 2016-05-27
Remove string mutations in Parser::Lexer::Literal.

iliabylich:
Required for https://github.com/opal/opal/pull/1465

PR #364, opened 2017-09-06
Fix performance regression for a string that contains a very long line

pocke:
`+=` is replaced with `<<` in `extend_string` method in #293.
This replacing did not introduce performance regressions about small string. However, it has introduced a performance regression about a string that contains very long line.

This seems to be a problem in an actual application. An issue was opened
in RuboCop repository about this problem. https://github.com/bbatsov/rubocop/issues/4701

`<<` is replaced with `+=` also in `initialize` method by #293. However this `<<` doesn't introduce regressions because it is a concatenation of small strings.

Benchmark
-----

```
File.open 'test.rb', 'w' do |f|
  f.write '"'
  f.write 'a' * 100000
  f.write '"'
end
```

Master:

```bash
$ bin/ruby-parse -B test.rb
Parsed 1 files (100002 characters) in 0.73 seconds (136.989 kchars/s).
Running on ruby 2.4.1.
```

Patch:

```bash
$ bin/ruby-parse -B test.rb
Parsed 1 files (100002 characters) in 0.16 seconds (625.012 kchars/s).
Running on ruby 2.4.1.
```

### DIFF
--- a/lib/parser/lexer/literal.rb
+++ b/lib/parser/lexer/literal.rb
@@ -187,7 +187,7 @@ module Parser
       @buffer_s ||= ts
       @buffer_e = te
 
-      @buffer += string
+      @buffer << string
     end
 
     def flush_string
--- a/lib/parser/lexer/literal.rb
+++ b/lib/parser/lexer/literal.rb
@@ -183,20 +183,11 @@ module Parser
       (@interp_braces == 0)
     end
 
-    if RUBY_ENGINE == 'opal'
-      def extend_string(string, ts, te)
-        @buffer_s ||= ts
-        @buffer_e = te
+    def extend_string(string, ts, te)
+      @buffer_s ||= ts
+      @buffer_e = te
 
-        @buffer += string
-      end
-    else
-      def extend_string(string, ts, te)
-        @buffer_s ||= ts
-        @buffer_e = te
-
-        @buffer << string
-      end
+      @buffer += string
     end
 
     def flush_string
--- a/lib/parser/lexer/literal.rb
+++ b/lib/parser/lexer/literal.rb
@@ -183,11 +183,20 @@ module Parser
       (@interp_braces == 0)
     end
 
-    def extend_string(string, ts, te)
-      @buffer_s ||= ts
-      @buffer_e = te
+    if RUBY_ENGINE == 'opal'
+      def extend_string(string, ts, te)
+        @buffer_s ||= ts
+        @buffer_e = te
 
-      @buffer += string
+        @buffer += string
+      end
+    else
+      def extend_string(string, ts, te)
+        @buffer_s ||= ts
+        @buffer_e = te
+
+        @buffer << string
+      end
     end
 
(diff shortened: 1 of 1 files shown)


lib/parser/lexer/literal.rb
```diff
@@ -183,11 +183,20 @@ def end_interp_brace_and_try_closing
       (@interp_braces == 0)
     end
 
-    def extend_string(string, ts, te)
-      @buffer_s ||= ts
-      @buffer_e = te
+    if RUBY_ENGINE == 'opal'
```
whitequark: I strongly oppose having special implementations for different Ruby engines.

whitequark:
@iliabylich I'm likely going to revert #293, unless you have a better solution. It is Opal's job to support mutable strings properly, like all other Ruby implementations, not mine to work around its issues...

iliabylich:
@whitequark I agree, opal doesn't support mutable strings, so I guess we have to make a patch directly in opal.

May I ask you (and @pocke ) why do you think that this performance regression is important? That's a really rare case when you have such strings in your code (we can always put them to a separate file). I'm sure we can find a lot of other edges cases like "parser slowly parses some long code construction" - are you going to fix all of them no matter are they real or not? Just asking because mutable strings somewhere in the lexer/parser (generated files) can't be patched on the opal's side without maintaining a fork.

whitequark:
> May I ask you (and @pocke ) why do you think that this performance regression is important? That's a really rare case when you have such strings in your code (we can always put them to a separate file).

I'm generally not happy with having quadratic and above behavior anywhere in Parser. This case has resulted in 10+ minute runtimes, which is not something I can justify declaring _not a bug_, especially given that Ruby itself parses it just fine.

A compromise solution is to add all string parts to an array and concatenate it once in the end. @iliabylich, @pocke, are you willing to implement this?

iliabylich:
@whitequark To be fair I'd go with a string mutation in this case. Opal has to avoid mutation to support `Kernel#eval` and run parser in the browser, but as I know nobody uses it (because it's slow and a bit broken 😄 ). In 99.99% of cases we run opal on the backend, so using string mutation actually makes opal faster.

So, I propose to completely revert https://github.com/whitequark/parser/pull/293 and make everyone happy.

And in this case I can simply port the following part of this PR to opal:
```
if RUBY_ENGINE == 'opal'
  class Parser::Lexer::Literal
    def extend_string(string, ts, te)
      @buffer_s ||= ts
      @buffer_e = te

      @buffer += string
    end
  end
end
```

whitequark:
@pocke, would you please do the revert?

lib/parser/lexer/literal.rb
bbatsov: Opal is not a Ruby engine. 😃 It's a Ruby to JS compiler with quite a few differences from "standard" Ruby.

pocke:
I reverted "remove string mutations", but I didn't revert `+=` in initialize method, because it is not slow.

PR #366, opened 2017-09-16
Parser::Current: update for Ruby 2.2.8

koic:
Ruby 2.2.8 has been released.
https://www.ruby-lang.org/en/news/2017/09/14/ruby-2-2-8-released/

As far as I saw, there seems to be no change in syntax to `parse.y` etc.
https://github.com/ruby/ruby/compare/v2_2_7...v2_2_8


whitequark:
Thanks!

PR #368, opened 2017-09-19
- Builders::Default: fix location for __ENCODING__

ABaldwinHunter:
A constant `Logger::INFO` gets parsed as a node with location of class
`Parser::Source::Map::Constant`. Since "\_\_ENCODING\_\_" gets parsed as a
constant node, it seems consistent to use the same Map::Constant class.

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -416,7 +416,7 @@ module Parser
 
       when :__ENCODING__
         n(:const, [ n(:const, [ nil, :Encoding], nil), :UTF_8 ],
-          node.loc.dup)
+          constant_map(nil, nil, ["Encoding", node.loc.expression]))
 
       when :ident
         name, = *node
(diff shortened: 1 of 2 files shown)


ABaldwinHunter:
Thanks for all of the work on this great tool! I'm currently working on a project that uses the Parser and is getting tripped up on processing the `__ENCODING__` constant because of the difference in location type:

```
Parser::CurrentRuby.parse("__ENCODING__").children.first.location
=> nil

Parser::CurrentRuby.parse("Foo::SUP").children.first.location
=> #<Parser::Source::Map::Constant:0x007fa01e222ed0 @name=#<Parser::Source::Range
 (string) 0...3>, @double_colon=nil, @expression=#<Parser::Source::Range (string) 0...3>,
 @node=s(:const, nil, :Foo)>
```

I wasn't sure if the difference was intentional and considered opening an issue, but the PR seemed more friendly and equally useful as a discussion point.

Thanks!

ABaldwinHunter:
Hm after spending some more time with this problem, I realize some of the trickiness with the `location` might be that we're kind of morphing `__ENCODING__` into `Encoding::UTF_8` (which parses a bit differently).

Going to close and open an issue instead.

PR #370, opened 2017-09-21
Fix reference output (doc only)

marcandre:
Fixes some `ruby-parse` outputs given for reference only that are clearly very far from current output.

Not related, but I can't resist saying how impressive this gem is. 🙇

### DIFF
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -660,10 +660,10 @@ Ruby_parser output for reference:
 
 ~~~
 "@a.b += 1"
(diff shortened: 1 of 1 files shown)


marcandre:
Oh my, didn't realize the doc was giving `ruby_parser`'s output. Sorry for the noise.

whitequark:
@marcandre Actually, it's good that you noticed this, I have no idea why AST_FORMAT has ruby_parser's output. I added this myself according to git blame, not that I have recollection from five years ago.

Could you remove all that?

marcandre:
My pleasure #371

whitequark:
Thanks!

PR #373, opened 2017-09-29
Travis: skip extra gem update --system

olleolleolle:
This PR tries to make the CI a little more efficient.

  - rvm nowadays update rubygems on its own, at install of a Ruby
  - `gem install bundler` will install latest (_and_ if the version of Ruby does not ship with a Bundler, such as perhaps jruby-head, for example, the update command may fail)


whitequark:
Thanks!

Issue #380, opened 2017-10-19
ruby/ruby@4a375f3e


whitequark:
See upstream commit https://github.com/ruby/ruby/commit/4a375f3ef518797bc8788abf0891655d79aa719c. We need no warning, just the grammar change.

iliabylich:
@whitequark What exactly should be changed? The following code:
``` ruby
$a = Object.new

a = Object.new
b = Object.new
c = Object.new

def a.<(*); puts "a < b"; $a; end
def $a.<(*); puts ("(a < b) < c"); end

puts(a < b < c)
```

prints exactly the same output on ruby 2.4 and 2.5.0dev (from ruby trunk). There where no changes in operators precedence. Or do you mean the replication of rules structure (new `rel_expr/relop` rules)?

PR #381, opened 2017-10-19
Extracted 'relop' rule to simplify comparisons.

### LINKED ISSUE #379: ruby/ruby@c1c3d073
See the upstream commit https://github.com/ruby/ruby/commit/c1c3d07332d32f0981d6019c3f851e51739f0eb2.

iliabylich:
Closes #379, #380.


whitequark:
I believe this will only be in 2.5 so you'll need to split off ruby25.y first. Look at the commit introducing ruby24.y to see how that is done.

whitequark:
Also, commits starting with `*` get changelog entries, and these two don't, so I would use something like

    ruby25.y: extract rel_expr rule.

    This commit tracks upstream commits  ruby/ruby@c1c3d07 and ruby/ruby@4a375f3.

(Squash the two since it makes little sense to keep them separate.)

iliabylich:
@whitequark Done.

whitequark:
Thanks. Can you rebase?

iliabylich:
@whitequark What do you mean? There's nothing to rebase. This branch has 3 commits on top of the latest master.

whitequark:
I mean fold df3323c into bb5c54d to get clean history, so that bisecting would never have to skip a broken commit.

iliabylich:
@whitequark Done

whitequark:
Thanks!

PR #372, opened 2017-09-29
Rake task to make CHANGELOG.md: Avoid wrong release v2.4.0.0 (2016-01-16) [sic]

olleolleolle:
~This PR **drops** the heading for **a non-existent release** v2.4.0.0 (in _2016_) from the change log.~

  - ~it had wrongly been `v2.3.3.1 (2016-01-16)` but then wrongly replaced with `v2.4.0.0 (2016-01-16)`~
  - ~this edit removes the heading, so that the releases listed in this document mirrors the released Git tags~

**Update**: What needs to be done is to avoid this mis-rendering by the Rake task which produces the change log.

PR #382, opened 2017-10-24
Change log rake task should not generate tag sections w/o commits

rgo:
It tries to fix the PR: https://github.com/whitequark/parser/pull/372

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,19 @@
 Changelog
 =========
 
+Not released (2017-10-24)
+-------------------------
+
+API modifications:
+ * parser/current: update for Ruby 2.1.10 and 2.2.7. (Koichi ITO)
+
+Features implemented:
+ * ruby25.y: branch parser. (Ilya Bylich)
+
+Bugs fixed:
+ * ruby-rewrite: allow passing paths to --load. (whitequark)
+ * builders/default: "def x; else; end": don't crash. (whitequark)
+
 v2.4.0.0 (2017-02-07)
 ---------------------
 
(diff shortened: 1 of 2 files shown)


whitequark:
Can you also regenerate the changelog so I see what the actual difference is?

rgo:
Because it's big I uploaded it to gist: https://gist.github.com/rgo/7de4b499a7ca6f972eba93dbefd70d43

whitequark:
No, I mean, regenerate the changelog and include it in the PR. `rake changelog` and then push.

rgo:
Give some time because after regenerate changelog I've seen something wrong.

Tag 2.4.0.0 is  duplicated with different date:

screen shot 2017-10-25 at 11 04 58

rgo:
I saw the problem.

[These lines of code](https://github.com/rgo/parser/blob/2155a543641045a2480e93a40d5f01b3b4509bdd/Rakefile#L102-L104)[1]:

```
...
      current_version = "#{$1} (#{date})" if version =~ /(v[\d\w.]+)/
      current_version = "v#{Parser::VERSION} (#{date})" \
        if version =~ /(^| |\/)#{Regexp.escape branch}$/
...
```

If version matches with version format sets `current_version` as it.
If version matches with branch name sets as version Parser::VERSION that right now is the same that the current one.

It makes a duplicated entry with the same version but different date as you can see in my previous comment.

If you like:

  -  I can leave it as is knowing why it happens
  - Add some text which indicate is not released yet like "v2.4.0.0 (2017-10-24) - Not released"
  - Remove version number and change "Not released (2017-10-24)"

What would you like? perhaps you have any other idea

[1] https://github.com/rgo/parser/blob/2155a543641045a2480e93a40d5f01b3b4509bdd/Rakefile#L102-L104

whitequark:
Thanks for investigating!

> Remove version number and change "Not released (2017-10-24)"

This option sounds good.

rgo:
When you do a new release check the changelog commit/file to ensure is right (I hope so).

PR #387, opened 2017-11-08
Skip magic comments

### LINKED ISSUE #386: Should Comment::Associator ignore frozen_string_literal?
I believe that the `skip_directives` attribute on `Comment::Associator` should also ignore the `frozen_string_literal` directive.

Given

```ruby
require 'parser/current'

ast, comments = Parser::CurrentRuby.parse_with_comments(<<-CODE)
# frozen_string_literal: true

def hi
end
CODE

p Parser::Source::Comment.associate(ast, comments)
```

… then the output is

```
{
  s(:def, :hi, s(:args), nil) =>[
    #<Parser::Source::Comment (string):1:1 "# frozen_string_literal: true">
  ]
}
```

… but _should_ be (I believe) empty.

I couldn’t find any other mention of this in the source code, commits or issues, so I don’t think this has come up before.

ddfreyne:
This makes the comment associator ignore the following magic comments:

* `frozen_string_literal`
* `warn_indent`
* `warn_past_scope`

Fixes #386.

Questions:

* `MAGIC_COMMENT_RE` and `Buffer::ENCODING_RE` are very similar. I’d prefer for them to be combined somehow, but I can’t think of a good way to do it. Might it make sense to move `MAGIC_COMMENT_RE` into `Buffer`, at least?

* `warn_past_scope` is only activated when `WARN_PAST_SCOPE` is set when compiling Ruby (it isn’t by default). Given that, does it still make sense to support it?

### DIFF
--- a/lib/parser/source/comment/associator.rb
+++ b/lib/parser/source/comment/associator.rb
@@ -184,9 +184,16 @@ module Parser
         advance_comment
       end
 
+      MAGIC_COMMENT_RE = /^#\s*(-\*-|)\s*(frozen_string_literal|warn_indent|warn_past_scope):.*\1$/
+
       def advance_through_directives
         # Skip shebang.
         if @current_comment && @current_comment.text =~ /^#!/
+          advance_comment
+        end
+
+        # Skip magic comments.
+        if @current_comment && @current_comment.text =~ MAGIC_COMMENT_RE
           advance_comment
         end
 
--- a/test/test_source_comment_associator.rb
+++ b/test/test_source_comment_associator.rb
@@ -213,6 +213,96 @@ end
     assert_equal 0, associations.size
   end
 
+  def test_associate_frozen_string_literal
+    ast, associations = associate(<<-END)
+# frozen_string_literal: true
+class Foo
+end
+    END
+
+    assert_equal 0, associations.size
+  end
+
+  def test_associate_frozen_string_literal_dash_star_dash
+    ast, associations = associate(<<-END)
+# -*- frozen_string_literal: true -*-
+class Foo
+end
+    END
+
+    assert_equal 0, associations.size
+  end
+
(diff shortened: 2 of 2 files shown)


whitequark:
> MAGIC_COMMENT_RE and Buffer::ENCODING_RE are very similar. I’d prefer for them to be combined somehow, but I can’t think of a good way to do it. Might it make sense to move MAGIC_COMMENT_RE into Buffer, at least?

Don't bother.

whitequark:
> warn_past_scope is only activated when WARN_PAST_SCOPE is set when compiling Ruby (it isn’t by default). Given that, does it still make sense to support it?

Sure. Let's be conservative here.

lib/parser/source/comment/associator.rb
```diff
@@ -184,12 +184,19 @@ def associate_and_advance_comment(node)
...
 
+      MAGIC_COMMENT_RE = /^#\s*(-\*-|)\s*(frozen_string_literal|warn_indent|warn_past_scope):.*\1$/
+
       def advance_through_directives
         # Skip shebang.
         if @current_comment && @current_comment.text =~ /^#!/
           advance_comment
         end
 
+        # Skip magic comments.
```
whitequark: I thought you could put multiple directives in a single magic comment. Is that not so?

backus: Yeah you can. Parser does it https://github.com/whitequark/parser/blob/a31d2c0926fb084a3afe06feb0473e8ca5e398b7/Rakefile#L37

Here it is in the parse.y https://github.com/ruby/ruby/blob/3ba9e6b639afaf4c88145d56c9bd6746af16cfb7/parse.y#L7188-L7207

ddfreyne: My implementation ignores the entire comment line if it matches any magic comment, so that should be fine.

PR #389, opened 2017-11-12
use String#start_with? instread of Regexp match

walf443:
String#start_with? is faster than Regex match If it does not need Regexp match.

### DIFF
--- a/lib/parser/source/comment/associator.rb
+++ b/lib/parser/source/comment/associator.rb
@@ -188,7 +188,7 @@ module Parser
 
       def advance_through_directives
         # Skip shebang.
-        if @current_comment && @current_comment.text =~ /^#!/
+        if @current_comment && @current_comment.text.start_with?('#!')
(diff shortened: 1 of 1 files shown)


lib/parser/source/comment/associator.rb
```diff
@@ -188,7 +188,7 @@ def associate_and_advance_comment(node)
 
       def advance_through_directives
         # Skip shebang.
-        if @current_comment && @current_comment.text =~ /^#!/
+        if @current_comment && @current_comment.text.start_with?('#!')
```
mbj: Note there is a small semantic difference between them: `^#!` would match ANY line that begins with `#!` not only the first. AFAIK the `@current_comment` is only a single line so the difference should not matter?

walf443: shebang shoud be located at first line of a file. So it always should match as `/\A#!/`, I think.

Issue #337, opened 2017-01-19
[Feature #12906]


whitequark:
See the upstream commit https://github.com/ruby/ruby/commit/0ec889d7ed34f80b534dfc7a93bdd3175aba9ff9 and Ruby bug https://bugs.ruby-lang.org/issues/12906.

PR #392, opened 2017-11-13
Allow rescue/else/ensure inside do/end blocks. [Feature #12906]

iliabylich:
Closes https://github.com/whitequark/parser/issues/337

### DIFF
--- a/lib/parser/ruby25.y
+++ b/lib/parser/ruby25.y
@@ -1603,7 +1603,7 @@ opt_block_args_tail:
                       result = @lexer.cmdarg.dup
                       @lexer.cmdarg.clear
                     }
-                    opt_block_param compstmt
+                    opt_block_param bodystmt
                     {
                       result = [ val[2], val[3] ]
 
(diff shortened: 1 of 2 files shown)


iliabylich:
@whitequark Does it make sense to add more tests to cover cases with `else` and `ensure`?

whitequark:
Not really, my general rule is one test per one codepath, and you're only changing one here.

whitequark:
I see you're fixing the issues in the bugtracker. Please note that the condition for a Parser release is all relevant (as per GitHub issues and position in an upstream branch) commits being in master. Once those are done I will be glad to cut a release.

iliabylich:
Could you please organize issues (maybe using tags) to clearly show which of them are required to release a new version (2.4/2.5)?

whitequark:
Sure: https://github.com/whitequark/parser/issues?q=is%3Aissue+is%3Aopen+label%3Ablocking-2.4

Note that it is very easy to determine which branches/releases a commit is in, provided that it's old enough. Of course, the information from the Ruby bug tracker takes precedence here.
screenshot_20171113_170925

Issue #330, opened 2016-12-04
ruby/ruby@80045bc + ruby/ruby@7e8b910


whitequark:
See https://github.com/ruby/ruby/commit/80045bc2a9e12012d6d16517ea5cb037e67eb2c1.

JoshCheek:
Would this one fall under https://github.com/whitequark/parser/issues/328#issuecomment-264353903

If so, can you describe what it would take to close the issue, or link me to a similar one I can use as a reference?

My guess: add a test showing that it parses the spacing on 2.4 (maybe nearby [this](https://github.com/whitequark/parser/blob/ecbdf65442abe8c1d24dfa28e472189a1156a155/test/test_parser.rb#L5143-L5149) test, or maybe add it to lexer tests), but not older Rubies. Then fix it to allow the spacing on 2.4, somewhere around [here](https://github.com/whitequark/parser/blob/ecbdf65442abe8c1d24dfa28e472189a1156a155/lib/parser/lexer.rl#L733-L743).

whitequark:
@JoshCheek Actually, yes!

> add a test showing that it parses the spacing on 2.4 (maybe nearby this test, or maybe add it to lexer tests

This is just a lexer feature, so yes. I usually copy the MRI tests nearly verbatim, with minimal necessary adjustment to the Parser testsuite.

> Then fix it to allow the spacing on 2.4, somewhere around here.

Yup.

This particular case is a bit odd because we cannot easily branch on Ruby version in the lexer rules. So, I think an easy (although somewhat dirty) way to fix it is as follows: add a copy of the Unicode lexing rules under the existing ones; change them to accept multiple whitespace characters in a row; and add a fatal diagnostic like the one that's already there. In this way, old-style literals will trigger the old rules, and new-style literals will trigger the new rules, which will error out on old Ruby targets.

whitequark:
Also ruby/ruby@7e8b910.

iliabylich:
Looks like an issue with zero codepoints is not fixed. The test was removed in https://github.com/ruby/ruby/commit/7e8b910, but it still passes:
``` sh
$ ./ruby -v test.rb

```

whitequark:
Depressing.

iliabylich:
Fixed in upstream in https://github.com/ruby/ruby/commit/50702d16ab5862413351520b65da820194dbf8d7 but not ported yet to 2.4 / 2.5 branches

PR #394, opened 2017-11-15
Relax restriction spaces inside "\u{...}".

iliabylich:
Closes https://github.com/whitequark/parser/issues/330.

This is my first contribution to lexer, so I need some help:
1. ~~There's 1 failing spec - `"\u{}"` should produce an error, there's an explicit machine (line 738) that handles it, but for some reason ragel doesn't enter it (I'm still reading ragel guide, so I guess I simply don't understand something). This rule is on top of all `u{` rules, and it matches a string content completely. Ragel emits an unescaped string.~~
2. Is there some style guide for ragel syntax? I've tried to write rules using existing code style, is it ok?
3. Also I have a feeling that probably it makes sense to extract parts related to whitespace handling to separate machines (something like `optional_surrounding_spaces_in_unicode_codepoints` / `spaces_in_unicode_codepoints`, as those rules are different).

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -657,17 +657,37 @@ class Parser::Lexer
     codepoints  = tok(@escape_s + 2, p - 1)
     codepoint_s = @escape_s + 2
 
-    codepoints.split(/[ \t]/).each do |codepoint_str|
-      codepoint = codepoint_str.to_i(16)
-
-      if codepoint >= 0x110000
-        diagnostic :error, :unicode_point_too_large, nil,
-                   range(codepoint_s, codepoint_s + codepoint_str.length)
-        break
+    if @version < 24
+      if codepoints.start_with?(" ") || codepoints.start_with?("\t")
+        diagnostic :fatal, :invalid_unicode_escape, nil,
+          range(@escape_s + 2, @escape_s + 3)
       end
 
-      @escape     += codepoint.chr(Encoding::UTF_8)
-      codepoint_s += codepoint_str.length + 1
+      if spaces_p = codepoints.index(/[ \t]{2}/)
+        diagnostic :fatal, :invalid_unicode_escape, nil,
+          range(codepoint_s + spaces_p + 1, codepoint_s + spaces_p + 2)
+      end
+
+      if codepoints.end_with?(" ") || codepoints.end_with?("\t")
+        diagnostic :fatal, :invalid_unicode_escape, nil, range(p - 1, p)
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


whitequark:
Oh dear, that's one complicated change. It might take me a while, so please feel free to work on others in the meantime.

iliabylich:
@whitequark What do you think about changing rule to match both `"\u{1}"` and `"\u{ \t 1 \t }"` and validating it later in `uncode_points` action?

iliabylich:
The validation can be something like:
``` ruby
if @version < 24 && (codepoints.start_with?(" ") || codepoints.start_with?("\t"))
  diagnostic :fatal, :invalid_unicode_escape, nil, <range pointing to the first character>
end

if @version < 24 && (codepoints.end_with?(" ") || codepoints.end_with?("\t"))
  diagnostic :fatal, :invalid_unicode_escape, nil, <range pointing to the last character>
end

codepoints.each_char do |char|
  if <two spaces in a row> && @version < 24
    diagnostic :fatal, :invalid_unicode_escape, nil, <range pointing to the second space>
    break
  end

  <other code for unicode processing>
end
```

By making validation on-the-fly we could simplify the ragel rule, because right now it's overcomplicated. 

Then the following part:
```
( zlen
  | ( [ \t]+
      > { @u_whitespaces_p = p }
      % ensure_spaces_are_allowed_in_unicode_points
    )
)
```
could be simply rewritten to `[ \t]*`. And the same place for spaces between codepoints can be changed `[ \t]+`. The cons here is that `unicode_points` action becomes more complex.  What do you think?

whitequark:
Let's do it.

iliabylich:
@whitequark Done, could take a look please?

lib/parser/lexer.rl
```diff
@@ -657,17 +657,37 @@ class Parser::Lexer
...
-      if codepoint >= 0x110000
-        diagnostic :error, :unicode_point_too_large, nil,
-                   range(codepoint_s, codepoint_s + codepoint_str.length)
-        break
+      if codepoints.end_with?(" ") || codepoints.end_with?("\t")
+        diagnostic :fatal, :invalid_unicode_escape, nil, range(p - 1, p)
       end
+    end
+
+    codepoints.scan(/([0-9a-hA-H]+)|([ \t]+)/).each do |(codepoint_str, spaces)|
```
whitequark: `[a-h]`? Shouldn't it be `[a-f]`?

iliabylich: Fixed 😄

Issue #334, opened 2016-12-16
ruby/ruby@d25faa4


whitequark:
See https://github.com/ruby/ruby/commit/d25faa4ed2d850f086aff7ec2aa40c9f50e168db.

PR #395, opened 2017-11-15
Allow newlines in heredoc identifiers.

iliabylich:
Closes https://github.com/whitequark/parser/issues/334


whitequark:
I think a better changelog entry would be "Reject newlines in heredoc messages", not accept, since that's the eventual goal.

iliabylich:
@whitequark Before this change lexer didn't support newlines in heredoc identifiers. Now it does, but with some restrictions. After all this being said do you still want me to change the message?

PR #399, opened 2017-11-18
Change relative order of insert_after_multi and insert_before_multi

marcandre:
While the order of multiple calls to either `insert_after_multi` or `insert_before_multi` is well tested and documented, the relative order of `insert_after_multi` and `insert_before_multi` done at the same point isn't.

My use case is I wanted to rewrite:
```
foo.bar
```
into:
```
<span class="...">foo</span><span>.</span><span ...>bar</span>
```

To do so, I inserted my opening tags before each range, and my closing tags after. This gives me the wrong result! I'd have to insert my opening tags using `insert_after` and my closing tags using `insert_before`!

This PR fixes it by prioritizing `insert_after` over `insert_before`. This is a potentially breaking change, but neither the doc nor the tests condone the current behavior. 

Implementation isn't super pretty, but I favored it over changing `Change#<=>` to account for signs because this relative order is not `Change`'s responsibility but the `Rewriter`'s.

### DIFF
--- a/lib/parser/source/rewriter.rb
+++ b/lib/parser/source/rewriter.rb
@@ -71,16 +71,40 @@ module Parser
         append Rewriter::Action.new(range.begin, content)
       end
 
+      INSERT_GROUP_SEPARATOR = 0x1fff_ffff # Any big enough integer will do. This one fits in a Fixnum
+
       ##
       # Inserts new code before the given source range by allowing other
       # insertions at the same position.
-      # Note that an insertion with latter invocation comes _before_ earlier
-      # insertion at the same position in the rewritten source.
+      # Note that multiple insertions at the same position are done in groups:
+      #  - insert_after_multi (with non empty ranges ending at insertion point)
+      #  - insert_before_multi at insertion point (i.e. empty range)
+      #  - insert_after_multi at insertion point (i.e. empty range)
+      #  - insert_before_multi (with non empty ranges beginning at insertion point)
       #
-      # @example Inserting '[('
+      #  Within each group of `insert_before_multi`, the insertions are done
+      #  in the _reverse_ order they were given, while each group of `insert_after_multi` is
+      #  processed in the _same_ order they were given.
+      #
+      # @example Inserting '[(' before and ')]' after a range:
       #   rewriter.
       #     insert_before_multi(range, '(').
+      #     insert_after_multi(range, ')').
       #     insert_before_multi(range, '[').
+      #     insert_after_multi(range, ']').
+      #     process
+      #
+      # @example Inserting '>{}<'
+      #   insertion_point = range.end
+      #   # Assume that range and other_range non empty ranges such that
+      #   insertion_point == other_range.begin # => true
+      #   # The following will insert '>{}<' for any order of the calls to `insert...`:
+      #
+      #   rewriter.
+      #     insert_after_multi(range, '>').
+      #     insert_before_multi(insertion_point, '{').
+      #     insert_after_multi(insertion_point, '}').
+      #     insert_before_multi(other_range, '<').
       #     process
       #
       # @param [Range] range
@@ -90,7 +114,8 @@ module Parser
       #
       def insert_before_multi(range, content)
         @insert_before_multi_order -= 1
-        append Rewriter::Action.new(range.begin, content, true, @insert_before_multi_order)
+        group_delta = range.empty? ? 0 : INSERT_GROUP_SEPARATOR
+        append Rewriter::Action.new(range.begin, content, true, @insert_before_multi_order + group_delta)
       end
 
       ##
@@ -108,14 +133,7 @@ module Parser
       ##
       # Inserts new code after the given source range by allowing other
       # insertions at the same position.
-      # Note that an insertion with latter invocation comes _after_ earlier
-      # insertion at the same position in the rewritten source.
-      #
-      # @example Inserting ')]'
-      #   rewriter.
-      #     insert_after_multi(range, ')').
-      #     insert_after_multi(range, ']').
-      #     process
+      # See `insert_before_multi` for examples and details on the order.
       #
       # @param [Range] range
       # @param [String] content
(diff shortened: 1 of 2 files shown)


marcandre:
If you'd rather keep the current behavior, I can tweak my PR to only change the doc and add the relevant tests.

lib/parser/source/rewriter.rb
```diff
@@ -74,8 +75,10 @@ def insert_before(range, content)
       ##
       # Inserts new code before the given source range by allowing other
       # insertions at the same position.
-      # Note that an insertion with latter invocation comes _before_ earlier
-      # insertion at the same position in the rewritten source.
+      # Note that multiple insertions at the same position are done in the
+      # following order:
+      #   - insertions done with insert_after_multi (in order)
+      #   - insertons done with insert_before_multi (in _reverse_ order).
```
alexdowad: Typo, "insertons" -> "insertions"

lib/parser/source/rewriter.rb
```diff
@@ -108,8 +111,10 @@ def insert_after(range, content)
       ##
       # Inserts new code after the given source range by allowing other
       # insertions at the same position.
-      # Note that an insertion with latter invocation comes _after_ earlier
-      # insertion at the same position in the rewritten source.
+      # Note that multiple insertions at the same position are done in the
+      # following order:
+      #   - insertions done with insert_after_multi (in order)
+      #   - insertons done with insert_before_multi (in _reverse_ order).
```
alexdowad: Same typo here.

alexdowad:
Prioritizing `insert_after` over `insert_before` makes sense.

marcandre:
@alexdowad Thanks :-) Typos fixed.

JoshCheek:
If this gets merged, can we version it as a breaking change? My lib depends on the order ([here](https://github.com/JoshCheek/seeing_is_believing/blob/3f7910b904edfff302452daa0d5f91f6b9336d8e/lib/seeing_is_believing/wrap_expressions.rb#L22-L49)). Additional context:  https://github.com/whitequark/parser/pull/259, and https://github.com/whitequark/parser/pull/277

whitequark:
Parser does not break APIs. This needs to be done with an opt-in field, e.g. `Rewriter.insert_before_then_insert_after = true`.

marcandre:
Mmm, I'm now thinking we might want to distinguish `insert_after_multi` with an empty range and non empty range. Let me see what I can come up with.

marcandre:
So I've changed my PR so that `insert_before_multi` & `insert_after_multi` are split in 4 groups (for any given insertion point *P*)
* insert_after a non-empty range ending at *P*
* insert_before *P*
* insert_after *P*
* insert_before a non-empty range beginning at *P*

While this complicates the documentation, I believe it makes all scenarios work well: my use case of wrapping non empty ranges, as well as wrapping an empty range (that I always think of as an insertion point), or both.

@JoshCheek FWIW, the test case of `seeing_is_believing` that was failing with my previous PR now passes with this one. Thanks for bringing this up!

@whitequark I see this version as a bug fix, not an API change, but final word is yours. Would you still like a preference for this? I'm not sure what to call it tbh... Maybe `Rewriter.grouped_multiple_insertion`, although it was already grouped, just not well enough.

whitequark:
Your rewritten PR looks much more like a bugfix. I want at least one other review though. @JoshCheek ?

whitequark:
Also @alexdowad ?

alexdowad:
I think the modified behavior of `Rewriter` as modified by @marcandre is much more intuitive and makes more sense than the existing behavior.

The existing behavior in this situation (where `insert_before` and `insert_after` add text at the same point) was not intentional. Rather, the behavior of `Rewriter` was underspecified.

lib/parser/source/rewriter.rb
```diff
@@ -124,7 +140,8 @@ def insert_after(range, content)
       #
       def insert_after_multi(range, content)
         @insert_after_multi_order += 1
-        append Rewriter::Action.new(range.end, content, true, @insert_after_multi_order)
+        group_delta = range.empty? ? 0 : 0x1fff_ffff
+        append Rewriter::Action.new(range.end, content, true, @insert_after_multi_order - group_delta)
```
whitequark: Can we make this a bit more elegant? Pull `0x1fffffff` out and into a constant, at least.

marcandre:
Constant added.
While this PR uses the order to form the four groups (very negative numbers, negative numbers, positive numbers, highly positive numbers), I think that the grouping will have to be pushed to the `Action` for #400 (assuming it gets a green light)

whitequark:
CI fail

marcandre:
Oh, right, 1.9.2 is still supported... Removed call to `private_constant`.

whitequark:
Oh, bleh, should've asked you to annotate the commit to get a changelog entry. I'll do it myself.

JoshCheek:
This passes for me.

I'm not sure I understand the purpose of the empty range check. Is it that an empty range is smaller than a nonempty range, so therefore falls within the nonempty range, implying it is on the inside, regardless of the order it was applied? If so, it seems like that should extend to nonempty ranges where one of the ends falls within the other. I do recognize that this would be a lot more effort to track, just asking as I'm not sure I understand :)

marcandre:
@JoshCheek It took me a while to see it too, but yeah, the empty range is special case. Try imagining what should happen if you first replaced the empty range with come content, then did the `insert_after_multi`, then removed the content you added.

Here's another try at explaining it:
For a given position *P*, there can be three non-intersecting ranges with one end point at *P*:
`<P-n...P>` | `<P...P>` | `<P...P + m>`
This creates 4 possible insertion points (denoted with  🔽 ):
`<P-n...P>  🔽 ` | ` 🔽  <P...P>  🔽 ` | ` 🔽  <P...P + m>`

The change distinguishes those four insertion points.

I'm also proposing the same for the non-`multi` versions in #400. Did you have a look at it? I think the example with the `replace` vs the `inserts` may help to visualize this.

HTH

marcandre:
> If so, it seems like that should extend to nonempty ranges where one of the ends falls within the other.

I think you mean that `<4...10>.insert_before` should always be done before `<4...20>.insert_before`, right? I agree. It wouldn't be much more work, but it would be a breaking change though.

This current change is different in that it affects the relative order of `before` vs `after` at the same point.

PR #404, opened 2017-11-21
+ Add Range#with. Use it where convenient

marcandre:
Whenever I work with ranges, I find that I'm always creating ranges from other ranges, tweaking either end points, sometimes both.

I'm proposing `Range#with` to easily update either or both end points.

This not only saves keystrokes, it makes it easier to read and see what is being changed.

I've used it 8 times in `parser` itself.

I'm counting about 20 uses in Rubocop (not counting the specs), plus eliminating [`slice_source`](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/format_string_token.rb#L126-L132) which it makes redundant.

PS: I was waiting for Ruby 1.9 support to die, but this version can do until then.

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -1207,13 +1207,9 @@ module Parser
     def delimited_string_map(string_t)
       str_range = loc(string_t)
 
-      begin_l = Source::Range.new(str_range.source_buffer,
-                                  str_range.begin_pos,
-                                  str_range.begin_pos + 1)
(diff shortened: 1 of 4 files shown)


marcandre:
Oh, and there are over 100 uses of [`range_between` in Rubocop](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/util.rb#L122-L124), many of which could be simplified with `with`.

alexdowad:
Looks nice. I just hope it doesn't noticeably increase the runtime of
Rubocop.

PR #406, opened 2017-11-23
Kill 1.8

### LINKED ISSUE #375: It's time for 1.8/1.9 support in Parser to die
- [x] Remove all 1.8-specific feature checks
 - [x] Remove all 1.9-specific feature checks because it makes no sense to keep that in either
 - [x] Update README to say "runs on all versions of Ruby since 2.0, and historically, all versions of Ruby since 1.8"

alexdowad:
Fixes #375


bbatsov:
👍 I think these cleanup was long overdue.

lib/parser/ruby19.y
```diff
@@ -2147,8 +2147,6 @@ end
 
 require 'parser'
 
-Parser.check_for_encoding_support
```
bbatsov: Why didn't you delete the entire file?

bbatsov:
Btw, MacRuby is a dead project, that was only 1.9 compatible, so I guess killing support for it also makes sense.

alexdowad:
@bbatsov It's not clear whether @whitequark wants to kill support for
parsing Ruby 1.8 code, or for running the parser on a 1.8 VM. I assume it's
the latter.

lib/gauntlet_parser.rb
```diff
@@ -69,13 +67,7 @@ def parse(name)
...
 
       try(Parser::Ruby20, RUBY20, file) do
-        puts "Trying 1.9:"
-        try(Parser::Ruby19, RUBY19, file, show_ok: true) do
-          puts "Trying 1.8:"
-          try(Parser::Ruby18, RUBY18, file, show_ok: true) do
-            puts "Invalid syntax."
-          end
-        end
+        puts "Invalid syntax."
```
whitequark: This should actually stay in so we don't barf on 1.8-only gems.

lib/parser/current.rb
```diff
@@ -9,24 +9,6 @@ def warn_syntax_deviation(feature, version)
   end
 
   case RUBY_VERSION
-  when /^1\.8\./
```
whitequark: This should be kept and raise an error.

lib/parser/current.rb
```diff
@@ -9,24 +9,6 @@ def warn_syntax_deviation(feature, version)
...
-    current_version = '1.8.7'
-    if RUBY_VERSION != current_version
-      warn_syntax_deviation 'parser/ruby18', current_version
-    end
-
-    require 'parser/ruby18'
-    CurrentRuby = Ruby18
-
-  when /^1\.9\./
```
whitequark: Same here.

lib/parser/current.rb
alexdowad: Wouldn't it be better to put something in `lib/parser.rb` which either raises an error or prints a warning if run on 1.8/1.9?

If we put it in `lib/parser/current.rb`, the error will only be seen if that file is loaded.

whitequark: Sure, that's better.

bbatsov:
Seems you're right. I see little point in keeping the parsing support around, though, as there are almost no 1.8-compatible projects around these days and just a handful of 1.9-compatible ones. Dropping support for those encourages laggards to upgrade IMO.

whitequark:
> Dropping support for those encourages laggards to upgrade IMO.

One person's old junk is another person's history. I'm not in the business of making value judgements over here.

alexdowad:
Updated as per review from @whitequark. Let me know if version check in `parser.rb` looks ugly. Now just rebasing on the new `master`.

PR #410, opened 2017-11-27
Range relationship

marcandre:
This first fixes the edge case for equal and empty ranges for `disjoint?`, `overlaps?` and `intersect`.

Then, it adds `contains?`, `contained?` and `crossing?`, in such a way that any two ranges are one and only one of `==`, `disjoint?`, `contains?`, `contained?` or `crossing?

*Note*: Dealing with "empty" ranges can be confusing. I find it helps a lot to translate Range(1, 4) to 1..3 and Range(4, 4) to 3.5
At first I was thinking of them as actually empty; nothing was working and my head was close to exploding.

(This is a prelude to my `TreeRewriter` PR)

### DIFF
--- a/lib/parser/source/range.rb
+++ b/lib/parser/source/range.rb
@@ -209,7 +209,8 @@ module Parser
       # @return [Boolean] `true` if this range and `other` do not overlap
       #
       def disjoint?(other)
-        @begin_pos >= other.end_pos || other.begin_pos >= @end_pos
+        (@begin_pos >= other.end_pos || other.begin_pos >= @end_pos) &&
+          (!empty? || !other.empty?)
       end
 
       ##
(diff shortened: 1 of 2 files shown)


lib/parser/source/range.rb
```diff
@@ -205,19 +205,63 @@ def intersect(other)
...
+      # Two ranges must be one and only one of ==, disjoint?, contains?, contained? or crossing?
+      #
       # @param [Range] other
-      # @return [Boolean] `true` if this range and `other` do not overlap
+      # @return [Boolean]
       #
       def disjoint?(other)
-        @begin_pos >= other.end_pos || other.begin_pos >= @end_pos
+        (@begin_pos >= other.end_pos || other.begin_pos >= @end_pos) &&
+          (!empty? || !other.empty?)
```
whitequark: Hmm, this is an API breaking change, isn't it?

whitequark: Oh nevermind, I see this is a bugfix.

PR #411, opened 2017-11-27
Utilities

marcandre:
Adds `Range#adjust` to get a new range from offsets.
Adds `Buffer#source_range`, a trivial method to get a source range from a Buffer.

I could split in two PRs if need be

(Prelude to TreeRewriter. Not needed per say, just thought they were "natural" features that could be useful to others)


whitequark:
Please rebase.

marcandre:
> details irrelevant for public API away from the first line in the commit message

Oh, right.

Done.

Issue #359, opened 2017-07-26
[Bug #13547]


whitequark:
See upstream commit https://github.com/ruby/ruby/commit/7d6965f88d6abf0e1e9f92a58139dc56b27db07a and Ruby bug https://bugs.ruby-lang.org/issues/13547.

iliabylich:
@whitequark And I guess also https://github.com/ruby/ruby/commit/99871095e6b94d4851daec10ccc601bd3b23316f ? Did you backport it? The code in master can parse tests from this commit:
<details>
  <summary>Click to expand</summary>

```
$ cat test.rb
'm "x" {}'
'm 1 {}'
'm 1.0 {}'
'm :m {}'
'm :"#{m}" {}'
'm ?x {}'
'm %[] {}'
'm 0..1 {}'
'm [] {}'
'm /r/ {}'
$ cat test.rb | xargs -n 1 ruby-parse --24 -e
(block
  (send nil :m
    (str "x"))
  (args) nil)
(block
  (send nil :m
    (int 1))
  (args) nil)
(block
  (send nil :m
    (float 1.0))
  (args) nil)
(block
  (send nil :m
    (sym :m))
  (args) nil)
(block
  (send nil :m
    (dsym
      (begin
        (send nil :m))))
  (args) nil)
(fragment:0):1:6: error: unexpected token tLCURLY
(fragment:0):1: m ?x {}
(fragment:0):1:      ^
(block
  (send nil :m
    (dstr))
  (args) nil)
(block
  (send nil :m
    (irange
      (int 0)
      (int 1)))
  (args) nil)
(block
  (send nil :m
    (array))
  (args) nil)
(block
  (send nil :m
    (regexp
      (str "r")
      (regopt)))
  (args) nil)
```
</details>

whitequark:
Then I missed it. Please backport this too.

iliabylich:
@whitequark Do you know what does `SET_LEX_STATE(EXPR_END|EXPR_ENDARG) ` in `parse.y` mean? Is it a hack to simulate two states (and thus, call two handlers to process this pseudo-state)?

whitequark:
It's probably easiest to play around with our lexer and see what set of changes the intended token stream corresponds to.

alexdowad:
Are you asking about MRI's C lexer? It has various bit flags which are
turned on and off as the lex progresses. At various points, one or another
of those flags will be tested. The code you are quoting should be turning
on 2 of those flags.

[...]

PR #412, opened 2017-12-04
Reject brace after literal arg.

iliabylich:
Closes https://github.com/whitequark/parser/issues/359


whitequark:
Nope, has to be done in the lexer or you'll miss something else.

PR #415, opened 2017-12-12
Reject brace after literal arg.

iliabylich:
Hopefully closes https://github.com/whitequark/parser/issues/359

@whitequark A couple of questions:
1. Do you see any other cases that need to be tested?
2. This behavior is common for all ruby versions (tests pass when I change `SINCE_2_4` to `ALL_VERSIONS`) and I don't see any similar tests for older versions of parser. Should I put `ALL_VERSIONS` to these tests?

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -89,8 +89,6 @@ class Parser::Lexer
 
   REGEXP_META_CHARACTERS = Regexp.union(*"\\$()*+.<>?[]^{|}".chars).freeze
 
-  RBRACE_OR_RBRACK = %w"} ]".freeze
-
   attr_reader   :source_buffer
 
   attr_accessor :diagnostics
@@ -421,10 +419,8 @@ class Parser::Lexer
     if old_literal.type == :tREGEXP_BEG
       # Fetch modifiers.
       self.class.lex_en_regexp_modifiers
-    elsif @version < 24
-      self.class.lex_en_expr_end
     else
-      self.class.lex_en_expr_endarg
+      self.class.lex_en_expr_end
     end
   end
 
@@ -1122,13 +1118,7 @@ class Parser::Lexer
         end
 
         emit(:tREGEXP_OPT)
-
-        if @version < 24
-          fnext expr_end;
-        else
-          fnext expr_endarg;
-        end
-
+        fnext expr_end;
         fbreak;
       };
 
@@ -1136,11 +1126,7 @@ class Parser::Lexer
       => {
         emit(:tREGEXP_OPT, tok(@ts, @te - 1), @ts, @te - 1)
         fhold;
-        if @version < 24
-          fgoto expr_end;
-        else
-          fgoto expr_endarg;
-        end
+        fgoto expr_end;
       };
   *|;
 
@@ -2229,10 +2215,16 @@ class Parser::Lexer
       e_rbrace | e_rparen | ']'
       => {
         emit_table(PUNCTUATION)
-        @cond.lexpop; @cmdarg.lexpop
+        @cond.lexpop; @cmdarg.pop
 
-        if RBRACE_OR_RBRACK.include?(tok)
+        if tok == '}'.freeze
           fnext expr_endarg;
+        elsif tok == ']'
+          if @version >= 24
+            fnext expr_end;
+          else
+            fnext expr_endarg;
+          end
         else # )
           # fnext expr_endfn; ?
         end
--- a/lib/parser/ruby24.y
+++ b/lib/parser/ruby24.y
@@ -850,12 +850,11 @@ rule
                     }
 
     command_args:   {
-                      result = @lexer.cmdarg.dup
                       @lexer.cmdarg.push(true)
                     }
                   call_args
                     {
-                      @lexer.cmdarg = val[0]
+                      @lexer.cmdarg.pop
 
                       result = val[1]
                     }
@@ -1462,13 +1461,11 @@ opt_block_args_tail:
                     }
                   f_larglist
                     {
-                      result = @lexer.cmdarg.dup
-                      @lexer.cmdarg.clear
+                      @lexer.cmdarg.push(false)
                     }
                   lambda_body
                     {
-                      @lexer.cmdarg = val[2]
-                      @lexer.cmdarg.lexpop
+                      @lexer.cmdarg.pop
 
                       result = [ val[1], val[3] ]
 
@@ -1616,17 +1613,11 @@ opt_block_args_tail:
       brace_body:   {
                       @static_env.extend_dynamic
                     }
-                    {
-                      result = @lexer.cmdarg.dup
-                      @lexer.cmdarg.clear
-                    }
                     opt_block_param compstmt
                     {
-                      result = [ val[2], val[3] ]
+                      result = [ val[1], val[2] ]
 
                       @static_env.unextend
-                      @lexer.cmdarg = val[1]
-                      @lexer.cmdarg.pop
                     }
 
          do_body:   {
@@ -1846,7 +1837,7 @@ regexp_contents: # nothing
                     compstmt tSTRING_DEND
                     {
                       @lexer.cond.lexpop
-                      @lexer.cmdarg.lexpop
+                      @lexer.cmdarg.pop
 
                       result = @builder.begin(val[0], val[2], val[3])
                     }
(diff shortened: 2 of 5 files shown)


whitequark:
Okay, this is much better, I'll review it closer once I have a bit of time and can dig into the lexer guts.

iliabylich:
@whitequark I've just rebased my branch on the latest master. Did you have a chance to a take a look at this PR? I guess that's the only thing that doesn't let us to release 2.4 and 2.5.

By the way, I don't like this solution. Why do you want it to be implemented in the lexer? Lexer is too low-level, it doesn't really have a full context of parsing, so it doesn't now about arrays or hashes. At the same time MRI's parser does some horrible things (it enters two states at the same time, even though I guess the state is designed to be one-of defined states) and I don't really know how to backport it and preserve semantics.

Also a quick note: `2.5.0` has an undocumented feature. It allows passing block to `m[idx] {}` (MRI < 2.5 doesn't allow it), I've added a test for this case.

whitequark:
> Why do you want it to be implemented in the lexer?

Because MRI does it in the lexer. That's sufficient. The yacc grammars stay exactly equivalent to MRI and that's the only reason they're maintainable at all; the lexer used to stay quite close to MRI but thanks to the recent refactoring it diverged significantly. Still, I do not want any further divergence.

whitequark:
Also, I'm reasonably certain there's some elegant solution you're missing. That's usually the case with lexer changes that seem ugly at first; non-greedy disambiguation is probably the only case where I couldn't coax ragel into letting me define them in some particularly nice way.

iliabylich:
@whitequark I think `cmdarg` manipulation is broken (in both MRI and parser).

Consider the following example:
``` ruby
m a(b c), "x" do end
```

`cmdarg` should be active (i.e. end with 1) when the lexer is in the command args and should be inactive (i.e. end with 0) when the lexer is not in the command args.

So, for this case I think it should be:
```
identifier "m"
identifier "a" - command args detected > pushing true > cmdarg is 1
lparen "(" - beginning of a method call detected > pushing false > cmdarg is 10
identifier "b"
identifier "c" - command args > pushing true > cmdarg is 101
... end of command args in parse.y > pop > cmdarg is  10
rparen ")" - end of the method call > pop > cmdarg is 1
string "x" - nothing
"do" that must be interpreted as kDO_BLOCK, because cmdarg is active
```

But for some reason (for both MRI and parser) `cmdarg` stack is not active, so both emit `kDO` in the `emit_do` method:
```
a b(x y), c,"x" do end
^ tIDENTIFIER "a"                               expr_cmdarg  [0 <= cond] [0 <= cmdarg]
a b(x y), c,"x" do end
  ^ tIDENTIFIER "b"                             expr_arg     [0 <= cond] [0 <= cmdarg]
a b(x y), c,"x" do end
   ^ tLPAREN2 "("                               expr_beg     [0 <= cond] [10 <= cmdarg]
a b(x y), c,"x" do end
    ^ tIDENTIFIER "x"                           expr_arg     [0 <= cond] [10 <= cmdarg]
a b(x y), c,"x" do end
      ^ tIDENTIFIER "y"                         expr_arg     [0 <= cond] [10 <= cmdarg]
a b(x y), c,"x" do end
       ^ tRPAREN ")"                            expr_end     [0 <= cond] [11 <= cmdarg]
a b(x y), c,"x" do end
        ^ tCOMMA ","                            expr_beg     [0 <= cond] [10 <= cmdarg]
a b(x y), c,"x" do end
          ^ tIDENTIFIER "c"                     expr_arg     [0 <= cond] [10 <= cmdarg]
a b(x y), c,"x" do end
           ^ tCOMMA ","                         expr_beg     [0 <= cond] [10 <= cmdarg]
a b(x y), c,"x" do end
            ^~~ tSTRING "x"                     expr_end     [0 <= cond] [10 <= cmdarg]
a b(x y), c,"x" do end
                ^~ kDO "do"                     expr_value   [0 <= cond] [10 <= cmdarg]
```

Is it a bug? And is this dual state in MRI a workaround to ignore it?

iliabylich:
By the way, `@cmdarg` is active in `emit_do` for 
``` ruby
m "x" do end
```

```
a "x" do end
^ tIDENTIFIER "a"                               expr_cmdarg  [0 <= cond] [0 <= cmdarg]
a "x" do end
  ^~~ tSTRING "x"                               expr_end     [0 <= cond] [0 <= cmdarg]
a "x" do end
      ^~ kDO_BLOCK "do"                         expr_value   [0 <= cond] [1 <= cmdarg]
a "x" do end
         ^~~ kEND "end"                         expr_end     [0 <= cond] [0 <= cmdarg]
a "x" do end
            ^ false "$eof"                      expr_end     [0 <= cond] [0 <= cmdarg]
(block
  (send nil :a
    (str "x"))
  (args) nil)
```

```
$ ruby -ye "a 'x' do end" | grep -E "cmdarg"
cmdarg_stack(set): 0 at line 11171
cmdarg_stack(push): 1 at line 2554
cmdarg_stack: 1 at line 8223
cmdarg_stack(set): 0 at line 2559
cmdarg_stack(set): 0 at line 3850
cmdarg_stack(set): 0 at line 3855
cmdarg_stack(set): 0 at line 11192
```

whitequark:
You might be on to something, yes.

iliabylich:
@whitequark How about this? Now it looks really similar to MRI, tests are the same as before.

whitequark:
This looks *much* better indeed. I'll still need to closely dig into the causes myself before I feel comfortable merging it, but it should happen much sooner.

iliabylich:
Not sure if it makes you more confident about merging this change but it doesn't break rubocop's and opal's test suites (when `Parser::Ruby24` is activated)

whitequark:
Oh, I fully trust that you implemented it in a way that accepts the same code that MRI does. What I'm worried about is rejecting the code that MRI rejects, and not hindering the evolution of the lexer/parser.

iliabylich:
@whitequark This PR removes all invocations of `cmdarg.lexpop`. Looks like there's no more `LEXPOP` in MRI: https://github.com/ruby/ruby/commit/afcbd388658badcb963f6891779ce5868974f8c8. Should we also remove `cond_stack.lexpop` from `ruby25.y`? And does it make you more comfortable now? 😄

lib/parser/lexer.rl
```diff
@@ -2229,10 +2215,16 @@ class Parser::Lexer
       e_rbrace | e_rparen | ']'
       => {
         emit_table(PUNCTUATION)
-        @cond.lexpop; @cmdarg.lexpop
+        @cond.lexpop; @cmdarg.pop
```
whitequark: Still have a lexpop here.

whitequark: Oh, you actually mentioned that.

whitequark:
So, I get the frustration with the delay, but I really need to get back in the context of the parser and make sure I understand every detail. This is pretty high on my list of priorities, even though it may not look so...

iliabylich:
@whitequark Just out of curiosity, how long is your list of priorities? 😄 
rubocop/opal users are happy with what we have in `ruby25`, because it includes all major syntax changes introduced in MRI 2.5, but further work is blocked. Could you somehow find a time to review it please?

whitequark:
> Just out of curiosity, how long is your list of priorities?

My health hasn't been the best for the last year, and I haven't had much time to dedicate to non-core projects.

iliabylich:
@whitequark Thanks a lot!

iliabylich:
@whitequark What do you think about removing lexpop from all parsers (and from `StackState`)? This would cleanup the lexer without introducing any compatibility issues.

whitequark:
I'm not sure. Are you certain this won't result in regressions? If yes, how?

iliabylich:
[MRI commit](https://github.com/ruby/ruby/commit/afcbd388658badcb963f6891779ce5868974f8c8) that removes `lexpop` says that it was originally a hack. Though I'm not sure that it's valid all versions

whitequark:
Well yes, of course it was a hack. Have you read RHG Ch. 11?

iliabylich:
Yes, the whole part 2, many times. But I still don't get the idea of `lexpop` (especially the fact that sometimes it `pop`s twice, maybe that's because of the naming).

whitequark:
Huh? When does it `pop` twice?
```
    def lexpop
      @stack = ((@stack >> 1) | (@stack & 1))
      @stack[0] == 1
    end
```

iliabylich:
My bad, just some random memories after https://github.com/whitequark/parser/pull/405/files.
Previous version was `push(pop || pop)`, which literally means `pop once if the top of the stack is 1, pop twice if the top of the stack is 0, and push the result back`. Frankly speaking I don't understand the new version. Why does it set last bit to `1`? Is it because `cmdarg` may have some garbage caused by lookahead?

whitequark:
I think the previous version actually had a bug? Not sure though, it's been a while.

Issue #367, opened 2017-09-19
ruby/ruby@50f7d64


whitequark:
[Ruby commit](https://github.com/ruby/ruby/commit/50f7d64f415aa2fa5767749e1076965308dfca05), [ruby-list message](http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-list/50578).

Note that this is a fairly major syntax change with no associated Ruby bug, even. No indication of whether it will be backported.

Right now:
```
$ ./bin/ruby-parse -e 'i (-1.3).abs'
(send nil :i
  (send
    (begin
      (float -1.3)) :abs))
$ ./bin/ruby-parse -e 'i = 0; i (-1.3).abs'
(begin
  (lvasgn :i
    (int 0))
  (send
    (send nil :i
      (float -1.3)) :abs))
```
Will be:
```
$ ./bin/ruby-parse -e 'i = 0; i (-1.3).abs'
(begin
  (lvasgn :i
    (int 0))
  (send nil :i
    (send
      (begin
        (float -1.3)) :abs)))
```

PR #417, opened 2017-12-20
Disable tLPAREN_ARG state by local variable.

iliabylich:
Closes https://github.com/whitequark/parser/issues/367
ruby/ruby@50f7d64.

### DIFF
--- a/test/test_parser.rb
+++ b/test/test_parser.rb
@@ -6278,4 +6278,41 @@ class TestParser < Minitest::Test
       %q{},
       SINCE_2_3)
   end
+
+  def test_lparenarg_after_lvar__since_25
+    assert_parses(
+      s(:send, nil, :meth,
+        s(:send,
+          s(:begin,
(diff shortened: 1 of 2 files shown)


test/test_parser.rb
```diff
@@ -6278,4 +6278,41 @@ def test_bug_do_block_in_hash_brace
       %q{},
       SINCE_2_3)
   end
+
+  def test_lparenarg_after_lvar__since_25
```
whitequark: Please add a test that the behavior before ruby25 stays the same.

iliabylich: First two tests check it. Old versions are `ALL_VERSIONS - SINCE_2_5`. Should I add constants like `BEFORE_2_5` ?

whitequark: Oh sorry, I missed that.

iliabylich: So... How about adding `BEFORE_X_Y` constants?

PR #421, opened 2017-12-22
Prohibit return in class/module body except for singleton class.

### LINKED ISSUE #396: ruby/ruby@8c15f40ac
See upstream commit ruby/ruby@8c15f40ac.

iliabylich:
Closes #396 

ruby/ruby@8c15f40

@whitequark If this solution is good enough, please, let me know if you need more tests (of if some tests are redundant).

### DIFF
--- a/lib/parser/context.rb
+++ b/lib/parser/context.rb
@@ -1,0 +1,40 @@
+module Parser
+  # Context of parsing that is represented by a stack of scopes.
+  #
+  # Supported states:
+  # + :class - in the class body (class A; end)
+  # + :sclass - in the singleton class body (class << obj; end)
+  # + :def - in the method body (def m; end)
+  # + :defs - in the singleton method body (def self.m; end)
+  # + :block - in the block body (tap {})
+  # + :lambda - in the lambda body (-> {})
+  #
+  class Context
+    attr_reader :stack
+
+    def initialize
+      @stack = []
+      freeze
+    end
+
+    def push(state)
+      @stack << state
+    end
+
+    def pop
+      @stack.pop
+    end
+
+    def reset
+      @stack.clear
+    end
+
+    def in_class?
+      @stack.last == :class
+    end
+
+    def indirectly_in_def?
+      @stack.include?(:def) || @stack.include?(:defs)
+    end
+  end
+end
--- a/lib/parser/messages.rb
+++ b/lib/parser/messages.rb
@@ -58,6 +58,7 @@ module Parser
     :masgn_as_condition      => 'multiple assignment in conditional context',
     :block_given_to_yield    => 'block given to yield',
     :invalid_regexp          => '%{message}',
+    :invalid_return          => 'Invalid return in class/module body',
 
     # Parser warnings
     :useless_else            => 'else without rescue is useless',
--- a/test/test_runner_rewrite.rb
+++ b/test/test_runner_rewrite.rb
@@ -27,7 +27,7 @@ class TestRunnerRewrite < Minitest::Test
       end
 
       assert_equal expected_output.chomp, stdout.chomp
-      assert_equal expected_error.chomp, stderr.chomp unless RUBY_PLATFORM == 'java' # JRuby has warnings on Travis
+      assert_match expected_error.chomp, stderr.chomp
       assert_equal File.read(expected_file.expand_path), File.read(sample_file)
     end
   end
--- a/lib/parser.rb
+++ b/lib/parser.rb
@@ -69,6 +69,8 @@ module Parser
     require 'parser/builders/default'
   end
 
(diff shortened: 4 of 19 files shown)


lib/parser/messages.rb
```diff
@@ -58,6 +58,7 @@ module Parser
     :masgn_as_condition      => 'multiple assignment in conditional context',
     :block_given_to_yield    => 'block given to yield',
     :invalid_regexp          => '%{message}',
+    :invalid_return          => 'Invalid return in class/module body',
```
whitequark: `'return cannot be used in class or module body'`

lib/parser/context.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def in_class?
+      @stack.last == :class
+    end
+
+    def in_sclass?
+      @stack.last == :sclass
+    end
+
+    def in_def?
```
whitequark: We already have a `Parser::Base.in_def?`; I think you should replace that with the new context machinery, and update all the older parsers as well.

iliabylich: This flag is used to check that the parser is currently **somewhere** in `def` (to prevent class definition in the method body), not directly in `def`, so it's similar more to checking that the stack includes a `:def` symbol. Should I change the `Context` class to have methods like `in_def?` and `directly_in_def?` (and same stuff for other contexts)? Do you see any better names for these methods?

lib/parser/messages.rb
iliabylich: This message was copied from MRI, I thought parser should match them.

whitequark: Not really, we try to make our messages readable. The only thing that directly matches MRI is *when* they're emitted, not *what*. MRI does not have particularly good ones.

lib/parser/context.rb
whitequark: Take a look at how `parse*.y` handle `@def_level`, they save and restore it to achieve the same behavior as here. Unless I'm missing something. Can you show the difference with some code example, please?

lib/parser/messages.rb
iliabylich: I mean, I understand how `@def_level` works (and it looks similar to the `Context`), the difference is *how* parser uses it. `in_def?` checks for `@def_level > 0` (i.e. that parser is in the `def` context), so the difference comes in the following code:
``` ruby
def m
  class << something
    # right here, @def_level is 1, so in_def? returns true
    # but Context#in_def? returns false,
    # because current context is singleton class (in_sclass? returns true)
  end
end
```
So we need to differentiate these methods somehow:
1. a method that checks that the parser is *directly* in some context (like `directly_in_def?` that returns `@stack.last == :def`) 
2. a method that checks that the parser is *somewhere* in the context, but not necessary *directly* in this context (like `somewhere_in_def?` that returns `@stack.include?(:def)`)

Hope it makes sense 😄

whitequark: `in_def?` and `indirectly_in_def?` sound good to me, and the overall concept too. Thanks for being patient with me!

iliabylich:
@whitequark Fixed, could you check it please?

lib/parser/context.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    def in_block?
+      @stack.last == :block
+    end
+
+    def in_lambda?
+      @stack.last == :lambda
+    end
+
+    def indirectly_in_class?
```
whitequark: Do we really need all those? I'm not opposed but it seems extraneous.

test/test_runner_rewrite.rb
```diff
@@ -27,7 +27,7 @@ def assert_rewriter_output(path, args, input: 'input.rb', output: 'output.rb', e
       end
 
       assert_equal expected_output.chomp, stdout.chomp
-      assert_equal expected_error.chomp, stderr.chomp unless RUBY_PLATFORM == 'java' # JRuby has warnings on Travis
+      assert_match expected_error.chomp, stderr.chomp unless RUBY_PLATFORM == 'java' # JRuby has warnings on Travis
```
whitequark: If it's an `assert_match` now I think  we can remove the `unless RUBY_PLATFORM  == 'java'` part?

iliabylich: Yes, you are right. I've changed it because I have an old `2.4` branch, so I had a warning about non-compliant ruby version and the spec was failing.

lib/parser/context.rb
iliabylich: I've added these methods to make the code self-documented. I agree, it's better to keep only `in_class?/indirectly_in_def?` and add some comments about possibles states.

iliabylich:
@whitequark Fixed.

PR #422, opened 2017-12-25
Bump latest Ruby versions

stanhu:
See https://www.ruby-lang.org/en/downloads/releases/

### DIFF
--- a/lib/parser/current.rb
+++ b/lib/parser/current.rb
@@ -28,7 +28,7 @@ module Parser
     CurrentRuby = Ruby21
 
   when /^2\.2\./
-    current_version = '2.2.8'
+    current_version = '2.2.9'
     if RUBY_VERSION != current_version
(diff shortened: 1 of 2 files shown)


lib/parser/current.rb
```diff
@@ -46,7 +46,7 @@ def warn_syntax_deviation(feature, version)
     CurrentRuby = Ruby23
 
   when /^2\.4\./
-    current_version = '2.4.0'
+    current_version = '2.4.3'
```
whitequark: Nope, we still don't even support 2.4.1. If you remove this change I'll merge the rest.

stanhu: Interesting, thanks!

stanhu:
Thanks! Would you mind releasing a new version?

gitlab-winnie:
@whitequark Do you have a release planned including these changes?

whitequark:
When I have time for it, yes.

Issue #423, opened 2017-12-26
parser accepts :&&, but CRuby does not accept it.


pocke:
```bash
$ ruby -ce ':&&'
-e:1: syntax error, unexpected &&, expecting tSTRING_CONTENT or tSTRING_DBEG or tSTRING_DVAR or tSTRING_END

$ ruby-parse -e ':&&'
(sym :"&&")
```

CRuby does not accept `:&&`, but parser accepts.
`:||` has the same problem.

CRuby does not allow non overridable operator as a symbol literal without quotes.

https://github.com/ruby/ruby/blob/trunk/parse.y#L4340-L4343
https://github.com/ruby/ruby/blob/trunk/parse.y#L2087-L2090
https://github.com/ruby/ruby/blob/trunk/parse.y#L2136-L2165

----

I tried using a patch to solve this problem, but it does not work well. 

```diff
diff --git a/lib/parser/lexer.rl b/lib/parser/lexer.rl
index 32d7d07..a559410 100644
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -1761,7 +1761,7 @@ class Parser::Lexer
       };
 
       ':' ( bareword | global_var | class_var | instance_var |
-            operator_fname | operator_arithmetic | operator_rest )
+            operator_fname | operator_arithmetic - ('&&' | '||') | operator_rest )
       => {
         emit(:tSYMBOL, tok(@ts + 1), @ts)
         fnext expr_end; fbreak;
```

Parser with the patch rejects `:&&`, but it accepts `:&&1` as `:& & 1`. It is unexpected behaviour.

```bash
$ ruby -ce ':&&1'
-e:1: syntax error, unexpected &&, expecting tSTRING_CONTENT or tSTRING_DBEG or tSTRING_DVAR or tSTRING_END

# parser with the patch
$ ruby-parse -e ':&&1'
(send
  (sym :&) :&
  (int 1))

# parser without the patch
$ ruby-parse -e :&&1
warning: parser/current is loading parser/ruby24, which recognizes
warning: 2.4.x-compliant syntax, but you are running 2.6.0.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
(fragment:0):1:4: error: unexpected token tINTEGER
(fragment:0):1: :&&1
(fragment:0):1:    ^
```

whitequark:
Good catch!

whitequark:
Look carefully at what MRI is doing:
```
[...]

PR #424, opened 2017-12-26
Emit :&& as tSYMBEG + tANDOP, :|| as tSYMBEG + tOROP.

iliabylich:
Closes https://github.com/whitequark/parser/issues/423

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -1740,6 +1740,13 @@ class Parser::Lexer
       # SYMBOL LITERALS
       #
 
+      # :&&, :||
+      ':' ('&&' | '||') => {
+        fhold; fhold;
+        emit(:tSYMBEG, tok(@ts, @ts + 1), @ts, @ts + 1)
+        fgoto expr_fname;
+      };
+
       # :"bar", :'baz'
       ':' ['"] # '
       => {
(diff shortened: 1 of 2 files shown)


lib/parser/lexer.rl
```diff
@@ -1746,6 +1746,15 @@ class Parser::Lexer
       # SYMBOL LITERALS
       #
 
+      # :&&, :||
+      ':' % { emit(:tSYMBEG) }
+      ( '&&' % { op_tok = :tANDOP }
+      | '||' % { op_tok = :tOROP  }
+      )
+      => {
+        emit(op_tok, tok(@te - 2, @te), @te - 2, @te)
```
whitequark: We typically do it like the following elsewhere:
```
# :&&, :||
':' ("&&" | "||") => {
  fhold; fhold; emit(:tSYMEG, @ts, @ts + 1)
}
```

iliabylich: But then the lexer emits `tAMPER` (because there's no transition to the state that can handle `&&` as a `tANDOP`).  Adding `fgoto expr_end` (like in `1 && 2`) solves it, but is it a good idea?

whitequark: Now that I look at MRI, why not switch to EXPR_FNAME after `:` that's not `:'` or `:"` ?

iliabylich:
@whitequark Fixed.

PR #428, opened 2018-01-04
Enforcing frozen string literals with pragma comment.

pat:
A slightly different take on #354 - instead of relying on dependencies to be frozen-string-literal friendly, this PR adds the pragma comment to all Ruby files, and then string literals are frozen for Ruby versions that support it.

This ensures the test suite can remain green while supporting frozen string literals, separate of whether dependencies do. If/when dependencies are all supporting frozen string literals, then perhaps the env flag (as noted in the Travis CI config in #354) can be set as well.

### DIFF
--- a/test/test_source_comment_associator.rb
+++ b/test/test_source_comment_associator.rb
@@ -1,8 +1,4 @@
-#! /usr/bin/env ruby
-# coding: utf-8
 # frozen_string_literal: true
-# warn_indent: true
-# warn_past_scope: true
 
 require 'helper'
 require 'parser/ruby18'
@@ -33,6 +29,8 @@ class TestSourceCommentAssociator < Minitest::Test
 
   def test_associate
     ast, associations = associate(<<-END)
+#!/usr/bin/env ruby
+# coding: utf-8
 # class preceeding
 # another class preceeding
 class Foo # class keyword line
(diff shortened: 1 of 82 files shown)


test/test_source_comment_associator.rb
```diff
@@ -1,3 +1,9 @@
+#! /usr/bin/env ruby
+# coding: utf-8
+# frozen_string_literal: true
+# warn_indent: true
```
whitequark: `warn_ident`? `warn_past_scope`?

test/test_source_comment_associator.rb
```diff
@@ -27,8 +33,6 @@ def associate_locations(code)
 
   def test_associate
     ast, associations = associate(<<-END)
-#!/usr/bin/env ruby
```
whitequark: You shouldn't remove these. They're testing the behavior of the comment associator, and this code is not ever run.

test/test_source_comment_associator.rb
pat: Ah, I missed these unexpected changes - seems Pragmater got confused. I'll fix that up.

whitequark:
Thanks for the PR!

pat:
No worries - thanks for merging :)

Issue #430, opened 2018-01-09
ruby/ruby@0dfec65


iliabylich:
Upstream commit: https://github.com/ruby/ruby/commit/0dfec6596e9fd9ab81ffb5062028efbce0beed7f. The message of this commit is `insane syntax`:
``` ruby
while def foo; tap do end; end; break; end
```

iliabylich:
Also https://github.com/ruby/ruby/commit/5aa3523545d2bf053fb0b26ce84208ffeabe5c1e:
``` ruby
while class Foo; tap do end; end; break; end
```

PR #432, opened 2018-01-10
ruby25.y: Allow class and method definition in the while condition.

iliabylich:
Closes https://github.com/whitequark/parser/issues/430.
ruby/ruby@0dfec65 and ruby/ruby@5aa3523.
MRI doesn't have test cases for singleton method definition and singleton class handling in the while condition but in fact it handles it well (there are changes in corresponding sections in MRI's `parse.y`):
``` sh
$ dev-ruby -vce 'while def self.foo; tap do end; end; break; end;'
ruby 2.6.0dev (2018-01-10 trunk 61758) [x86_64-darwin17]
Syntax OK
$ dev-ruby -vce 'while class << self; tap do end; end; break; end'
ruby 2.6.0dev (2018-01-10 trunk 61758) [x86_64-darwin17]
Syntax OK
```

### DIFF
--- a/lib/parser/ruby25.y
+++ b/lib/parser/ruby25.y
@@ -1100,6 +1100,7 @@ rule
                     {
                       @static_env.extend_static
                       @lexer.push_cmdarg
+                      @lexer.push_cond
                       @context.push(:class)
                     }
                     bodystmt kEND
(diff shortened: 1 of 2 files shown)


whitequark:
Note to self: don't merge until Parser 2.5.0.

PR #442, opened 2018-01-25
+ Add option to emit JSON S-Expressions.

rattrayalex:
Hi there, 

I'm interested in being able to read a Ruby AST from JavaScript. I imagined that others might be interested in cross-language machine-readable access to the Ruby AST as well. 

I have yet to add docs & tests to this PR, but wanted to put it up now as a strawdog for feedback.

I believe that this PR title, once squashed, should pass your commit guidelines; please let me know if there's anything else expected. 

How does this look?

### DIFF
--- a/lib/parser/ast/node.rb
+++ b/lib/parser/ast/node.rb
@@ -34,7 +34,6 @@ module Parser
           @location = location
         end
       end
-
     end
 
   end
--- a/lib/parser/ast/node.rb
+++ b/lib/parser/ast/node.rb
@@ -34,21 +34,7 @@ module Parser
           @location = location
         end
       end
-      
-      ##
-      # Produces a machine-readable S-Expression.
-      #
-      def to_json
-        children_json = children.map do |child|
-          if child.is_a?(Node)
-            child.to_json
-          else
-            child
-          end
-        end
 
-        [type.to_s, *children_json]
-      end
     end
 
   end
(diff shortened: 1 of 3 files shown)


lib/parser/ast/node.rb
```diff
@@ -34,6 +34,21 @@ def assign_properties(properties)
...
+        children_json = children.map do |child|
+          if child.is_a?(Node)
+            child.to_json
+          else
+            child
+          end
+        end
+
+        [type.to_s, *children_json]
+      end
```
whitequark: This should go into the `ast` gem, I think.

rattrayalex: I was thinking the same. I'll submit a PR there. Thanks!

rattrayalex: (Do you have quick feedback on whether this is roughly the right code? I don't write a ton of ruby)

whitequark: Code looks OK.

rattrayalex: Thanks. Posted in https://github.com/whitequark/ast/pull/20, will remove this commit

rattrayalex:
Thanks for the quick feedback! Removed the to_json def. 

This now depends on https://github.com/whitequark/ast/pull/20

whitequark:
Add a dependency on ast 2.4.0. https://rubygems.org/gems/ast

rattrayalex:
Done

rattrayalex:
Thanks!

whitequark:
I still need to review and merge some other PRs before I can make a release.

rattrayalex:
Okay cool. Thanks!

PR #449, opened 2018-02-20
Fixed encoding comment handling.

iliabylich:
Ruby ❤️ 

``` sh
$ cat test.rb
# whatever -*-* !@##TYU23456 whatever encoding: koi8-r whatever -*- whatever
puts "".encoding

$ ruby test.rb
KOI8-R
```

I've checked other tests on MRI, the only difference is BOM meta-chars handling. MRI doesn't care about them. @whitequark Should we drop if from the parser as well?

### DIFF
--- a/lib/parser/source/buffer.rb
+++ b/lib/parser/source/buffer.rb
@@ -29,7 +29,7 @@ module Parser
       # @api private
       #
       ENCODING_RE =
-        /\A\#\s*(en)?coding\s*[:=]\s*
+        /[\s#](en)?coding\s*[:=]\s*
           (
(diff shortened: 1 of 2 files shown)


whitequark:
Drop what exactly?

iliabylich:
Parsing files as utf-8 if there are BOM chars (`EF BB BF`) on the first line.

iliabylich:
From what I see from quick googling Ruby < 2.0.0 (old times when UTF-8 wan't a default encoding) checks for BOM chars. For Ruby > 2.0.0 these chars don't make any sense. Let's keep it then!

iliabylich:
@whitequark Could you release it please?

Issue #435, opened 2018-01-15
Can't parse interpolated lambda with argument without parentheses 


iliabylich:
Without arguments everything is fine:
``` sh
$ bin/ruby-parse --25 -e '"#{->{}}"'
...
"#{->{}}"
      ^ tRCURLY "}"                             expr_endarg  [0 <= cond] [0 <= cmdarg]
"#{->{}}"
       ^ tSTRING_DEND "}"                       interp_string [0 <= cond] [0 <= cmdarg]
...
(dstr
  (begin
    (block
      (lambda)
      (args) nil)))
```

``` sh
$ bin/ruby-parse --25 -E -e '"#{-> arg {}}"'
...
"#{-> arg {}}"
      ^~~ tIDENTIFIER "arg"                     expr_arg     [0 <= cond] [0 <= cmdarg]
"#{-> arg {}}"
          ^ tLAMBEG "{"                         expr_value   [0 <= cond] [0 <= cmdarg]
"#{-> arg {}}"
           ^ tRCURLY "}"                        expr_endarg  [0 <= cond] [0 <= cmdarg]
"#{-> arg {}}"
            ^ tRCURLY "}"                       expr_endarg  [0 <= cond] [0 <= cmdarg]
(fragment:0):1:13: error: unexpected token tRCURLY
(fragment:0):1: "#{-> arg {}}"
(fragment:0):1:             ^
```

Wrapping the argument with parentheses solves the issue.

PR #450, opened 2018-02-20
Fix parsing of interpolated lambda with open args.

iliabylich:
The bug is caused by `p -=` and thus calling action attached to `e_lbrace` twice. This patch removes lookahead and emits the token instantly to guarantee that the action of `e_lbrace` gets called only once.

Basically, lambda-branch in the modified action is replaced to what corresponding action does later in
```
  expr_end := |*

    e_lbrace | 'do'
      => {
        if @lambda_stack.last == @paren_nest
          @lambda_stack.pop

          if tok == '{'.freeze
            emit(:tLAMBEG, '{'.freeze)
          else # 'do'
            emit(:kDO_LAMBDA, 'do'.freeze)
          end
        else
          if tok == '{'.freeze
            emit(:tLCURLY, '{'.freeze)
          else # 'do'
            emit_do
          end
        end

        fnext expr_value; fbreak;
      };
*|;
```

Closes https://github.com/whitequark/parser/issues/435

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -1437,12 +1437,12 @@ class Parser::Lexer
       w_space* e_lbrace
       => {
         if @lambda_stack.last == @paren_nest
-          p = @ts - 1
-          fgoto expr_end;
+          @lambda_stack.pop
+          emit(:tLAMBEG, '{'.freeze, @te - 1, @te)
         else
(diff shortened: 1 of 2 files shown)


whitequark:
Thanks, good fix! This needed a changelog entry but I've added one for you.

whitequark:
In recognition of your many high-quality contributions I've added you as a gem owner. Just in case.

PR #471, opened 2018-03-16
ruby26.y: branch parser.

iliabylich:
Probably should be merged after 2.6.0-stable release.


lib/parser/current.rb
```diff
@@ -65,10 +65,19 @@ def warn_syntax_deviation(feature, version)
     require 'parser/ruby25'
     CurrentRuby = Ruby25
 
+  when /^2\.6\./
+    current_version = '2.6.0'
```
whitequark: If you change this to `current_version = '2.6.0-dev'`...

lib/parser/current.rb
```diff
@@ -65,10 +65,19 @@ def warn_syntax_deviation(feature, version)
...
+
+    require 'parser/ruby26'
+    CurrentRuby = Ruby26
+
   else # :nocov:
     # Keep this in sync with released Ruby.
-    warn_syntax_deviation 'parser/ruby25', '2.5.x'
-    require 'parser/ruby25'
-    CurrentRuby = Ruby25
+    warn_syntax_deviation 'parser/ruby26', '2.6.x'
```
whitequark: ... and revert this change, this PR can be merged right now, and other work landed on top of it. This is what we generally do.

whitequark:
Also, branching a parser is not a user-visible feature, so it doesn't need a changelog entry.

iliabylich:
@whitequark Done. Rebased on the latest master, included latest fix into 2.6 parser, applied changes per your comments.

PR #476, opened 2018-03-21
ruby-parse displays escape sequences only in tty

### LINKED ISSUE #474: ruby-parse command should not display escape sequences when output is not a tty.
Problem
===

Currently ruby-parse displays escape sequences always.

It is useful, but the escape sequences are unnecessary when output is not a tty. For example, when I want to copy the output with `pbcopy`, `xsel` and so on, I didn't need coping escape sequences.

```
$ ruby-parse -Ee 'foo' | xsel --input --clipboard
```

The clipboard

```
[4mfoo[0m   
[1;31m^~~[0m [0;32mtIDENTIFIER "foo"[0m                           expr_cmdarg  [0 <= cond] [0 <= cmdarg] [0m
foo[4m[0m   
   [1;31m^[0m [0;32mfalse "$eof"[0m                               expr_cmdarg  [0 <= cond] [0 <= cmdarg] [0m
(send nil :foo)
```

Solution
===

Disable the escape sequences if output is not a tty.

I have a question.
We can implement this feature easily with [rainbow](https://github.com/sickill/rainbow) gem. But we should add the gem to dependencies to use rainbow.
Can we use the gem? Or should we keep the dependencies simple?

I'll work for it if the proposal is acceptable.

pocke:
Fix #474

### DIFF
--- a/lib/parser/lexer/explanation.rb
+++ b/lib/parser/lexer/explanation.rb
@@ -21,14 +21,14 @@ module Parser
       more = "(in-kwarg)" if @in_kwarg
 
       puts decorate(range,
-                    "\e[0;32m#{type} #{val.inspect}\e[0m",
-                    "#{state.to_s.ljust(12)} #{@cond} #{@cmdarg} #{more}\e[0m")
+                    Color.green("#{type} #{val.inspect}"),
+                    "#{state.to_s.ljust(12)} #{@cond} #{@cmdarg} #{more}")
 
       [ type, [val, range] ]
     end
 
     def state=(new_state)
-      puts "  \e[1;33m>>> STATE SET <<<\e[0m " +
+      puts "  #{Color.yellow(">>> STATE SET <<<", bold: true)} " +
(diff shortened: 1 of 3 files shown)


lib/parser/lexer/explanation.rb
```diff
@@ -40,12 +40,12 @@ def decorate(range, token, info)
...
       line = range.source_line + '   '
-      line[from...to] = "\e[4m#{line[from...to]}\e[0m"
+      line[from...to] = Color.underline(line[from...to])
 
       tail_len   = to - from - 1
       tail       = '~' * (tail_len >= 0 ? tail_len : 0)
-      decoration =  "#{" " * from}\e[1;31m^#{tail}\e[0m #{token} ".
-                        ljust(70) + info
+      decoration =  "#{" " * from}#{Color.red("^#{tail}", bold: true)} #{token} ".
+                        ljust(68) + info
```
pocke: L24 contained unnecessary `0;`, and this change removes the `0;`. So we need to reduce the ljust value to align them.

Issue #480, opened 2018-03-26
Error parsing code as Ruby 2.4 or 2.5


rrosenblum:
The following ruby code produces a syntax error when using the Ruby 2.4 or Ruby 2.5 parser. I believe that this is related to #473. I was not able to reproduce the issue using `v2.5.0.3`.

`test.rb`
```ruby
puts "#{foo} #{(bar - baz)}"
bar.each do |qux|
  puts "#{qux}"
end
```

```
bundle exec ruby-parse --24 test.rb
test.rb:2:10: error: unexpected token kDO_BLOCK
test.rb:2: bar.each do |qux|
test.rb:2:   
```

iliabylich:
@whitequark The error comes from the difference between the `parser` and MRI's grammar. On interpolation code lexer emits two tokens (and thus, calls surrounding logic for two tokens).

The first line of your code is not handled properly. Minimal case for reproducing: `m "#{}#{(foo)}"` (it's valid for parser, but `cmdarg` is `1` in the end of the line).

When lexer comes to the second interpolation it scans `#{`, emits tSTRING_BEG and tLPAREN:
```
m "#{}#{(foo)}"
     ^ tSTRING_DEND "}"                         interp_string  [0 <= cmdarg.1]
Shifting tSTRING_DEND }
pop_cmdarg
Lexer#emit tSTRING_DBEG #{
cmdarg.push(0)
Lexer#emit tLPAREN (
m "#{}#{(foo)}"
      ^~ tSTRING_DBEG "\#{"                     expr_beg      [10 <= cmdarg]
Shifting tSTRING_DBEG #{
push_cmdarg
m "#{}#{(foo)}"
        ^ tLPAREN "("                           expr_beg      [0 <= cmdarg.1]
Shifting tLPAREN (
```

Basically, instead of parentheses there can be array or hash brackets (`m "#{}#{[foo]}"` or `m "#{}#{{}}"`), because both instantly start a non-command context by doing `cmdarg.push(0)`.

MRI emits tokens one by one, that's the thing. That's why backported code doesn't work as it should.
Any ideas how to fix it?

iliabylich:
@whitequark Changing the following action:
```
  action extend_interp_code {
    current_literal = literal
    current_literal.flush_string
    current_literal.extend_content

    emit(:tSTRING_DBEG, '#{'.freeze)

    if current_literal.heredoc?
      current_literal.saved_herebody_s = @herebody_s
      @herebody_s = nil
    end

[...]

PR #482, opened 2018-03-28
Emit tSTRING_BEG and tSTRING_DBEG one by one to allow parser to properly manipulate cmdarg stack

iliabylich:
Also I've added missing `cmdarg_push` and `cond_push` for `{` in the string literal. MRI always does if, even if it's not important now (because of the `push_cmdarg`/`pop_cmdarg` around the string) - https://github.com/ruby/ruby/blob/trunk/parse.y#L8008-L8010

``` ruby
$ dev-ruby -ye '"#{}#{}#{}"' | grep "stack(pop)"
p->cond_stack(pop): 0 at line 8009
p->cmdarg_stack(pop): 0 at line 8010
p->cond_stack(pop): 0 at line 8009
p->cmdarg_stack(pop): 0 at line 8010
p->cond_stack(pop): 0 at line 8009
p->cmdarg_stack(pop): 0 at line 8010
```

Closes https://github.com/whitequark/parser/issues/480

@whitequark Does this bug mean that using `fcall` may be dangerous? Should we avoid it in the future?

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -379,27 +379,30 @@ class Parser::Lexer
   def push_literal(*args)
     new_literal = Literal.new(self, *args)
     @literal_stack.push(new_literal)
+    next_state_for_literal(new_literal)
+  end
 
-    if new_literal.words? && new_literal.backslash_delimited?
-      if new_literal.interpolate?
+  def next_state_for_literal(literal)
+    if literal.words? && literal.backslash_delimited?
+      if literal.interpolate?
         self.class.lex_en_interp_backslash_delimited_words
       else
         self.class.lex_en_plain_backslash_delimited_words
       end
-    elsif new_literal.words? && !new_literal.backslash_delimited?
-      if new_literal.interpolate?
+    elsif literal.words? && !literal.backslash_delimited?
+      if literal.interpolate?
         self.class.lex_en_interp_words
       else
         self.class.lex_en_plain_words
       end
-    elsif !new_literal.words? && new_literal.backslash_delimited?
-      if new_literal.interpolate?
+    elsif !literal.words? && literal.backslash_delimited?
+      if literal.interpolate?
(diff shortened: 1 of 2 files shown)


lib/parser/lexer.rl
```diff
@@ -1026,8 +1029,15 @@ class Parser::Lexer
...
 
+        @cond.lexpop
+        if @version < 24
+          @cmdarg.lexpop
+        else
+          @cmdarg.pop
+        end
+
         fhold;
-        fnext *stack_pop;
```
whitequark: Is the stack still used anywhere? If not then `stack_pop` should be removed.

iliabylich: Yes:
```
  expr_variable := |*
      global_var
      => {
        if    tok =~ /^\$([1-9][0-9]*)$/
          emit(:tNTH_REF, tok(@ts + 1).to_i)
        elsif tok =~ /^\$([&`'+])$/
          emit(:tBACK_REF)
        else
          emit(:tGVAR)
        end

        fnext *stack_pop; fbreak;
      };

      class_var_v
      => {
        if tok =~ /^@@[0-9]/
          diagnostic :error, :cvar_name, { :name => tok }
        end

        emit(:tCVAR)
        fnext *stack_pop; fbreak;
      };

      instance_var_v
      => {
        if tok =~ /^@[0-9]/
          diagnostic :error, :ivar_name, { :name => tok }
        end

        emit(:tIVAR)
        fnext *stack_pop; fbreak;
      };
  *|;
```

because there are still a couple of `fcall`-sin the code. From what I see none of them emit tokens that change `cmdarg` in the grammar.

iliabylich:
@whitequark Could you take a look please? I have a fix for https://github.com/whitequark/parser/issues/483, but it depends a bit on this PR.

whitequark:
@iliabylich by the way, do you think you can add changelog entries (by formatting the commits in the corresponding way)? I've fixed this PR manually for now.

Issue #479, opened 2018-03-25
[Bug #14621]


whitequark:
See [Ruby bug](https://bugs.ruby-lang.org/issues/14621) and commits https://github.com/ruby/ruby/commit/491f523ab18cae6e14a2929de14d51803c362a7c, https://github.com/ruby/ruby/commit/58fbe69a5ba2c6c896745dab5e41b3bbd1950015.

We have a slightly different bug than Ruby itself
```
$ ./bin/ruby-parse t.rb
(send nil :puts
  (send
    (str "1 \n  2\n") :inspect))
$ ruby t.rb
"1   2\n"
```

iliabylich:
@whitequark Should we backport if only for 26?

whitequark:
I don't understand the question. Our behavior is wrong on 2.3-2.6; MRI's behavior is wrong on 2.3-2.6 and will be fixed there.

iliabylich:
My bad, I've forgot that squiggly heredoc is supported only for >= 23

iliabylich:
@whitequark Need some help. Parser handles squiggly heredocs in the `Parser::Lexer::Dedenter`. The lexer can't emit
```
<<~HERE
  a \
  b
HERE
```
instantly as `"a b\n"` because it needs to keep the information about spacing (next lines may have a different indentation). Does it mean that the lexer should emit it as two lines and merge them later in the `Dedenter`? It makes the `Dedenter` much more complex, but from what I see that's the right way to handle it. Could you confirm it? Do you have any better ideas?

PR #487, opened 2018-04-02
Parse slash + newline in heredocs as a line continuation.

iliabylich:
Upstream commits: ruby/ruby@491f523 and ruby/ruby@58fbe69.
Closes https://github.com/whitequark/parser/issues/479

The lexer must emit `"\\\n"` without escaping to let `Dedenter` know that there's a newline (and so that the code after must be dedented).


iliabylich:
@whitequark Rebased and (if the code is fine) ready for merging. Could you take a look please? tbh I don't like this solution but I don't see any other way of backporting it.

iliabylich:
@whitequark Did you have a chance to take a look at this PR? Is there anything you'd like me to change or clarify?

iliabylich:
It could be written much, much better without buffer and mutation. But that would also affect performance.

whitequark:
Yeah, I get it.

PR #493, opened 2018-04-04
call String#getbyte

cremno:
I forgot about pushing this. It's related to my previous PR #457 but even though `value.b[0].ord` already would've been an improvement, `value.getbyte(0)` is even better.


whitequark:
Thanks!

Issue #492, opened 2018-04-02
parse error: cond ? a + '"': ''


iliabylich:
``` ruby
$ bin/ruby-parse --26 -E -e "a ? b + '': nil"
a ? b + '': nil
^ tIDENTIFIER "a"                               expr_cmdarg  [0 <= cond] [0 <= cmdarg]
a ? b + '': nil
  ^ tEH "?"                                     expr_value   [0 <= cond] [0 <= cmdarg]
a ? b + '': nil
    ^ tIDENTIFIER "b"                           expr_cmdarg  [0 <= cond] [0 <= cmdarg]
a ? b + '': nil
      ^ tPLUS "+"                               expr_beg     [0 <= cond] [0 <= cmdarg]
a ? b + '': nil
        ^ tSTRING_BEG "'"                       expr_labelarg [0 <= cond] [0 <= cmdarg]
a ? b + '': nil
         ^~ tLABEL_END "'"                      expr_labelarg [0 <= cond] [0 <= cmdarg]
(fragment:0):1:10: error: unexpected token tLABEL_END
(fragment:0):1: a ? b + '': nil
(fragment:0):1:          ^~
```

But `a ? '': nil` can be parsed.

iliabylich:
@whitequark In MRI `EXPR_VALUE` is an alias for `EXPR_BEG`:
``` c
enum lex_state_e {
    DEF_EXPR(BEG),
    DEF_EXPR(END),
    DEF_EXPR(ENDARG),
    DEF_EXPR(ENDFN),
    DEF_EXPR(ARG),
    DEF_EXPR(CMDARG),
    DEF_EXPR(MID),
    DEF_EXPR(FNAME),
    DEF_EXPR(DOT),
    DEF_EXPR(CLASS),
    DEF_EXPR(LABEL),
    DEF_EXPR(LABELED),
    DEF_EXPR(FITEM),
    EXPR_VALUE = EXPR_BEG,
    // ...
}
```
 and code constructions like `{ 'a': 'b' }` are handled using a dual state `EXPR_BEG|EXPR_LABEL` after tLBRACE (so if next token is `'` the lexer creates a literal that can become a label). But in the Parser `EXPR_VALUE` is a special state that allows everything that is supported by `EXPR_BEG` except labels and quoted labels. Is it intentional? That's a serious difference in implementation.

[...]

PR #498, opened 2018-04-11
lexer.rl: Fix parsing of `a ? b + '': nil`.

iliabylich:
Closes https://github.com/whitequark/parser/issues/492

All operators in `operator_arithmetic` and `operator_rest` (except `|`, `~`, `!`) can be followed by quoted labels if this symbol is actually a string + tCOLON from the ternary operator.


whitequark:
This is pretty elegant.

PR #500, opened 2018-04-16
Fix parsing of `let (:a) { m do; end }`.

iliabylich:
In this case lexer emits multiple tokens (`tLBRACE_ARG` + one more token because there's no `fbreak`) and `cmdarg.push` gets called for both tokens. Later parser does `pop` after `command_args` and as a result it pops a wrong value. 

Also I've added `empty?` method to the `StackState` class to use it in tests and rely on the interface.

Found it during stress testing (~40k files from rubygems)


whitequark:
Thanks! You are doing impressive work and I appreciate it a lot.

Issue #502, opened 2018-04-17
ruby/ruby@f168dbd


iliabylich:
https://github.com/ruby/ruby/commit/f168dbd9ef3d27ec629d323f8a0834d63b5d7baa

iliabylich:
Depends on https://github.com/whitequark/parser/pull/500

Issue #501, opened 2018-04-16
do-end inside {} code block


jmarianer:
parser version 2.5 and up fail on a do-end class definition inside a {} code block. The parse succeeds if the braces are replaced with do-end or vice versa (i.e., the blocks are made consistent). This also succeeds when using parser-2.4.0.2.

```
[rmarianer@iMac ~]$ ruby -e "def let(i) 4 end; p let (:aaa) { Class.new do end }"
4
[rmarianer@iMac ~]$ ruby-parse -e "def let(i) 4 end; p let (:aaa) { Class.new do end }"
(fragment:0):1:44: error: unexpected token kDO_BLOCK
(fragment:0):1: def let(i) 4 end; p let (:aaa) { Class.new do end }
(fragment:0):1:                                            ^~      
```

mikegee:
I think https://github.com/whitequark/parser/pull/500 will resolve this issue.

iliabylich:
No, it will not. Ruby trunk also fails parsing this code, it will be fixed after merging (hopefully) https://bugs.ruby-lang.org/issues/14690 and backporting it to the Parser.

mikegee:
👍 Thanks for the clarification.

jmarianer:
Thanks, everyone. I'll just sit tight, then.

PR #503, opened 2018-04-17
Fix cmdargs after command_args followed by tLBRACE_ARG.

iliabylich:
Closes https://github.com/whitequark/parser/issues/502
Fixes https://github.com/whitequark/parser/issues/501.

This commit tracks upstream commit ruby/ruby@f168dbd.

The spec says `SINCE_2_0` because 1.8 doesn't emit empty parens (so there's no `s(:begin)` in the ast) and 1.9 fails parsing it (no ideas why).

There's one test in the suite that verifies it - https://github.com/whitequark/parser/blob/master/test/test_parser.rb#L4128-L4164 - and it expect exactly what I described above. @whitequark do you remember why this syntax is not supported by 1.9?

### DIFF
--- a/lib/parser/ruby24.y
+++ b/lib/parser/ruby24.y
@@ -875,7 +875,20 @@ rule
                     }
                   call_args
                     {
-                      @lexer.cmdarg.pop
+                      # call_args can be followed by tLBRACE_ARG (that does cmdarg.push(0) in the lexer)
+                      # but the push must be done after cmdarg.pop() in the parser.
+                      # So this code does cmdarg.pop() to pop 0 pushed by tLBRACE_ARG,
+                      # cmdarg.pop() to pop 1 pushed by command_args,
+                      # and cmdarg.push(0) to restore back the flag set by tLBRACE_ARG.
+                      last_token = @last_token[0]
+                      lookahead = last_token == :tLBRACE_ARG
+                      if lookahead
+                        top = @lexer.cmdarg.pop
+                        @lexer.cmdarg.pop
+                        @lexer.cmdarg.push(top)
+                      else
+                        @lexer.cmdarg.pop
+                      end
 
                       result = val[1]
                     }
--- a/lib/parser/ruby25.y
+++ b/lib/parser/ruby25.y
@@ -885,7 +885,20 @@ rule
                     }
                   call_args
                     {
-                      @lexer.cmdarg.pop
+                      # call_args can be followed by tLBRACE_ARG (that does cmdarg.push(0) in the lexer)
+                      # but the push must be done after cmdarg.pop() in the parser.
+                      # So this code does cmdarg.pop() to pop 0 pushed by tLBRACE_ARG,
+                      # cmdarg.pop() to pop 1 pushed by command_args,
+                      # and cmdarg.push(0) to restore back the flag set by tLBRACE_ARG.
+                      last_token = @last_token[0]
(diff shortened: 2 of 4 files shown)


whitequark:
Is this really going to be backported all the way back? The Ruby bug doesn't indicate so.

whitequark:
> do you remember why this syntax is not supported by 1.9?

Yes. Well, sorta. There is no inherent reason it's not supported, no one just came around to backporting it from 2.0 back when this was fixed. I think if you trawl git history you might find the Ruby bug introducing this syntax.

iliabylich:
> Is this really going to be backported all the way back? The Ruby bug doesn't indicate so.

good question. I'm completely confused. `parse.y` in the `v2_5_1` branch still does a few lexpops (while I thought that it was completely removed and backported to 24/25). and also it doesn't include a patch for doing only `cmdarg.push/pop` instead of save/restore mechanism. Looks like I misinterpreted a couple of `backport` tags on the bugs.ruby-lang.

Should I revert recent changes in `ruby24/ruby25.y` to match corresponding GH branches? I'm quite sure that the code in MRI's `trunk` is semantically equivalent to `v2_5_1` branch (in terms of cmdarg handling, because there were no changes to blocks semantics). But frankly speaking I was almost as confident 20 PRs before, so it's up to you. I can do it.

iliabylich:
Sure

iliabylich:
@nobu doesn't reply. So, should I revert this PR and https://github.com/whitequark/parser/pull/485 for 24 and 25 parsers? I believe that the code does the same thing but in a better manner. And 24/25/26 don't have any difference in the blocks handling.

whitequark:
@iliabylich If it's not backported then please revert. We don't do *everything* to be bug-compatible with MRI, it's more of a best-effort, but here the resolution is clear.

iliabylich:
@whitequark https://github.com/ruby/ruby/commit/1727c83 and https://github.com/ruby/ruby/commit/9c5d90b were not backported to 24/25. We have a problem if both of them need reverting in `ruby24`/`ruby25`.

Without https://github.com/ruby/ruby/commit/1727c83 it's impossible to backport these lines - https://github.com/ruby/ruby/blob/v2_4_4/parse.y#L7976-L7977 (actually https://github.com/ruby/ruby/commit/1727c83 drops them). MRI parser goes into `EXPR_END|EXPR_ENDARG` state to make this dirty check. I had (well, accidentally) to backport it to parsers 24/25 because otherwise it would be impossible to backport another change. So the code in MRI/v2_4_4 branch can't be backported. It does things that are not supported by ragel.

> We don't do everything to be bug-compatible with MRI, it's more of a best-effort, but here the resolution is clear

This bug was introduced in MRI 26, previous versions are not affected. It was originally introduced in https://github.com/ruby/ruby/commit/35b2038ee40ae3ccff140a504f780e965acfc19f#diff-6d840db07406437770b610102b284273 and https://github.com/ruby/ruby/commit/9c5d90b516e665294fa673ca6d9819d7406a7629#diff-6d840db07406437770b610102b284273 revealed it. This PR touches 24/25 because `@cmdarg` handling for `ruby24`/`ruby25` follows `ruby26`. So Parser has this bug for 24/25/26 because 24/25 do what 26 does.

whitequark:
Thanks for explanation, I misunderstood. Then this is fine as it is, I think.

whitequark:
Thank you for implementing this fix!

PR #508, opened 2018-04-25
Specify release version

y-yagi:


whitequark:
The changelog is generated automatically using `rake changelog`. I've updated it.

Issue #507, opened 2018-04-25
error: unexpected token kDO 


y-yagi:
It is a bug similar to #466 and #447, but it is reproduced with v2.5.1.0.  

```ruby 
# test.rb
check_open = -> *args do
  # ...
end

check_open.call
```

```
$ ruby-parse -v
ruby-parse based on parser version 2.5.1.0 

$ ruby-parse test.rb
test.rb:1:23: error: unexpected token kDO
test.rb:1: check_open = -> *args do
test.rb:1:                       ^~ 

$ ruby-parse _2.4.0.2_ test.rb
warning: parser/current is loading parser/ruby25, which recognizes
warning: 2.5.0-compliant syntax, but you are running 2.5.1.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
(begin
  (lvasgn :check_open
    (block
      (lambda)
      (args
        (restarg :args)) nil))
  (send
    (lvar :check_open) :call)) 

$ ruby -vc test.rb
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
Syntax OK 
```

PR #509, opened 2018-04-25
Fix parsing of 'm = -> *args do end'.

iliabylich:
That's a regression introduced in f0a5c6b5a55ccadc784489d966efc8975b3f4e32.

`fgoto` doesn't exit from `advance`, so `@command_state` remains the same and later `arg_or_cmdarg` can properly select the next state. Sorry, that was not obvious in https://github.com/whitequark/parser/pull/498.

@whitequark A quick question. `Parser#last_token` in some cases requires lexer to emit tokens one by one, but in this case we have to use `fgoto` to continue parsing without changing `@command_state`.  I feel like it can become a problem in the future. There are 54 occurrences of `fgoto` in the lexer, I know, some of them are safe because they do `fhold; fgoto ...`. But this specific action emits a token **and** does `fgoto`. Is there a way to rewrite `@command_state` somehow to not depend on `Lexer#advance` (and thus do `fnext; fbreak` when we need)?

Closes https://github.com/whitequark/parser/issues/507#event-1593145674

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -2249,7 +2249,7 @@ class Parser::Lexer
       ( operator_arithmetic | operator_rest ) - ( '|' | '~' | '!' )
       => {
         emit_table(PUNCTUATION);
-        fnext expr_value; fbreak;
+        fgoto expr_value;
       };
 
       ( e_lparen | '|' | '~' | '!' )
--- a/test/test_parser.rb
+++ b/test/test_parser.rb
@@ -7013,4 +7013,16 @@ class TestParser < Minitest::Test
       %q{},
       SINCE_2_0)
   end
+
+  def test_parser_bug_507
+    assert_parses(
+      s(:lvasgn, :m,
+        s(:block,
+          s(:lambda),
+          s(:args,
+            s(:restarg, :args)), nil)),
+      %q{m = -> *args do end},
+      %q{},
+      SINCE_1_9)
+  end
 end


whitequark:
> Is there a way to rewrite @command_state somehow to not depend on Lexer#advance (and thus do fnext; fbreak when we need)?

I don't understand the question.

iliabylich:
> I don't understand the question.

Sorry, I'm not a native English speaker. Let me start again from the beginning.

`Lexer#advance` sets `@command_state` by looking at `@cs`. If the state is `expr_value` or `line_begin` the value becomes `true`, otherwise it's `false`.

The bug that this PR aims to fix comes from `fnext; fbreak` that exits from the `#advance` method. `fgoto` doesn't exit from the lexing loop, it simply jumps back to the beginning of the huge `case`, so it's a workaround to preserve the value of the `@command_state` variable and properly handle the next token.

But emitting multiple tokens from a single `#advance` call is potentially dangerous because it's a lookahead on the lexer level that may work incorrectly with a lookahead in the parser. The parser expects lexer to emit tokens one by one. I've sent a few PRs in the past to fix such issues. Ideally we shouldn't have any constructions like `emit; fgoto` in the lexer.

So the question is: is there any way to fix without doing dangerous things? What this PR should actually do is something like
``` ruby
emit_table(PUNCTUATION)
# do something to force @command_state
# to be false for the next #advance call
# and ignore that the state will be expr_value
fnext expr_value; fbreak;
```

Hope it makes more sense.

whitequark:
Thanks, I understand now.

I can't think of any workaround that you wouldn't already know, i.e. apart from just adding another variable, or a state that sets `@command_state` and `fgoto`'s somewhere else.

PR #517, opened 2018-07-04
Fix return type annotation for Parser::Base#parse

gmalette:
`Parser::Base#parse` sometimes returns `false` but it's return type annotations doesn't show that

Example

```ruby
require 'parser/current'

parser = Parser::CurrentRuby.new
buf = Parser::Source::Buffer.new("(string)")
buf.source = "foo *"
exp = parser.parse(buf)
#=> false
```

### DIFF
--- a/lib/parser/base.rb
+++ b/lib/parser/base.rb
@@ -158,7 +158,7 @@ module Parser
     # Parses a source buffer and returns the AST.
     #
     # @param [Parser::Source::Buffer] source_buffer The source buffer to parse.
-    # @return [Parser::AST::Node]
+    # @return [Parser::AST::Node, false]
     #
     def parse(source_buffer)
       @lexer.source_buffer = source_buffer


whitequark:
This is actually a bug, and is unnoticed because `default_parser` [makes all errors fatal](https://github.com/whitequark/parser/pull/517/files#diff-4a07315825736d48d2522e02ae2e5831R84). The default diagnostic handler should probably raise an exception on an error even if errors are not set to fatal. Not sure. @iliabylich ?

iliabylich:
@whitequark Are you going to introduce a breaking change? 😄 Actually I'm fine with the default behavior since it's configurable.

whitequark:
> Are you going to introduce a breaking change?

Breaking changes are fine when the behavior we're breaking 1) has never been documented or guaranteed 2) is buggy. It just seems like a footgun--e.g. someone could do `ast = parser.parse; return if not ast` expecting that `ast` would be `nil` if there's no actual code, but it's actually `false` because of a parse error.

whitequark:
Please do.

Issue #518, opened 2018-07-10
Something's broken in 2.5.1.1


mvz:
Follow-up from https://github.com/rubocop-hq/rubocop/issues/6092#issuecomment-403715689: The following Ruby code is not parsed with parser 2.5.1.1 anymore:

```ruby
class Foo < Bar
end
```

Result:

```
$ ruby-parse --25 foo.rb
foo.rb:2:1: error: unexpected token kEND
foo.rb:2: end
foo.rb:2: ^~~
```

The 1.9 parser works:

```
$ ruby-parse --19 foo.rb
(class
  (const nil :Foo)
  (const nil :Bar) nil)
```

whitequark:
Parser 2.5.1.2 with the fix released.

mvz:
Thanks for the quick response @whitequark and @iliabylich!

PR #520, opened 2018-07-10
Partially revert 5ba072d and properly handle 'm = -> *args do end'.

iliabylich:
The bug was introduced in https://github.com/whitequark/parser/commit/5ba072dba6d0f1270eab9cc782dad2fa5e759abe, but we didn't know about it. Sad, but our test suite doesn't cover all cases.

@whitequark I'm getting more and more unconfident in every patch that I send. Is there any way to improve our CI to handle trivial cases like this one? Any ideas?

Closes https://github.com/whitequark/parser/issues/518

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -2243,13 +2243,19 @@ class Parser::Lexer
       # OPERATORS
       #
 
+      '*'
+      => {
+        emit(:tSTAR2)
+        fgoto expr_value;
+      };
+
       # When '|', '~', '!', '=>' are used as operators
       # they do not accept any symbols (or quoted labels) after.
       # Other binary operators accept it.
-      ( operator_arithmetic | operator_rest ) - ( '|' | '~' | '!' )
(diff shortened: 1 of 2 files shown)


bbatsov:
@iliabylich RuboCop's codebase is huge and it's by far the largest project using `parser`, so running it's test suit is a very good way to see if you didn't break anything. :-)

whitequark:
> @whitequark I'm getting more and more unconfident in every patch that I send. Is there any way to improve our CI to handle trivial cases like this one? Any ideas?

I understand why you feel this way, but so far I think you're doing a great job maintaining Parser. Certainly not much worse than me, in any case, and that's all the more impressive given that you didn't write it. Please don't let this discourage you.

I agree with @bbatsov that running the Rubocop test suite on CI is a good idea. Probably as a PR gate, but definitely as a release gate.

bbatsov:
> I understand why you feel this way, but so far I think you're doing a great job maintaining Parser. Certainly not much worse than me, in any case, and that's all the more impressive given that you didn't write it. Please don't let this discourage you.

I totally agree! Thanks for all your work on `parser`, @iliabylich! 🙇

PR #521, opened 2018-07-13
Setup diagnostic to throw a Parser::SyntaxError on any error.

iliabylich:
Also I think #517 should be merged because `parse` *technically* still can return `false` if parser is configured with `diagnostics.all_errors_are_fatal = false`. In such case `Diagnostic::Engine` doesn't throw any errors.

Also I'm not sure why there are 3 kinds of diagnostic messages, I think `warning` and `error` should be enough. @whitequark Could you explain it please? I just can't imagine a case when someone may want parser to ignore errors.

### DIFF
--- a/lib/parser/base.rb
+++ b/lib/parser/base.rb
@@ -120,6 +120,7 @@ module Parser
     #
     def initialize(builder=Parser::Builders::Default.new)
       @diagnostics = Diagnostic::Engine.new
+      @diagnostics.all_errors_are_fatal = true
 
       @static_env  = StaticEnvironment.new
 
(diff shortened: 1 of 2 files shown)


whitequark:
> Could you explain it please? I just can't imagine a case when someone may want parser to ignore errors.

Racc provides a way to do error recovery. This allows you to e.g. recover from a broken semicolon by looking at where the method ends, and proceeding from there. It's generally a good idea for tooling because if a file contains multiple errors, it lets you see more than one at a time. Take a look at how clang handles compiler errors, for example.

PR #540, opened 2018-12-31
Update CHANGELOG.md

spheric:


spheric:
running the rake task generates the same change minus a few --- and commit message

spheric:
you may have run the rake task before you change the version?

PR #535, opened 2018-12-08
Relax bundler dev dependency to allow bundler v2.0.0.pre2

walf443:
ruby 2.6.0-rc1 was released.
https://www.ruby-lang.org/en/news/2018/12/06/ruby-2-6-0-rc1-released/

it seems to bundles bundler v2.0.0.pre2 https://github.com/bundler/bundler/releases/tag/v2.0.0.pre.2

I'd like to test rubocop with v2.6.0-rc1. But without this changes, it got following errors.

```
$ bunlde install
NOTE: Gem::Specification#has_rdoc= is deprecated with no replacement. It will be removed on or after 2018-12-01.
Gem::Specification#has_rdoc= called from /Users/yoshimin/src/github.com/whitequark/parser/parser.gemspec:15.
Fetching gem metadata from https://rubygems.org/.............
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    bundler (~> 1.15)

  Current Bundler version:
    bundler (2.0.0)
This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?

Could not find gem 'bundler (~> 1.15)' in any of the relevant sources:
  the local ruby installation
```

PR #543, opened 2019-01-08
Fix example in TreeRewriter documentation

mvz:
The `#rewrite` method takes at least two arguments: A buffer and an AST.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,6 @@ matrix:
     - rvm: rbx-2
     - script: ./ci/run_rubocop_specs
 before_install:
-  - gem install bundler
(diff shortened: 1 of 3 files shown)


mvz:
I pushed a fix that should make the build pass but I can also open another pull request for that if you prefer.

mvz:
Hm ... looks like the fix from #535 will be needed as well.

mvz:
I cherry-picked the commit from #535 and now finally Travis is green.

iliabylich:
@whitequark Can we merge it?

iliabylich:
Thanks!

PR #552, opened 2019-02-18
Remove unreachable code from lexer.rl

pocke:
It is a meaningless condition. Because it checks the version is less than 2.4, but it has been checked the version is 1.8 or 1.9.


whitequark:
@iliabylich You've made this change in https://github.com/whitequark/parser/commit/ebcd32c90c8caede85178963005d624ce22bba14. Can you please look at it?

Issue #558, opened 2019-03-16
ruby/ruby@7006fde


iliabylich:
https://github.com/ruby/ruby/commit/7006fdecf2ff5e832cd6da1bf02462146a9190b8

iliabylich:
Fixed in https://github.com/whitequark/parser/pull/560

PR #560, opened 2019-03-18
+ ruby27.y: dsym should be treated as string.

iliabylich:
This commit tracks upstream commit ruby/ruby@7006fde.

`xstring_contents` and `string_contents` rules are equivalent, so it doesn't change anything for parser.


iliabylich:
Closes https://github.com/whitequark/parser/issues/558

Issue #559, opened 2019-03-17
ruby/ruby@12acc75


iliabylich:
https://github.com/ruby/ruby/commit/12acc751e3e7fd6f8aec33abf661724ad76c862a

iliabylich:
Also https://github.com/ruby/ruby/commit/d0e25ed277549af21242009e649d93b73c82576c

iliabylich:
Also https://github.com/ruby/ruby/commit/964bbc1686c7dd42d58da504a6968c90d5a428fd

iliabylich:
@whitequark MRI tracks the max value of `@<digits>` and saves it into `args` node metadata:
```
static NODE *
args_with_numbered(struct parser_params *p, NODE *args, int max_numparam)
{
    if (max_numparam > 0) {
	if (!args) args = new_args_tail(p, 0, 0, 0, 0);
	args->nd_ainfo->pre_args_num = max_numparam;
```

And later it's used in the arity checking:
```
$ dev-ruby -ve 'p -> { @1 + @2 }.arity'
ruby 2.7.0dev (2019-03-19 trunk 67299) [x86_64-darwin18]
2
```

I'm pretty sure we need to do something similar. Otherwise there's no (simple) way to get the number of arguments. Any thoughts?

iliabylich:
I know that you'd like to not make any breaking changes, so maybe let's introduce a new node `numbered_args` that has a single child: the amount of numbered parameters?

whitequark:
OK.

iliabylich:
Also https://github.com/ruby/ruby/commit/ae07b66aaa092c59ac9d544c9b582712290dc357 + https://github.com/ruby/ruby/commit/6ca9e7cc0785c33f6d382176dbd79d6c91db72fe

iliabylich:
I think it's buggy:
```
->(a = ->{@1}) {}
SyntaxError ((irb):6: ordinary parameter is defined)
..._; rescue _.class; ->(a = ->{@1}) {}
...                             ^~
```
... because MRI's implementation is not stack-based.

Same error for `def m(a = ->{@1}); end`.

My PR handles it properly:
```
$ bin/ruby-parse --27 -e 'def m(a = ->{@1}); end'
(def :m
  (args
    (optarg :a
      (numblock
        (lambda) 1
        (numparam 1)))) nil)

$ bin/ruby-parse --27 -e '->(a = ->{@1}) {}'
(block
  (lambda)
  (args
    (optarg :a
[...]

PR #565, opened 2019-03-29
Add numbered parameters support.

iliabylich:
Closes https://github.com/whitequark/parser/issues/559

I tried to make this change as similar to MRI's one as possible.

Changes:
1. Moved parsing `Context` to lexer. Parser delegates it to lexer, so this diff contains a lot of `@` deletions.
2. Added `MaxNumparamStack` that tracks max numparam value in the current scope. MRI uses ugly values and comparison to determine ability to parse numbered params. I've changed it to make the code a bit more readable. Not sure if it's a good idea, but methods like `cant_have_numparams!` sound way better to me than `value = -1`


whitequark:
Can you please split the change into two commits? As it is, it's hard to review, and I imagine bisecting would be hell, too.

bbatsov:
I still have hope they'd reconsider this https://bugs.ruby-lang.org/issues/15723 Not sure if it won't be better to wait a bit, as I see there's a reasonable amount of people who dislike this new syntax.

iliabylich:
@whitequark PTAL

doc/AST_FORMAT.md
```diff
@@ -1067,6 +1067,31 @@ However, the following code results in a parse error:
 def f(*a: b); end
 ~~~
 
+## Numbered arguments
+
+Ruby 2.7 introduced a feature called "numbered parameters". This AST node
+doesn't have any sourcemaps and its purpose is to indicate the amount of
```
whitequark: That's troublesome, as all of our exisiting nodes have a source map with at least `expr`...

whitequark:
This looks generally okay, but I'd need to find some time to investigate it closely.

doc/AST_FORMAT.md
iliabylich: Is there any other way to attach metadata to the node? Can we add a `metadata` attribute to the `AST::Node`? I guess that's a breaking change, libraries that use `parser` may define this field on their own, they'll have to introduce some changes to support it.

Another option could be introducing a new method `parse_with_metadata` that could return `[ast, node_to_its_metadata_map]`. Sounds ugly, that's just an idea.

It's all about attaching arbitrary metadata to a single node (I agree that this metadata can't be represented by an artificial node)

whitequark: I think this should be a different node. Perhaps instead of `block` there should be a `numblock` node that would have the number of parameters instead of a `args` child node.

iliabylich: Like this?
```
s(:numblock,
  send_or_lambda_node,
  max_numparam,
  body_node)
```

lib/parser/builders/default.rb
```diff
@@ -437,6 +437,11 @@ def cvar(token)
         variable_map(token))
     end
 
+    def numparam(token)
+      n(:numparam, [ value(token)[1..-1].to_i ],
```
mbj: I think you can use `value(token).drop(1)` here that should be equivalent but does not create the range object.

iliabylich: I'd prefer to not use any destructive methods on values that I don't own unless there's a good reason for that. This string has at most 4 chars

doc/AST_FORMAT.md
whitequark: Yes.

lib/parser/builders/default.rb
mbj: Its not destructive (and misnamed):

```
irb(main):001:0> a = [1,2,3]
=> [1, 2, 3]
irb(main):002:0> a.drop(1)
=> [2, 3]
irb(main):003:0> a
=> [1, 2, 3]
```

whitequark: That should be done by the lexer anyway.

mbj: OMG forget the noise, this is a string, not an array.

lib/parser/ruby27.y
```diff
@@ -319,9 +319,10 @@ rule
                       method_call = @builder.call_method(nil, nil, val[0],
                                         nil, val[1], nil)
 
-                      begin_t, args, body, end_t = val[2]
+                      begin_t, args, max_numparam, body, end_t = val[2]
```
iliabylich: @whitequark Any thoughts on this? We flush `@lexer.max_numparam` when the body gets reduced and so in the `block` rule there's no way to get it. That's why I had to pass it down via `result`. Do you see any other way to store this value?

doc/AST_FORMAT.md
iliabylich: Done, PTAL

lib/parser/ruby27.y
iliabylich: Maybe let's create a "dummy" node `numbered_args` so `args` here is either `args` or `numbered_args`. Later builder can use it to emit `block` or `numblock`. This is basically similar to `ident` node that never appears in the final output.

whitequark: Sorry for the delay, this fell between the cracks. I agree with `numbered_args` plan. (Only it should be called `numargs` to keep our convention.)

iliabylich: Done, PTAL when you have time

iliabylich:
Included commits from https://github.com/whitequark/parser/issues/559#issuecomment-485910843, however we are still blocked by https://bugs.ruby-lang.org/issues/15789

iliabylich:
ok, all changes are backported, @whitequark PTAL when you have time. One thing that worries me a bit is that arrow lambda args are getting parsed in the context of the lambda itself:
```
->(a, b, c) { body }
   ^^^^^^^ lambda
              ^^^^ lambda
```
Which I think is correct. MRI handles default values of `optarg` and `kwoptarg` as if they are inlined into a lambda body, i.e. `->(a = (return 1; 1)) {}.call` returns `1`.

> I still have hope they'd reconsider this https://bugs.ruby-lang.org/issues/15723 Not sure if it won't be better to wait a bit, as I see there's a reasonable amount of people who dislike this new syntax.

@bbatsov From what I see the discussion has stopped. Even if someday they'll decide to change it, for example, to `@` and support only one numbered parameter - then most probably it will look really similar to the current implementation. Anyway, we need to move forward.

bbatsov:
> @bbatsov From what I see the discussion has stopped. Even if someday they'll decide to change it, for example, to @ and support only one numbered parameter - then most probably it will look really similar to the current implementation. Anyway, we need to move forward.

Fair enough. Matz said in RubyKaigi he'll think more about this, but I agree that probably it won't see big changes.

whitequark:
This looks essentially OK, can you rebase and merge please?

iliabylich:
@whitequark Thanks!

Issue #571, opened 2019-04-17
ruby/ruby@9738f96 - pattern matching


iliabylich:
https://github.com/ruby/ruby/commit/9738f96fcfe50b2a605e350bdd40bd7a85665f54

iliabylich:
I'm slowly working on backporting this feature and got a bunch of questions.

Basically pattern matching works via existing `case` construction:
``` ruby
case value
in single_match if condition # conditional is optional
  # no destructuring
  single_match
in single_match,
  # destructuring like in the branch below because of the trailing comma
  single_match
in a, b, c unless condition # conditional is optional
  [a, b]
in ^existing_lvar then result3
in one, *rest
  one
in a:, b: 1
  [a, b]
else
  'no match'
end
```

`^` is a pin operator that prevents assignment and forces usage of existing local variable (just like in Elixir).

Here's the list of questions:
1. Should we emit `case` as `s(:case)`? `in` and `when` inside `case` are mutually exclusive, so it's either a `case` for pattern matching or a regular case that uses `===`. That could be useful to distinguish normal case and pattern matching. Also existing `case` has either `when` or `else` children, so I think adding a new type of child is a breaking change. Is that correct?
2. What should be the name of the node that wraps `in` pattern and its `then` branch? `s(:in)`?
3. What should be the name of the array and hash patterns? `array_pattern`/`hash_pattern`?
4. Should we emit non-pinned variables as `lvasgn`/`masgn` (depending on the amount of items passed to `in`) and pinned variables as `lvar`? Pattern matching technically has two steps: matching and (if matches) mass-assignment. I'm thinking about parsing `in a, ^b, *c` as
```
s(:in,
  s(:array_pattern,
    s(:lvasgn, :a),
    s(:lvar, :b),
    s(:splat, s(:lvasgn, :c))))
```
5. In the last branch you can see an example of matching a hash. Obviously we shouldn't use `kwarg`/`kwoptarg` nodes to represent them. Existing `s(:pair)` doesn't work as well (because sometimes there's no `value`). Maybe it's worth introducing a new node for that?

@whitequark @bbatsov @Drenmi @marcandre @mbj any thoughts?

bbatsov:
Is there some spec we can check about the pattern matching behaviour? I saw the feature doesn't have many tests and I recall reading somewhere that the full spectrum of pattern matching was quite complex. It's hard for me to advise on the node structure without knowing what's possible in general.

iliabylich:
[...]

PR #574, opened 2019-05-08
2.7 Pattern matching

iliabylich:
This commit tracks upstream commit ruby/ruby@9738f96.

Closes https://github.com/whitequark/parser/issues/571

No rush, this commit is 100% isolated (literally 0 deletions)

Specs cover 100% of the new grammar.

Feel free to to add comments!

A bunch of code for error handling is still missing, but at least that's something to discuss.

### DIFF
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -1821,3 +1821,322 @@ Format:
 "__ENCODING__"
  ~~~~~~~~~~~~ expression
 ~~~
+
+## Pattern matching
+
+### Using `in` modifier
+
+Format:
+
+~~~
+(in-match
+  (int 1)
+  (match-var :a))
+"1 in a"
+   ~~ operator
+ ~~~~~~ expression
+~~~
+
+### Case with pattern matching
+
+#### Without else
+
+Format:
+
+~~~
+(case-match
+  (str "str")
+  (in-pattern
+    (match-var :foo)
+    (lvar :bar)) nil)
+"case "str"; in foo; bar; end"
+ ~~~~ keyword             ~~~ end
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~ expression
+~~~
+
+#### With else
+
+Format:
+
+~~~
+(case-match,
+  (str "str")
+  (in-pattern
+    (match-var :foo)
+    (lvar :bar))
+  (lvar :baz))
+"case "str"; in foo; bar; else; baz; end"
+ ~~~~ keyword             ~~~~ else  ~~~ end
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ expression
+~~~
+
+### In clause
+
+Format:
+
+~~~
+(in-pattern
+  (match-var :foo)
+  (lvar :bar))
+"in foo then bar"
+ ~~ keyword
+        ~~~~ begin
+ ~~~~~~~~~~~~~~~ expression
+~~~
+
+### If guard
+
+This guard runs after matching, so it's not an `if` modifier.
+
+Format:
+
+~~~
+(in-pattern
+  (match-var :foo)
+  (if-guard
+    (lvar :bar)) nil)
+"in foo if bar"
+        ~~ keyword
+        ~~~~~~ expression
+~~~
+
+### Unless guard
+
+This guard runs after matching, so it's not an `unless` modifier.
+
+Format:
+
+~~~
+(in-pattern
+  (match-var :foo)
+  (unless-guard
+    (lvar :bar)) nil)
+"in foo unless bar"
+        ~~~~~~ keyword
+        ~~~~~~~~~~ expression
+~~~
+
+### Match variable
+
+Format:
+
+~~~
+(match-var :foo)
+"in foo"
+    ~~~ name
+    ~~~ expression
+~~~
+
+### Match rest
+
+#### With name
+
+Format:
+
+~~~
+(match-rest
+  (match-var :foo))
+"in *foo"
+    ~ operator
+    ~~~~ expression
+~~~
+
+#### Without name
+
+Format:
+
+~~~
+(match-rest)
+"in *"
+    ~ operator
+    ~ expression
+~~~
+
+### Pin operator
+
+Format:
+
+~~~
+(pin
+  (lvar :foo))
+"in ^foo"
+    ~ selector
+    ~~~~ expression
+~~~
+
+### Match alternative
+
+Format:
+
+~~~
+(match-alt
+  (pin
+    (lvar :foo))
+  (int 1))
+"in ^foo | 1"
+         ~ operator
+    ~~~~~~~~ expression
+~~~
+
+### Match with alias
+
+Format:
+
+~~~
+(match-as
+  (int 1)
+  (match-var :foo))
+"in 1 => foo"
+      ~~ operator
+    ~~~~~~~~ expression
+~~~
+
+### Match using array pattern
+
+#### Explicit
+
+Format:
+
+~~~
+(array-pattern
+  (pin
+    (lvar :foo))
+  (match-var :bar))
+"in [^foo, bar]"
+    ~ begin   ~ end
+    ~~~~~~~~~~~ expression
+~~~
+
+#### Explicit with tail
+
+Adding a trailing comma in the end works as `, *`
+
+Format:
+
+~~~
+(array-pattern-with-tail
+  (pin
+    (lvar :foo))
+  (match-var :bar))
+"in [^foo, bar,]"
+    ~ begin    ~ end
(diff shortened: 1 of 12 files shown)


iliabylich:
@whitequark The absence of comments scares me. Sorry to ask, but have you looked into updated AST_FORMAT?

@bbatsov @marcandre Could you take a look as well please?

marcandre:
I'm confused about `(array-pattern-with-tail ...)`. Wouldn't it be best to have `(array-pattern, ..., (match-rest))`, where `match-rest` has potentially no expression? I couldn't find examples of arrays with rest that wasn't at the end, but I think `in [1, *rest, 2]` is possible, so maybe I'm missing something.

marcandre:
Also, I didn't see guards `in ... if ...`; is the PR complete?

marcandre:
Finally, there seems to be no example of `match_with_trailing_comma` in the `AST_FORMAT` file.
(BTW, I mean these comments to be constructive and positive, but I'm in a rush today so I hope I don't come on too brusk :-) )

iliabylich:
> Wouldn't it be best to have `(array-pattern, ..., (match-rest),` where `match-rest` has potentially no expression?

All nodes have source maps, that's the rule. I don't think that we can break it.

> Also, I didn't see guards in ... if ...; is the PR complete?

Good catch, I'll put it into the doc.

> Finally, there seems to be no example of `match_with_trailing_comma`

That's an internal node that never appears in the final AST, it's used to represent an array item that has a trailiing comma after it - https://github.com/whitequark/parser/pull/574/files#diff-2f7266204a663f1f7e66698bdcd59cbfR1250

> so I hope I don't come on too brusk

This is exactly what I wanted to get, thank you!

whitequark:
> The absence of comments scares me. Sorry to ask, but have you looked into updated AST_FORMAT?

Yes. It generally seems OK architecturally. I haven't written any Ruby in years and I'm very disappointed in the language so I'm not particularly motivated to spend days studying every single detail in depth.

iliabylich:
Hmm, found (probably?) a bug with the `if` check in my PR:
``` ruby
class A
  def deconstruct
    puts 'deconstruct called'
    [1]
  end
end

p case A.new
in A[1] if (puts 'if check'; true)
  'yes'
else
  'no'
end
```

prints

```
deconstruct called
if check
"yes"
```

So Ruby first checks if the object matches given pattern and only after that it runs an `if` modifier. @whitequark Should we turn `if` into a child of the "match" node? Is it necessary? Because if so, this `if` doesn't have truthy/falsey branches (which I guess requires us to introduce one more AST node?)

bbatsov:
I'll try to find some time to review the PR as well. //cc @Drenmi @koic 

> Yes. It generally seems OK architecturally. I haven't written any Ruby in years and I'm very disappointed in the language so I'm not particularly motivated to spend days studying every single detail in depth.

@whitequark I completely understand you, as I my frustrations with Ruby have also been growing in recent years. I'm very grateful that you're still helping us out, though! Btw, if it's not too much to ask - what were you most disappointed about? What are you up to these days?

whitequark:
@iliabylich I'll take a look later, I don't feel well recently.

@bbatsov 

> Btw, if it's not too much to ask - what were you most disappointed about?

As I grew as a language designer, I realized that Ruby represents a lot of what I see as harmful patterns in language design. (Including, and especially, the very things that brought me to Ruby.) Sure, I'd still use it as I grew somewhat insensitive to language flaws—there's far too many so if I hung up on them I'd never get anything done—but I lost all patience for developing it, because that means thinking about and implementing features that I don't think should exist.

> What are you up to these days?

Mostly FPGAs. I've implemented (but, for the most part, not designed) [nMigen](https://github.com/m-labs/nmigen/), which is a Python eDSL for generating code that runs on FPGAs. I don't like Python any more than Ruby, but I think I made a good compromise with what I could work with. 

nMigen in Python because Python is far more common in the scientific and embedded areas, which are the key areas it serves. If left to my own devices, I'd probably use OCaml instead; that's a combination of practical and elegant that I just can't dislike.

iliabylich:
ok, so this `in 1 if 2` is not an `if modifier`, that's a guard which has slightly different semantics. Let's turn it into a `guard-if`/`guard-unless` node? Its only child is the given `condition` (because the code inside the `in` branch is the child of the `in-pattern` node), and its parent is the `in-pattern` node:
```
...
in 1 if 2
  3
...

s(:in_pattern,
  s(:int, 1),
  s(:guard_if, s(:int, 2)),
  s(:int, 3))
```
... and of course the guard is nullable

iliabylich:
@whitequark Could you take a look into last 3 commits please? I guess we are good to merge

whitequark:
@iliabylich Your changes look reasonable. But now I'm hesitant to merge it until it appears in a Ruby release. What other modifications is upstream going to pull on us?

iliabylich:
> What other modifications is upstream going to pull on us?

not sure if I understand the question. Are you asking what else is upcoming in 2.7?

> But now I'm hesitant to merge it until it appears in a Ruby release.

2.7.0-preview1 is out, so I'm pretty sure they won't change it. Have you ever seen ruby committers changing their decisions in terms of grammar changes?

Anyway, I'm ok with it. This PR still has 0 deletions, so it can wait. We can merge it anytime

whitequark:
> Have you ever seen ruby committers changing their decisions in terms of grammar changes?

Yeah? Not reverting one wholesale, but fiddling with it right up until release in a way that's annoying to downstream tooling devs.

Morozzzko:
Sorry, I just have to chime in

>     Have you ever seen ruby committers changing their decisions in terms of grammar changes?

That's exactly what happened with numbered parameters. See: https://bugs.ruby-lang.org/issues/15723#note-126

I doubt this will happen with PM, though. Still, not the problem: for the time being, we can just pull the new code from github.

```ruby
gem 'parser', github: 'iliabylich/parser', ref: 'pattern-matching'
```

palkan:
@iliabylich Looks like a bug with `__FILE__` & co:

```ruby
case [__FILE__, __LINE__ + 1, __ENCODING__]
in [__FILE__, __LINE__, __ENCODING__]
  true
end
```

Produces:

```
(case-match
  (array
    (str "(string)")
    (send
      (int 1) :+
      (int 1))
    (--ENCODING--))
  (in-pattern
    (array-pattern
      (--FILE--)
      (--LINE--)
      (--ENCODING--)) nil
    (true)) nil)
```

`__FILE__` and `__LINE__` are not transformed into literal nodes within `in_pattern`.

(This is one the Ruby's own tests: https://github.com/ruby/ruby/blob/9deca1a3b91271159e164188367f003978737f42/test/ruby/test_pattern_matching.rb#L356)

iliabylich:
@palkan Good catch! The fix is trivial, I'll apply it after backporting other commits mentioned in #571

palkan:
Another interesting case:

```ruby
case i; in a:; a==0 end
```
=>

```
(case-match
  (send nil :i)
  (in-pattern
    (hash-pattern
      (match-var :a)) nil
    (send
      (lvar :a) :==
      (int 0))) nil)
```

```ruby
case i; in "a":; a==0 end
```

=>

```
(case-match
  (send nil :i)
  (in-pattern
    (hash-pattern
      (match-var :a)) nil
    (send
      (send nil :a) :==
      (int 0))) nil)
```

Note that in the first case we have `(lvar :a) :==` and In the second `(send nil :a) :==`.

Adding `@parser.static_env.declare(name)` [here]( https://github.com/whitequark/parser/pull/574/files#diff-2f7266204a663f1f7e66698bdcd59cbfR1256) seems to fix it.

iliabylich:
I've backported all commits tracked in https://github.com/whitequark/parser/issues/571.

@palkan Both of your issues are fixed, thank you for testing and reporting.

2.7.0-preview2 has been released a few weeks ago, most changes that I backported today are really small (took me literally a few hours) and I haven't seen any active discussions on ruby bugtracker related to pattern matching.

Feel free to leave comments, if everything is ok I think it makes sense to merge it next week (and release `2.6.5.1` if anyone wants to play with it)

palkan:
@iliabylich Awesome. Works perfectly. I'm using this patch to [transpile new syntax to older Rubies](https://github.com/ruby-next/ruby-next/pull/9/files), and now I have the whole `test_pattern_matching.rb` passing  🙂 That means that parser's PM representations conforms with the Ruby's one.

Thanks for your work!

>  I haven't seen any active discussions on ruby bugtracker related to pattern matching

Keeping in mind the situation with the method reference operator, I wouldn't be so sure that nothing gonna happen 🙂

iliabylich:
2.7.0.0 has been released - https://rubygems.org/gems/parser/versions/2.7.0.0

Nakilon:
> features that I don't think should exist

You have intrigued me very much.

Issue #573, opened 2019-04-29
ruby/ruby@330b376


iliabylich:
https://github.com/ruby/ruby/commit/330b376133e00ae418bcf01e641e127df01fbc28

iliabylich:
also https://github.com/ruby/ruby/commit/1432471a759dc0cbc80c53766894dba45e6da887

PR #575, opened 2019-05-13
Reject \n and \r in heredoc identifiers starting from 2.7.

iliabylich:
This commit tracks upstream commits ruby/ruby@330b376 and ruby/ruby@1432471.

Closes https://github.com/whitequark/parser/issues/573.

I've added one more rule to reject the code like
``` ruby
<<"HERE
# EOF
```
Before this patch it'd throw `expected token tLSHIFT`


whitequark:
That should be +, since it's a 2.7 feature.

iliabylich:
> That should be +

Fixed, thanks.

Issue #579, opened 2019-05-28
ruby/ruby@c730c25


iliabylich:
https://github.com/ruby/ruby/commit/c730c25354a18e99b9147c30ecc8f986d6a172f1 - not sure if it needs any work. Need to check `assert_syntax_error` calls from tests in this commit.

iliabylich:
Here's the summary of this change:
+ MRI 2.6.4 accepts all chars from 0 to 127 after `"\C-"`
+ MRI 2.7.0-dev accepts all chars from 0 to 127 except 0-8, 14-31

Parser atm accepts all chars from 0 to 127 except 0, 4, 26 (not sure why)

whitequark:
These characters are EOF. 4 is `^D` (Unix), 26 is `^Z` (Windows), 0 is, well, null.

iliabylich:
Got it, thanks for the explanation. But MRI accepts these chars:
```
2.6.3 :001 > eval "\"\\C-" + 0.chr + "\""
 => "\u0000"
2.6.3 :002 > eval "\"\\C-" + 4.chr + "\""
 => "\u0004"
2.6.3 :003 > eval "\"\\C-" + 26.chr + "\""
 => "\u001A"
```

And it's not related to `eval`:
```
$ ruby -e "p File.read('test.rb').chars"
["p", " ", "\"", "\\", "C", "-", "\u0000", "\"", "\n"]
$ ruby test.rb
"\u0000"
```

I see the place in the `lexer.rl` that uses `c_any` which is set to `any - c_eof`, and I guess it should be `any` instead of `c_any` in the control/meta char rules.

whitequark:
Sounds about right.

PR #585, opened 2019-06-13
Reject whitespaces in meta and control chars.

iliabylich:
This commit tracks upstream commit ruby/ruby@c730c25.

Closes https://github.com/whitequark/parser/issues/579

meta and control chars are a bit broken, the algorithm of parsing escaped chars is recursive, and `\M-`/`\C-` prefixes work as functions:
```
"\C-<char>" => apply_C_mask_on(parse_escaped("<char>")
"\M-<char>" => apply_M_mask_on(parse_escaped("<char>")
```
however, in `lexer.rl` it's not recursive, and because of that it parse code like `"\C-\377"` incorrectly. I don't want to fix it for now, literally nobody uses syntax like this.

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -715,6 +715,14 @@ class Parser::Lexer
     diagnostic :fatal, :invalid_escape
   }
 
+  action read_post_meta_or_ctrl_char {
+    @escape = @source_buffer.slice(p - 1).chr
+
+    if @version >= 27 && ((0..8).include?(@escape.ord) || (14..31).include?(@escape.ord))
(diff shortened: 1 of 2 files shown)


dgollahon:
> it parse code like "\C-\377" incorrectly. 

If this is a parsing difference from MRI / creates invalid output, should it be added to the list of Known Issues in the readme? I can understand why some sufficiently obscure syntax might not be worth the effort to parse perfectly, but whenever I've used parser my expectation is that it will generate a perfectly valid AST (minus unintended bugs) except for the documented cases.

whitequark:
I agree that this should be either fixed or added to known issues in README, no matter how obscure it is.

Issue #527, opened 2018-08-21
procarg0 with parentheses


saberduck:
In the following snippet 

      [[1,2],[3,4]].each do |(a,b)| puts "#{a}:#{b}" end

Block arguments are parsed as 
```
(args
  (procarg0
    (arg :a)
    (arg :b)))
```

However it seems to me that `on_argument` handler in the `Processor` expects argument to have only two children -> [argument and value](https://github.com/whitequark/parser/blob/master/lib/parser/ast/processor.rb#L103) . This means that in this case node `(arg :a)` is never processed as argument because `process_all` is not invoked on it.

Am I right that decomposition in `on_argument` is wrong?

saberduck:
If it's relevant, real world code using this syntax can be found here 
https://github.com/discourse/discourse/blob/master/app/models/permalink.rb#L47

whitequark:
> Am I right that decomposition in `on_argument` is wrong?

Yes. Please add a separate visitor for `on_procarg0`.

PR #587, opened 2019-06-13
Add custom visitor for procarg0 that supports both single argument and mlhs.

iliabylich:
Closes https://github.com/whitequark/parser/issues/527

@whitequark Do you see any better way to detect single argument / mlhs?

### DIFF
--- a/lib/parser/ast/processor.rb
+++ b/lib/parser/ast/processor.rb
@@ -124,7 +124,22 @@ module Parser
       alias on_kwarg          process_argument_node
       alias on_kwoptarg       process_argument_node
       alias on_kwrestarg      process_argument_node
-      alias on_procarg0       process_argument_node
+
+      def on_procarg0(node)
+        if node.children[0].is_a?(Symbol)
+          # This branch gets executed when the builder
+          # is not configured to emit and 'arg' inside 'procarg0', i.e. when
+          #   Parser::Builders::Default.emit_arg_inside_procarg0
+          # is set to false.
+          #
+          # If this flag is set to true this branch is unreachable.
+          # s(:procarg0, :a)
+          on_argument(node)
+        else
+          # s(:procarg0, s(:arg, :a), s(:arg, :b))
+          process_regular_node(node)
+        end
+      end
 
       alias on_arg_expr       process_regular_node
       alias on_restarg_expr   process_regular_node
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -933,10 +933,15 @@ the sole argument (e.g. `|foo,|`).
 Format:
 
 ~~~
-(procarg0 :foo)
+(procarg0 (arg :foo))
 "|foo|"
   ~~~ expression
-  ~~~ name
+
+(procarg0 (arg :foo) (arg :bar))
+"|(foo, bar)|"
+  ~ begin
+           ~ end
+  ~~~~~~~~~~ expression
 ~~~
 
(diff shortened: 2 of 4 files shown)


whitequark:
Hm, why do we emit `s(:procarg0, :a)` in the first place? Shouldn't it be `s(:procarg0, s(:arg, :a))`?

iliabylich:
Because we do `arg.updated(:procarg0)`, and sometimes it's a `s(:arg)`, sometimes it's a `(:mlhs, s(:arg, :a), s(:arg, :b))`

Should we fix the format of AST?

whitequark:
Yeah, that sounds all around better. But if it breaks user code it might need an opt-in...

iliabylich:
Are you talking about adding one more flag like `Builder.emit_mlhs_inside_procarg0`?

whitequark:
Yeah. Though we should probably not emit mlhs but only change `:x` to `(arg :x)`.

iliabylich:
> Though we should probably not emit mlhs but only change :x to (arg :x). 

Right. to sum up, if the flag is enabled:
+ `a` becomes `s(:procarg0, s(:arg, :a))` (that's the change)
+ `(a, b)` is still `s(:procarg0, s(:arg, :a), s(:arg, :b))`

if it's disabled we use current behavior:
+ `a` is `s(:procarg0, :a)`
+ `(a, b`) is `s(:procarg0, s(:arg, :a), s(:arg, :b))`

And then what should be the name of the flag? `Builder.emit_arg_inside_procarg0`?

And `Builder.modernize` should enable it, right?

Also it's a bit unclear what is the source map of this node. Right now it's either `Map::Collection` (if it was generated from `mlhs_node.updated(:procarg0)`) or `Map::Variable` (for `arg_node.updated(:procarg0)`). I guess it should be `Map::Collection` because it wraps 1+ arguments, but could you confirm it please?

whitequark:
> Right. to sum up

Correct.

> And then what should be the name of the flag? `Builder.emit_arg_inside_procarg0`?

That's fine.

> And `Builder.modernize` should enable it, right?

Correct. `modernize` enables all flags.

> I guess it should be `Map::Collection` because it wraps 1+ arguments, but could you confirm it please?

Again sounds fine.

iliabylich:
Done. @whitequark PTAL when you have time.

I wasn't able to find any references to `Builder` in `AST_FORMAT`, so I assume it documents a "modern" version of AST. Or it a problem with documentation? 😄

whitequark:
> so I assume it documents a "modern" version of AST

It does.

lib/parser/ast/processor.rb
```diff
@@ -124,7 +124,16 @@ def process_argument_node(node)
       alias on_kwarg          process_argument_node
       alias on_kwoptarg       process_argument_node
       alias on_kwrestarg      process_argument_node
-      alias on_procarg0       process_argument_node
+
+      def on_procarg0(node)
+        if node.children[0].is_a?(Symbol)
+          # s(:procarg0, :a)
```
whitequark: I think we don't need this branch anymore.

iliabylich: But what if the builder is not "modernized"?

whitequark: Then it works just like before.

But I think it's fine to leave this in. Just needs a comment to indicate why this is using an AST format that's not what our docs describe.

iliabylich:
@whitequark Fixed. Anything else to change/fix before merging?

Issue #515, opened 2018-06-30
--emit-json fails with empty / commented files


lsenta:
Hi there, thanks for maintaining this library,
I just started using Ruby & ruby-parse,
it crashes on empty files or files that contains only comments:

```
› echo "# helloworld" > helloworld.rb && ruby-parse ./helloworld.rb --emit-json
Failed on: ./helloworld.rb
Traceback (most recent call last):
        13: from /usr/local/bin/ruby-parse:23:in `<main>'
        12: from /usr/local/bin/ruby-parse:23:in `load'
        11: from /usr/local/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/bin/ruby-parse:7:in `<top (required)>'
        10: from /usr/local/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/runner.rb:13:in `go'
         9: from /usr/local/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/runner.rb:35:in `execute'
         8: from /usr/local/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/runner/ruby_parse.rb:137:in `process_all_input'
         7: from /usr/local/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/runner.rb:174:in `process_all_input'
         6: from /usr/local/Cellar/ruby/2.5.1/lib/ruby/2.5.0/benchmark.rb:293:in `measure'
         5: from /usr/local/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/runner.rb:176:in `block in process_all_input'
         4: from /usr/local/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/runner.rb:196:in `process_files'
         3: from /usr/local/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/runner.rb:196:in `each'
         2: from /usr/local/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/runner.rb:207:in `block in process_files'
         1: from /usr/local/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/runner.rb:214:in `process_buffer'
[...]

PR #590, opened 2019-06-17
Print empty string when --emit-json and empty input are given.

iliabylich:
```
$ bin/ruby-parse --emit-json -e ''
warning: parser/current is loading parser/ruby26, which recognizes
warning: 2.6.3-compliant syntax, but you are running 2.6.0.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.

```

Closes https://github.com/whitequark/parser/issues/515


bbatsov:
Maybe this should get an unit test?

iliabylich:
@bbatsov Good call

PR #611, opened 2019-09-16
Detect and warn about empty else

iliabylich:
`else_` is nil in `def m; else; end`, but `else_t` is there.

### DIFF
--- a/lib/parser/macruby.y
+++ b/lib/parser/macruby.y
@@ -81,7 +81,7 @@ rule
                       else_t,   else_   = val[2]
                       ensure_t, ensure_ = val[3]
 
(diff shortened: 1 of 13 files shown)


whitequark:
What's the motivation here?

iliabylich:
to warn/reject code like `def m; else; end` because it's invalid

whitequark:
Ah, I misread the code.

iliabylich:
well, let me describe it better: useless else (i.e. without `rescue`) gives a warning before 2.6 and an error after. But if it's empty we don't generated this warning/error. Checking for `else_t` works for both empty and non-empty else

Issue #610, opened 2019-09-16
ruby/ruby@3a3f48f


iliabylich:
https://github.com/ruby/ruby/commit/3a3f48fb8fbdbb810d9b675159529970015316b9

PR #613, opened 2019-09-18
+ lexer.rl: support comments before leading dot in 27 mode.

iliabylich:
This commit tracks upstream commit ruby/ruby@3a3f48f.

Closes https://github.com/whitequark/parser/issues/610

I've added a bunch comments, please LMK if there's anything that could be improved.

Also I've fixed parsing of `.:` operator for pre-27 lexer. Previously it was emitted as `tMETHREF` for all versions of lexer, but since it's a 27-specific token we had an "unknown token " error for pre-27 parsers.

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -2465,7 +2465,25 @@ class Parser::Lexer
   leading_dot := |*
       # Insane leading dots:
       # a #comment
+      #  # post-2.7 comment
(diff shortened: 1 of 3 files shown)


whitequark:
Looks good, but I'd prefer to not have a bugfix and a feature in one commit.

iliabylich:
@whitequark Done, could you look one more time please?

Issue #616, opened 2019-09-24
2.7: numparams -> implicit params


iliabylich:
https://github.com/ruby/ruby/commit/e73cc3eead75b27eabf93a7163bed511f2bb3b17
https://github.com/ruby/ruby/commit/ea68bb914a3c806a1c5188993b96791a76ab0849
https://github.com/ruby/ruby/commit/0e84eecc1764ec6e2a28654b3e66b8c006f73432

The first commit has 0 tests, so it needs some investigation

iliabylich:
also
https://github.com/ruby/ruby/commit/e663299a5f7a980aae39f3e87cec064ae19a870c
https://github.com/ruby/ruby/commit/55e1e22b2d44a8a1eca0f6ed2b11dc0f564f7192
https://github.com/ruby/ruby/commit/33c5ad3154099769cec7f77b6d95cbf4b8d52bb0

iliabylich:
Alright, this sequence of changes is simply about renaming `@1 - @9` to `_1 - _9`. First change (that I really like by the way, but we don't need to do anything about it, because there's no `procarg0` for `numblock`):
``` ruby
# previously:
proc { @1 }.call([1, 2])
 => 1

# now:
proc { [_1] }.call([1, 2])
 => [[1, 2]]
```
so a single argument is now `|x|` (not `|x,|`)

The next thing is that this `_N` is a local variable:
``` ruby
2.1.8 :027 > -> { _1; local_variables }.call(1)
 => [:_1, :_]
```

And it can be used in assignment:
``` ruby
2.1.8 :047 > proc { _1 = 'a'; _1 }.call('b')
(irb):47: warning: `_1' is used as numbered parameter
 => "a"
```

But in this case it's no longer a parameter:
``` ruby
2.1.8 :049 > ->{ _1 = 'a'; _1 }.call('b')
(irb):49: warning: `_1' is used as numbered parameter
Traceback (most recent call last):
        4: from -e:1:in `<main>'
        3: from <internal:prelude>:206:in `irb'
        2: from (irb):49:in `<main>'
        1: from (irb):49:in `block in <main>'
ArgumentError (wrong number of arguments (given 1, expected 0))
2.1.8 :050 > ->{ _1 = 'a'; _1 }.arity
(irb):50: warning: `_1' is used as numbered parameter
 => 0
```

... which means that "reading" this parameter declares it, simple assignment works as a local variable assignment, here's an example (`_1` is a parameter, but `_2` is not):
[...]

PR #620, opened 2019-10-17
- ruby27.y: changed prefix of numparams to "_"

iliabylich:
This commit tracks upstream commits ruby/ruby@e73cc3e, ruby/ruby@ea68bb9, ruby/ruby@0e84eec, ruby/ruby@e663299, ruby/ruby@55e1e22, ruby/ruby@33c5ad3, ruby/ruby@c2065c6.

Closes https://github.com/whitequark/parser/issues/616

1. Since new numparams have `_` prefix they are local variables, so I moved `MaxNumparamStack` to the parser
2. Added a `Builder#check_and_register_numparam` method that registers given local variable as a numbered parameter
3. Updated `AST_FORMAT.md`, `META::NODE_TYPES`, error messages and tests. I removed a few of them simply because there are no such errors anymore (`_0`/`_100` are valid lvar names).

I don't really like what I've got in the `Builder#check_and_register_numparam`, this method knows about internals of `Context` and `MaxNumparamStack`, but at the same time introducing objects around layers of these stacks would affect performance. Not sure if it's worth it.


lib/parser/builders/default.rb
```diff
@@ -817,6 +812,54 @@ def objc_restarg(star_t, name=nil)
...
+            # found an outer scope that can't have numparams
+            # like def/class/etc
+            break
+          end
+        end
+
+        @parser.static_env.declare(name)
+        @parser.max_numparam_stack.register(name[1].to_i)
+      end
+    end
```
whitequark: My gut feel is that this shouldn't be in the builder. But maybe we don't have the infrastructure to do it any other way.

whitequark:
By the way, this PR is as good time as any to discuss something that's been on my mind. I think you are at least as good a maintainer of Parser as I am. If you are fine with this position, I would like to completely hand off day-to-day maintenance to you; not only is being a critical path (by virtue of having reviews blocked on me) a burden to me, but also, having not followed Ruby development for years at this point, I am not really qualified to make many kinds of nuanced but ultimately architecturally bland decisions like "what precise shape should this AST node be".

What do you think?

iliabylich:
@whitequark I'm ok with it. Does it mean that I shouldn't ask you for review even I have some questions? Or is it ok to ask and I just shouldn't block merging PRs by your review? In that case, should I review PRs from other people (if any)?

lib/parser/builders/default.rb
iliabylich: Because it doesn't look generic and may change in the next version of MRI? I guess you are right. Would you move it straight into parser?

whitequark:
I suggest you treat Parser as your own project, using your judgement for day-to-day changes, and yes, collaborating with any other contributors. Of course, you can still ask me when in doubt, and I'll do my best to provide context or suggestions, but my opinions are neither infallible nor extraordinarily valuable, given that it has been many years since I touched Parser or Ruby. All I do is follow three general rules: (a) when adding code, do it in a similar way to code that's already there, (b) don't change the AST format without an opt-in, and (c) design APIs that are hard to misuse (and not only easy to use), including accidental misuse as a consequence of future changes.

lib/parser/builders/default.rb
whitequark: Yeah. I'm not really sure what's best here. But moving it into parser is one option.

There are other places where this happens, like `accessible` and `assignable`, but those seem more straightforward (and we can't change them anyway due to backwards compat).

iliabylich: ok, I moved it straight into parser. I tried implementing it in a more declarative way using enumerators and objects with predicates around stack values, but it still looks complicated. I feel like it should be extracted into a separate class, but then comes the problem of changing it for next rubies. I guess inlining is the best solution.

Issue #621, opened 2019-10-17
ruby/ruby@0162e7e - circular argument reference error


iliabylich:
https://github.com/ruby/ruby/commit/0162e7e6471b639dfeeded29943e9e27c9519826

iliabylich:
@whitequark design question: this PR turns optional arguments with circular references (like `def m(a = a); end`) into syntax errors. Is it worth adding `version` field into `Builder`? I guess that'd be a breaking change, right?

iliabylich:
I think I could use `---- inner` section of the parser to define version-specific methods (ugly `check_and_register_numparam` from the previous PR could also be placed there)

PR #622, opened 2019-10-21
- ruby27.y: reject circular argument reference.

iliabylich:
This commit tracks upstream commit ruby/ruby@0162e7e.

Closes https://github.com/whitequark/parser/issues/621

```
$ bin/ruby-parse --27 -e 'm { |a: proc { 1 + a }| }'
(fragment:0):1:6: error: circular argument reference a
(fragment:0):1: m { |a: proc { 1 + a }| }
(fragment:0):1:      ^             ~
```

### DIFF
--- a/lib/parser.rb
+++ b/lib/parser.rb
@@ -75,6 +75,10 @@ module Parser
   require 'parser/context'
   require 'parser/max_numparam_stack'
 
+  module Helpers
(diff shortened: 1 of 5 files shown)


iliabylich:
@whitequark PTAL when you have time, I'm using a custom AST visitor to throw this error. Completely unsure if it's a good idea, but it's necessary to handle complex ASTs.

Issue #638, opened 2019-11-14
ruby/ruby@4d615a0


iliabylich:
https://github.com/ruby/ruby/commit/4d615a0c8f4759ce4f6b96b643ef957837bdf20e

PR #639, opened 2019-11-20
- lexer.rl: parse embedded ivars/cvars starting with digit as str

iliabylich:
This commit tracks upstream commit ruby/ruby@4d615a0.

Closes https://github.com/whitequark/parser/issues/638

Before 2.7 code like
``` ruby
"#@1"
```
was interpreted as interpolation of the `@1` instance variable. Starting from 2.7 it's just a plain string.

@whitequark Could you take a look please? Do you see any better solution? Initially I tried to stick to existing `interp_var` rule (by setting a flag in the local variable and checking it in the code that emits ivar). However, its handler (`extend_interp_var`) emits `#{` and I don't think that un-doing it (and doing `literal.extend_string` afterwards) is a reliable solution.

By the way, have you ever considered branching lexer?

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -1019,6 +1019,20 @@ class Parser::Lexer
     fcall expr_variable;
   }
 
+  # Special case for Ruby > 2.7
+  # If interpolated instance/class variable starts with a digit we parse it as a plain substring
+  # However, "#$1" is still a regular interpolation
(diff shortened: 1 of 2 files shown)


whitequark:
> By the way, have you ever considered branching lexer?

I have not. Unlike with the parser (where it's simply impossible), in the lexer you can add version branches. There are 17 version checks and half of them are for Ruby 1.8, so it doesn't seem like 9 version checks warrant branching the entire thing.

PR #642, opened 2019-12-27
- dedenter.rb: prevent `ArgumentError` when processing binary encoded code

koic:
This PR prevents the following error when processing binary encoded code.

```ruby
"\xC0".split # => ArgumentError (invalid byte sequence in UTF-8)
```

RuboCop's CI has detected this error.
https://circleci.com/gh/rubocop-hq/rubocop/78846


iliabylich:
@koic 2.7.0.1 has been released. What's interesting the error appears only if the `Builder` ignores invalid encoding of the string in the `string_value` method (IIRC rubocop patches it).

PR #652, opened 2020-02-21
Improve pattern matching compatibility

palkan:
This PR aims to fix a few incompatibilities with MRI (based on the recently published [ruby/spec](https://github.com/ruby/spec/blob/master/language/pattern_matching_spec.rb) tests for pattern matching).

- [x] Raise syntax error if pin points to unknown local variable (see https://github.com/ruby/spec/blob/10fe05ce8f82eae325a6f259f67a054b9f58fb58/language/pattern_matching_spec.rb#L372-L384)

- [x] [**WONTFIX**](https://github.com/whitequark/parser/pull/652#issuecomment-590318925) ~~Disallow variable binding in alternation (see https://github.com/ruby/spec/blob/10fe05ce8f82eae325a6f259f67a054b9f58fb58/language/pattern_matching_spec.rb#L396-L405)~~

- [x] Empty array should return array-pattern anyway (see https://github.com/ruby/spec/blob/10fe05ce8f82eae325a6f259f67a054b9f58fb58/language/pattern_matching_spec.rb#L501-L512) (right now `Object[]` behaves the same as `Object`).

- [x] Disallow string interpolation in keys (see https://github.com/ruby/spec/blob/10fe05ce8f82eae325a6f259f67a054b9f58fb58/language/pattern_matching_spec.rb#L739-L749)

```ruby
case {a: 1}
  # this should raise syntax error
  in {"#{x}": 1}
end
```

- [x] Respect `else; end`.
For `case ... in` the empty else clause has a different meaning than no else clause (the first one returns `nil`, the second raises NoMatchingPattern); thus, we need to make them distinguishable in the resulting AST.

### DIFF
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -1870,6 +1870,27 @@ Format:
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ expression
 ~~~
 
+#### With empty else
+
+Empty `else` differs from the missing (or _implicit_) `else` for pattern matching, since
+the latter one raises a `NoMatchingPattern` exception. Thus, we need a way to distinguish this
+two cases in the resulting AST.
+
+Format:
+
+~~~
+(case-match,
+  (str "str")
+  (in-pattern
+    (match-var :foo)
+    (lvar :bar))
+  (empty-else))
+"case "str"; in foo; bar; else; end"
+ ~~~~ keyword             ~~~~ else
+                                ~~~ end
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ expression
+~~~
+
 ### In clause
 
 Format:
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -2140,3 +2140,20 @@ Format:
     ~ name (const-pattern.const)
     ~ expression (const-pattern.const)
 ~~~
+
+#### With array pattern without elements
+
+Format:
+
+~~~
+(const-pattern
+  (const nil :X)
+  (array-pattern))
+"in X[]"
+     ~ begin (const-pattern)
+      ~ end (const-pattern)
+     ~~ expression (const-pattern)
+    ~ name (const-pattern.const)
+    ~ expression (const-pattern.const)
+     ~~ expression (const-pattern.array_pattern)
+~~~
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -1371,11 +1371,13 @@ module Parser
         pair_keyword(label, value)
       else
         begin_t, parts, end_t = label
+        label_loc = loc(begin_t).join(loc(end_t))
 
         # quoted label like "label": value
         if (var_name = static_string(parts))
-          loc = loc(begin_t).join(loc(end_t))
-          check_duplicate_pattern_key(var_name, loc)
+          check_duplicate_pattern_key(var_name, label_loc)
+        else
+          diagnostic :error, :pm_interp_in_var_name, nil, label_loc
         end
 
         pair_quoted(begin_t, parts, end_t, value)
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -1213,6 +1213,7 @@ module Parser
     #
 
     def case_match(case_t, expr, in_bodies, else_t, else_body, end_t)
+      else_body = n(:empty_else, nil, token_map(else_t)) if else_t && !else_body
       n(:case_match, [ expr, *(in_bodies << else_body)],
         condition_map(case_t, expr, nil, nil, else_t, else_body, end_t))
     end
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -1313,6 +1313,8 @@ module Parser
     end
 
     def array_pattern(lbrack_t, elements, rbrack_t)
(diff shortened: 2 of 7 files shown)


palkan:
@iliabylich Hey,

That's my first time contributing to Parser, so I'd like to get some feedback before continuing.
I've just covered the first item; is that a right direction?)

Also, have some thoughts regarding the second one (variable binding in alternation).

Since `match_var` might occur in nested arrays and hashes, I don't think we can handle this by the grammar changes. Instead, I'm thinking about adding yet another `parser.pattern_mathching_something` variable to track whether we're in the alternation context or not (and add `check_alternation_match_var` guards to all `match_var`-like methods). WDYT? Do you see a better way of dealing with this?

lib/parser/ruby27.y
```diff
@@ -2096,7 +2096,7 @@ opt_block_args_tail:
 
        p_var_ref: tCARET tIDENTIFIER
                     {
-                      lvar = @builder.accessible(@builder.ident(val[1]))
+                      lvar = @builder.accessible_for_pin(@builder.ident(val[1]))
```
iliabylich: I think there's no need to create a separate method on a `Builder`. You have a local variable name in the `val[1]`, so you can do an inline check using `@static_env`

iliabylich:
> I've just covered the first item; is that a right direction?

Yes, and thanks for reporting it! Are you going to fix them? If not I can take a look later this week.

> Also, have some thoughts regarding the second one (variable binding in alternation).

This case is special as it's not a part of the grammar:

```sh
$ ./ruby -vce 'case [0, 1]; in [0, 0] | [0, a]; end'
ruby 2.8.0dev (2020-02-12T15:15:22Z master bc1dbed16c) [x86_64-darwin19]
-e:1: warning: Pattern matching is experimental, and the behavior may change in future versions of Ruby!
Syntax OK

$ ./ruby -ve 'case [0, 1]; in [0, 0] | [0, a]; end'
ruby 2.8.0dev (2020-02-12T15:15:22Z master bc1dbed16c) [x86_64-darwin19]
-e:1: warning: Pattern matching is experimental, and the behavior may change in future versions of Ruby!
-e:1: illegal variable in alternative pattern (a)
Traceback (most recent call last):
-e: compile error (SyntaxError)
```

So this is another place in `compile.c` that throws a "runtime" syntax error - https://github.com/ruby/ruby/blob/master/compile.c#L5900. I don't think that it should be backported. Ripper also successfully parses it (because it's not a "grammar" error). `compile.c` has many more places throwing syntax error, some of them can be backported, some of them can't (like special syntax errors that exist only when you do `eval('code')`). I'd ignore all them.

If you need you can detect it in your code, should be quite simple using a stateful visitor.

palkan:
>  Are you going to fix them? 

Yep.

> compile.c has many more places throwing syntax error

Oh, weird Ruby 😁

> If you need you can detect it in your code, should be quite simple using a stateful visitor.

Makes sense. Thanks!

palkan:
@iliabylich Finished all the items. Ready for review)

lib/parser/builders/default.rb
```diff
@@ -1213,6 +1213,7 @@ def begin_keyword(begin_t, body, end_t)
     #
 
     def case_match(case_t, expr, in_bodies, else_t, else_body, end_t)
+      else_body = n(:blank, nil, token_map(else_t)) if else_t && !else_body
       n(:case_match, [ expr, *(in_bodies << else_body)],
         condition_map(case_t, expr, nil, nil, else_t, else_body, end_t))
```
iliabylich: we pass `else_t` here to the source map. It seems to be enough to distinguish no-else/empty-else. Do we really need this `blank` node here?

doc/AST_FORMAT.md
```diff
@@ -1870,6 +1870,23 @@ Format:
...
+
+Format:
+
+~~~
+(case-match,
+  (str "str")
+  (in-pattern
+    (match-var :foo)
+    (lvar :bar))
+  (blank))
```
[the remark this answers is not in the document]
palkan: My initial suggestion was `blank_else` )) But I've realized that it has nothing specific with `else`.

What about `void`? For example, Ripper has `void_stmt`.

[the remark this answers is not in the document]
iliabylich: PTAL at the last comment below, I'm not sure if we need it all. There's an information about `else_t` in the `case_match` node.

lib/parser/builders/default.rb
palkan: My opinion is that the difference should be a part of AST not metadata. Using `else_t` could be enough but looks like a hack to me.

For example, when calling `ruby-parse` you can not see the difference:

```sh
$ ruby-parse --27 -e 'case a; in a; end'
(case-match
  (send nil :a)
  (in-pattern
    (match-var :a) nil nil) nil)

# without blank node it's the same as above
$ ruby-parse --27 -e 'case a; in a; else; end'
(case-match
  (send nil :a)
  (in-pattern
    (match-var :a) nil nil) nil)

# with blank node
$ ruby-parse --27 -e 'case a; in a; else; end'
(case-match
  (send nil :a)
  (in-pattern
    (match-var :a) nil nil)
  (blank))
```

iliabylich: Good point. Let's rename it to `empty-else` then?

palkan: Done!

iliabylich:
@palkan Thanks a lot! `2.7.0.3` has been released

PR #659, opened 2020-03-20
+ ruby27.y: fix array pattern with tail source map

palkan:
This change takes trailing comman into account for `array_pattern_with_tail` expression location:

```
# before
in foo,
   ~~~ expression

# after
in foo,
   ~~~~ expression
```

Currently, we lose comma location information, which makes it hard for rewriters to deal with this case.

### DIFF
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -2068,7 +2068,7 @@ so a single item match with comma gets interpreted as an array.
 (array-pattern-with-tail
   (match-var :foo))
 "in foo,"
-    ~~~ expression
+    ~~~~ expression
 ~~~
 
 ### Matching using hash pattern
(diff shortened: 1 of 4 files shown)


palkan:
@iliabylich Another source map related question I would like to ask about first: what is the reason for the existing `const_pattern` behavior with `expression` to not including the const itself?

```
in A[0,1]
    ~~~~~ expression (in_pattern.const_pattern)
   ~ expression (in_pattern.const_pattern.const)
```

Do we already have such examples when a child node's expression is _located_ outside of the parent node?

iliabylich:
> Do we already have such examples when a child node's expression is located outside of the parent node?

No, that's a bug. Could you send a PR to fix it please? If not I'll find some time next week.

palkan:
> Could you send a PR to fix it please?

Sure.

PR #666, opened 2020-04-06
Allow running without mutable strings

jscheid:
Hi, first off, thanks for this exceptionally well-made piece of software.

I need to run it in a project that gets converted to JavaScript via [Opal](https://github.com/opal/opal). Opal doesn't support mutable strings, which breaks here: https://github.com/whitequark/parser/blob/fad6bd2ecfb4877d52a313420a07d0fd8a36dcb7/lib/parser/lexer/literal.rb#L199

I've tried finding an alternate buffer implementation that works without mutable strings, passes all tests, and still performs reasonably well. I've succeeded in the first two points with something based on StringIO, but I couldn't get it faster than about 10% slower compared to the current implementation.

Now, this was measuring just the appending to and taking strings from the buffer, where only part of the overall runtime is spent. The total impact is probably more in the single percent digits, if that. I haven't bothered measuring it, since I'm conscious of the fact that compatibility with Opal is very much an edge case and so any significant performance impact, however small, feels wrong.

That's why I would like to propose this change instead. It would allow swapping out the buffer implementation for a custom one. The included test shows what that might look like.

There is an additional call to `to_s` every time a string is taken out of the buffer, but of course that is a no-op when called on String so the only overhead it adds is the method call itself. Do you agree it is negligible, all things considered?

### DIFF
--- a/lib/parser/lexer/literal.rb
+++ b/lib/parser/lexer/literal.rb
@@ -37,6 +37,14 @@ module Parser
     attr_reader   :heredoc_e, :str_s, :dedent_level
     attr_accessor :saved_herebody_s
 
+    class << self
+      attr_writer :buffer_prototype
+
+      def buffer_prototype
+        @buffer_prototype || ''
+      end
+    end
+
     def initialize(lexer, str_type, delimiter, str_s, heredoc_e = nil,
                    indent = false, dedent_body = false, label_allowed = false)
       @lexer       = lexer
@@ -151,7 +159,7 @@ module Parser
           emit(:tLABEL_END, @end_delim, ts, te + 1)
         elsif @monolithic
           # Emit the string as a single token.
-          emit(:tSTRING, @buffer, @str_s, te)
+          emit(:tSTRING, @buffer.to_s, @str_s, te)
         else
           # If this is a heredoc, @buffer contains the sentinel now.
           # Just throw it out. Lexer flushes the heredoc after each
@@ -206,7 +214,7 @@ module Parser
       end
 
       unless @buffer.empty?
-        emit(:tSTRING_CONTENT, @buffer, @buffer_s, @buffer_e)
+        emit(:tSTRING_CONTENT, @buffer.to_s, @buffer_s, @buffer_e)
 
         clear_buffer
         extend_content
@@ -246,7 +254,7 @@ module Parser
     end
 
     def clear_buffer
-      @buffer = ''.dup
+      @buffer = self.class.buffer_prototype.dup
 
       # Prime the buffer with lexer encoding; otherwise,
       # concatenation will produce varying results.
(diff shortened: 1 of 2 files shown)


bbatsov:
Someone should check the real-world impact of those changes on tools like RuboCop, which are very performance-sensitive. I'm all for Opal support, but we have to be mindful of any fallback from this. Can you run RuboCop on a big project with and without the patch and compare the results?

jscheid:
Good idea, I'll do now.

iliabylich:
Opal already has a patch - https://github.com/opal/opal/blob/master/lib/opal/parser/patch.rb#L28

jscheid:
Oh, good find. Why didn't I see that? But, you can't promise not to break it, can you? 😄 

I'm going to run with that monkey patch if you won't accept a proper one, no problem. However, do first take a look at this:

So I've hacked a test case to compare real-world performance by running Rubocop on latest Rails master and thought to myself, why not compare it to the ideal solution as well -- which is the patch that I had intended originally, that uses StringIO for everything. It turns out it's not faring too badly.

Here is the "full" patch: https://github.com/jscheid/parser/commit/f02b6de86c5f50dc1a67ead4de906c4e358a93c3
Here is the test case: https://github.com/jscheid/parser-perf-comparison
Here are the logs: https://gist.github.com/jscheid/1295698227dd810cd730f4383621ba9f

Here are the timings (3rd out of three runs):

Without a patch:
```
real	1m15.044s
user	1m14.367s
sys	0m0.589s
```

With the patch in this PR:
```
real	1m14.398s
user	1m13.724s
sys	0m0.589s
```

With the "full" patch (https://github.com/jscheid/parser/commit/f02b6de86c5f50dc1a67ead4de906c4e358a93c3):
```
real	1m14.902s
user	1m14.244s
sys	0m0.583s
```

Now, apparently there's some kind of cache warming going on here (if you look at the three iterations without a patch, for instance) so maybe I should run the whole thing again with five or ten iterations instead of three. In fact I'll do that now, but will take a little while to complete.

Yet, I reckon it's safe to say that if real-world performance is your concern, and Rubocop is your benchmark, then either patch should be fine to accept?

iliabylich:
Opal is literally the only library that needs it. Is it worth changing parser's code to support it? I'd prefer Opal to implement it.

Plus, benchmark values depend a lot on what is parsed. IIRC strings/heredocs suffer a lot of from this change (yes, it's been proposed multiple times before 😄 )

jscheid:
Fair enough. I wouldn't have opened this PR if I had known about the patch that ships with Opal. Thanks for pointing it out to me, I'll use that... all good.

For what it's worth, this change (or the other one) doesn't have much of an effect on Rubocop running on Rails. Fastest out of ten runs in terms of wall time:

```
real    1m14.865s # no patch
real    1m14.027s # with to_s
real    1m14.649s # with StringIO
```

PR #674, opened 2020-04-14
+ Source::TreeRewriter: Add #merge, #merge! and #empty?

marcandre:
This PR adds methods for merging `Source::TreeRewriter`s.

There is also a simple method to check if a `Source::TreeRewriter` is `empty?`, i.e. if it has received any non-trivial command.

I wanted to use these methods to improve the [autocorrection's API of rubocop](https://github.com/rubocop-hq/rubocop/pull/7868), which could benefit from building a `TreeRewriter`, check if any correction was actually made, and combine them later in the process.

### DIFF
--- a/lib/parser/source/tree_rewriter.rb
+++ b/lib/parser/source/tree_rewriter.rb
@@ -118,6 +118,43 @@ module Parser
       end
 
       ##
+      # Returns true iff no (non trivial) update has been recorded
+      #
+      # @return [Boolean]
+      #
+      def empty?
+        @action_root.empty?
+      end
+
+      ##
+      # Merges the updates of argument with the receiver.
+      # Policies of the receiver are used.
+      #
+      # @param [Rewriter] with
+      # @return [Rewriter] self
+      # @raise [ClobberingError] when clobbering is detected
+      #
+      def merge!(with)
+        raise 'TreeRewriter are not for the same source_buffer' unless
+          source_buffer == with.source_buffer
+
+        @action_root = @action_root.combine(with.action_root)
+        self
+      end
+
+      ##
+      # Returns a new rewriter that consists of the updates of the received
+      # and the given argument. Policies of the receiver are used.
+      #
+      # @param [Rewriter] with
+      # @return [Rewriter] merge of receiver and argument
+      # @raise [ClobberingError] when clobbering is detected
+      #
+      def merge(with)
+        dup.merge!(with)
+      end
+
+      ##
       # Replaces the code of the source range `range` with `content`.
       #
       # @param [Range] range
@@ -254,6 +291,10 @@ module Parser
       ].join("\n").freeze
 
       extend Deprecation
+
+      protected
+
+      attr_reader :action_root
 
       private
 
--- a/lib/parser/source/tree_rewriter/action.rb
+++ b/lib/parser/source/tree_rewriter/action.rb
@@ -25,10 +25,16 @@ module Parser
         freeze
       end
 
-      # Assumes action.children.empty?
       def combine(action)
-        return self unless action.insertion? || action.replacement # Ignore empty action
+        return self if action.empty? # Ignore empty action
         do_combine(action)
+      end
+
+      def empty?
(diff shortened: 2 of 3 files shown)


lib/parser/source/tree_rewriter.rb
```diff
@@ -117,6 +117,43 @@ def initialize(source_buffer,
         @action_root = TreeRewriter::Action.new(all_encompassing_range, @enforcer)
       end
 
+      ##
+      # Returns true iff no (non trivial) update has been recorded
```
iliabylich, line 121: typo: `iff` -> `if`

lib/parser/source/tree_rewriter/action.rb
```diff
@@ -69,14 +77,22 @@ def place_in_hierarchy(action)
...
           else                                 # or else it should become an additional child
             action
           end
           with(children: [*family[:sibbling], extra_sibbling])
         end
       end
 
+      # Assumes more_children all contained within @range
+      def combine_children(more_children)
+        more_children.inject(self) do |mother, new_child|
```
iliabylich, line 91: let's rename `mother` to `parent` if it makes sense. It took me a few minutes to understand this code because of the unusual naming.

test/test_source_tree_rewriter.rb
```diff
@@ -30,14 +32,23 @@ def apply(actions, **policy)
...
     [::Parser::ClobberingError, diags.call]
   end
 
+  # Returns either:
+  #  - String (Normal operation)
+  #  - [diagnostic, ...] (Diagnostics)
+  #  - Parser::ClobberingError
+  #
+  def apply(actions, **policy)
+    build(actions, **policy, &:process)
```
iliabylich, line 49: nit: could you expand `&:process` please? It's uUnclear without looking at the `build` method who is a receiver of the `.process`

test/test_source_tree_rewriter.rb
```diff
@@ -170,4 +181,70 @@ def test_out_of_range_ranges
...
+  end
+
+  # Checks that `actions+extra` give the same result when
+  # made in order or from subgroups that are later merged.
+  # The `extra` actions are always added at the end of the second group.
+  #
+  def check_all_merge_possibilities(actions, extra, **policy)
+    expected = apply(actions + extra, **policy)
+
+    each_split(actions) do |actions_1, actions_2|
```
iliabylich, line 223: this is error-prone because minitest doesn't complain if there are no assertions in the test. Do you really need to test all combinations? Or it enough to have just a single static test?

Plus, debugging it is a bit complicated if it fails. How many assertions there are? `(N/2)!`?

lib/parser/source/tree_rewriter.rb
marcandre: https://en.wikipedia.org/wiki/If_and_only_if

lib/parser/source/tree_rewriter/action.rb
marcandre: Absolutely.

test/test_source_tree_rewriter.rb
marcandre: You are right.

test/test_source_tree_rewriter.rb
marcandre: There are `(1..8).to_a.combination(4).size # => 70` possibilities. They are probably not all needed, but it seemed less error prone to test them all than to try to think of what could orders could yield errors...
I've added an error message to help debugging.

iliabylich:
@marcandre Thanks. Do you need a release?

marcandre:
Thanks so much for the quick review and constructive comments. I must say it is particularly appreciated tonight 😄 

A release would be appreciated but is not urgent.

Thanks again ❤️

marcandre:
Actually, might as well wait until I make a PR for `Source::Location#eql?`

iliabylich:
> Actually, might as well wait until I make a PR for `Source::Location#eql?`

Sure!

Issue #670, opened 2020-04-10
No Parser::Source::Range#eql?


marcandre:
I hesitated opening this as it's super minor, but is it intentional that there is no `Parser::Source::Range#eql?` (and `#hash`)?

This prevents using `Range`s as meaningful hash keys or `when` conditions.

I noticed this when wondering if [RuboCop should optimize this `O(n^2)` algorithm](https://github.com/rubocop-hq/rubocop/blob/9c3f904e77b3c3786eaface4fdb545764136beca/lib/rubocop/cop/cop.rb#L146) although it would probably be completely negligible in that case compared to the actual policing...

I'd gladly provide a PR if desired.

iliabylich:
Sorry, I missed this issue initially. Yes, it makes sense to me. Feel free to send a RP if you still need it.

iliabylich:
implemented in https://github.com/whitequark/parser/pull/675

PR #675, opened 2020-04-14
+ Add Source::Range#eql? and hash

marcandre:
As discussed in #670


iliabylich:
@marcandre thanks!

iliabylich:
@marcandre  2.7.1.1 has been released

PR #676, opened 2020-04-16
+ ruby28.y: endless method definition

palkan:
TBD

### DIFF
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -808,11 +808,11 @@ Format:
 
 ~~~
 (def_e :foo (args) (int 42))
-"def foo = 42"
+"def foo() = 42"
  ~~~ keyword
      ~~~ name
-         ^ assignment
- ~~~~~~~~~~~~ expression
+           ^ assignment
+ ~~~~~~~~~~~~~~ expression
 ~~~
 
 
@@ -822,11 +822,11 @@ Format:
 
 ~~~
 (defs_e (self) :foo (args) (int 42))
-"def self.foo = 42"
+"def self.foo() = 42"
  ~~~ keyword
           ~~~ name
-              ^ assignment
- ~~~~~~~~~~~~~~~~~ expression
+                ^ assignment
+ ~~~~~~~~~~~~~~~~~~~ expression
 ~~~
 
 ### Undefinition
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -802,6 +802,33 @@ Format:
  ~~~~~~~~~~~~~~~~~ expression
 ~~~
 
+### "Endless" method
+
+Format:
+
+~~~
+(def_e :foo (args) (int 42))
+"def foo = 42"
+ ~~~ keyword
+     ~~~ name
+         ^ assignment
+ ~~~~~~~~~~~~ expression
+~~~
+
+
+### "Endless" singleton method
+
+Format:
+
+~~~
+(defs_e (self) :foo (args) (int 42))
+"def self.foo = 42"
+ ~~~ keyword
+          ~~~ name
+              ^ assignment
+ ~~~~~~~~~~~~~~~~~ expression
+~~~
+
 ### Undefinition
 
 Format:
--- a/lib/parser.rb
+++ b/lib/parser.rb
@@ -46,6 +46,7 @@ module Parser
     require 'parser/source/map/variable'
     require 'parser/source/map/keyword'
     require 'parser/source/map/definition'
+    require 'parser/source/map/endless_definition'
     require 'parser/source/map/send'
     require 'parser/source/map/index'
     require 'parser/source/map/condition'
--- a/lib/parser/ruby28.y
+++ b/lib/parser/ruby28.y
@@ -310,6 +310,35 @@ rule
                       result = [ val[1], val[2] ]
                     }
 
+        def_name:  fname
+                    {
+                      @static_env.extend_static
+                      @lexer.cmdarg.push(false)
+                      @lexer.cond.push(false)
+                      @current_arg_stack.push(nil)
+
+                      result = val[0]
+                    }
+
+       defn_head: kDEF def_name
+                    {
+                      @context.push(:def)
+
+                      result = [ val[0], val[1] ]
+                    }
+
+       defs_head: kDEF singleton dot_or_colon
+                    {
+                      @lexer.state = :expr_fname
+                    }
+                  def_name
+                    {
+                      @context.push(:defs)
+
+                      result = [ val[0], val[1], val[2], val[4] ]
+                    }
+
+
     command_call: command
                 | block_command
 
@@ -1224,18 +1253,10 @@ rule
                       @lexer.cmdarg.pop
                       @static_env.unextend
                     }
-                | kDEF fname
+                | defn_head f_arglist bodystmt kEND
                     {
-                      @static_env.extend_static
-                      @lexer.cmdarg.push(false)
-                      @lexer.cond.push(false)
-                      @context.push(:def)
-                      @current_arg_stack.push(nil)
-                    }
-                    f_arglist bodystmt kEND
-                    {
-                      result = @builder.def_method(val[0], val[1],
-                                  val[3], val[4], val[5])
+                      result = @builder.def_method(*val[0], val[1],
+                                  val[2], val[3])
 
                       @lexer.cmdarg.pop
                       @lexer.cond.pop
@@ -1243,22 +1264,10 @@ rule
                       @context.pop
                       @current_arg_stack.pop
                     }
-                | kDEF singleton dot_or_colon
+                | defs_head f_arglist bodystmt kEND
                     {
-                      @lexer.state = :expr_fname
-                    }
-                    fname
-                    {
-                      @static_env.extend_static
-                      @lexer.cmdarg.push(false)
-                      @lexer.cond.push(false)
-                      @context.push(:defs)
-                      @current_arg_stack.push(nil)
-                    }
-                    f_arglist bodystmt kEND
-                    {
-                      result = @builder.def_singleton(val[0], val[1], val[2],
-                                  val[4], val[6], val[7], val[8])
+                      result = @builder.def_singleton(*val[0], val[1],
+                                  val[2], val[3])
 
                       @lexer.cmdarg.pop
                       @lexer.cond.pop
--- a/lib/parser/ruby28.y
+++ b/lib/parser/ruby28.y
@@ -846,7 +846,7 @@ rule
                       result = @builder.ternary(val[0], val[1],
                                                 val[2], val[4], val[5])
                     }
-                | defn_head f_arglist_opt tEQL arg
+                | defn_head f_paren_args tEQL arg
                     {
                       result = @builder.def_endless_method(*val[0],
                                  val[1], val[2], val[3])
@@ -857,7 +857,7 @@ rule
                       @context.pop
                       @current_arg_stack.pop
                     }
-                | defs_head f_arglist_opt tEQL arg
+                | defs_head f_paren_args tEQL arg
                     {
                       result = @builder.def_endless_singleton(*val[0],
                                  val[1], val[2], val[3])
@@ -2546,13 +2546,7 @@ keyword_variable: kNIL
                       result = nil
                     }
 
-   f_arglist_opt: # nothing
-                    {
-                      result = @builder.args(nil, [], nil)
-                    }
-                | f_arglist
-
-       f_arglist: tLPAREN2 f_args rparen
+   f_paren_args: tLPAREN2 f_args rparen
                     {
                       result = @builder.args(val[0], val[1], val[2])
 
@@ -2565,6 +2559,8 @@ keyword_variable: kNIL
 
                       @lexer.state = :expr_value
                     }
+
+       f_arglist: f_paren_args
                 |   {
                       result = @lexer.in_kwarg
                       @lexer.in_kwarg = true
--- a/lib/parser/ruby28.y
+++ b/lib/parser/ruby28.y
@@ -846,6 +846,28 @@ rule
                       result = @builder.ternary(val[0], val[1],
                                                 val[2], val[4], val[5])
                     }
+                | defn_head f_arglist_opt tEQL arg
+                    {
+                      result = @builder.def_endless_method(*val[0],
+                                 val[1], val[2], val[3])
+
+                      @lexer.cmdarg.pop
+                      @lexer.cond.pop
+                      @static_env.unextend
+                      @context.pop
+                      @current_arg_stack.pop
+                    }
+                | defs_head f_arglist_opt tEQL arg
+                    {
+                      result = @builder.def_endless_singleton(*val[0],
+                                 val[1], val[2], val[3])
+
+                      @lexer.cmdarg.pop
+                      @lexer.cond.pop
+                      @static_env.unextend
+                      @context.pop
+                      @current_arg_stack.pop
+                    }
                 | primary
 
            relop: tGT | tLT | tGEQ | tLEQ
@@ -2523,6 +2545,12 @@ keyword_variable: kNIL
                     {
                       result = nil
                     }
+
+   f_arglist_opt: # nothing
+                    {
+                      result = @builder.args(nil, [], nil)
+                    }
+                | f_arglist
 
        f_arglist: tLPAREN2 f_args rparen
                     {
--- a/lib/parser/ast/processor.rb
+++ b/lib/parser/ast/processor.rb
@@ -159,6 +159,8 @@ module Parser
         ])
       end
 
+      alias on_def_e on_def
+
       def on_defs(node)
         definee_node, name, args_node, body_node = *node
 
@@ -167,6 +169,8 @@ module Parser
           process(args_node), process(body_node)
         ])
       end
+
+      alias on_defs_e on_defs
 
       alias on_undef    process_regular_node
       alias on_alias    process_regular_node
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -652,19 +652,28 @@ module Parser
         definition_map(def_t, nil, name_t, end_t))
     end
 
+    def def_endless_method(def_t, name_t, args,
+                           assignment_t, body)
+      n(:def_e, [ value(name_t).to_sym, args, body ],
+        endless_definition_map(def_t, nil, name_t, assignment_t, body))
+    end
+
     def def_singleton(def_t, definee, dot_t,
                       name_t, args,
                       body, end_t)
-      case definee.type
-      when :int, :str, :dstr, :sym, :dsym,
-           :regexp, :array, :hash
+      return unless validate_definee(definee)
 
-        diagnostic :error, :singleton_literal, nil, definee.loc.expression
+      n(:defs, [ definee, value(name_t).to_sym, args, body ],
+        definition_map(def_t, dot_t, name_t, end_t))
+    end
 
-      else
-        n(:defs, [ definee, value(name_t).to_sym, args, body ],
-          definition_map(def_t, dot_t, name_t, end_t))
-      end
+    def def_endless_singleton(def_t, definee, dot_t,
+                              name_t, args,
+                              assignment_t, body)
+      return unless validate_definee(definee)
+
+      n(:defs_e, [ definee, value(name_t).to_sym, args, body ],
+        endless_definition_map(def_t, dot_t, name_t, assignment_t, body))
     end
 
     def undef_method(undef_t, names)
@@ -1743,6 +1752,14 @@ module Parser
                                   loc(end_t))
     end
 
+    def endless_definition_map(keyword_t, operator_t, name_t, assignment_t, body_e)
+      body_l = body_e.loc.expression
+
+      Source::Map::EndlessDefinition.new(loc(keyword_t),
+                                         loc(operator_t), loc(name_t),
+                                         loc(assignment_t), body_l)
+    end
+
     def send_map(receiver_e, dot_t, selector_t, begin_t=nil, args=[], end_t=nil)
       if receiver_e
         begin_l = receiver_e.loc.expression
@@ -1983,6 +2000,18 @@ module Parser
 
       if type == :error
         @parser.send :yyerror
+      end
+    end
+
+    def validate_definee(definee)
+      case definee.type
+      when :int, :str, :dstr, :sym, :dsym,
+           :regexp, :array, :hash
+
+        diagnostic :error, :singleton_literal, nil, definee.loc.expression
+        false
+      else
+        true
       end
     end
   end
--- a/lib/parser/meta.rb
+++ b/lib/parser/meta.rb
@@ -16,7 +16,7 @@ module Parser
         op_asgn and_asgn ensure rescue arg_expr
         or_asgn back_ref nth_ref
         match_with_lvasgn match_current_line
-        module class sclass def defs undef alias args
+        module class sclass def defs def_e defs_e undef alias args
         cbase arg optarg restarg blockarg block_pass kwarg kwoptarg
         kwrestarg kwnilarg send csend super zsuper yield block
         and not or if when case while until while_post
--- a/lib/parser/source/map/endless_definition.rb
+++ b/lib/parser/source/map/endless_definition.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Parser
+  module Source
+
+    class Map::EndlessDefinition < Map
+      attr_reader :keyword
+      attr_reader :operator
+      attr_reader :name
+      attr_reader :assignment
+
+      def initialize(keyword_l, operator_l, name_l, assignment_l, body_l)
+        @keyword  = keyword_l
+        @operator = operator_l
+        @name     = name_l
+        @assignment = assignment_l
+
+        super(@keyword.join(body_l))
+      end
+    end
+
+  end
+end
--- a/test/test_lexer.rb
+++ b/test/test_lexer.rb
@@ -3572,11 +3572,13 @@ class TestLexer < Minitest::Test
   def test_endless_method
     setup_lexer(28)
 
-    assert_scanned('def foo = 42',
+    assert_scanned('def foo() = 42',
                     :kDEF, "def", [0, 3],
-                    :tIDENTIFIER, 'foo', [4, 7],
-                    :tEQL, "=", [8, 9],
-                    :tINTEGER, 42, [10, 12])
+                    :tIDENTIFIER, "foo", [4, 7],
+                    :tLPAREN2, "(", [7, 8],
+                    :tRPAREN, ")", [8, 9],
+                    :tEQL, "=", [10, 11],
+                    :tINTEGER, 42, [12, 14])
   end
 
   def lex_numbered_parameter(input)
--- a/test/test_lexer.rb
+++ b/test/test_lexer.rb
@@ -3569,6 +3569,16 @@ class TestLexer < Minitest::Test
                    :tIDENTIFIER, 're', [1, 3])
   end
 
+  def test_endless_method
+    setup_lexer(28)
+
+    assert_scanned('def foo = 42',
+                    :kDEF, "def", [0, 3],
+                    :tIDENTIFIER, 'foo', [4, 7],
+                    :tEQL, "=", [8, 9],
+                    :tINTEGER, 42, [10, 12])
+  end
+
   def lex_numbered_parameter(input)
     @lex.max_numparam_stack.push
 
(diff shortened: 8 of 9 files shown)


iliabylich:
I think it's too early to start implementing these new features at least because I'm not sure what will be the next MRI version. Are you sure it's going to be 2.8 and not 3.0?

palkan:
> Are you sure it's going to be 2.8 and not 3.0?

Currently, it's named 2.8 🤷🏻‍♂️

I'm going to implement these (controversal) features for Ruby Next as soon as possible 🙂.

The purpose of this PR is to discuss the API to make it possible to merge these features to Parser easily when 3.0 became stable. And to let you know that I plan to work on this)

Probably, it's better to open a PR in https://github.com/ruby-next/parser and link to the issue instead.  WDYT?

iliabylich:
ok, let's keep it as 2.8 for now. We can rename files whenever we want.

Could you split it into 2 PRs please?

lib/parser/ruby28.y
```diff
@@ -2515,6 +2515,24 @@ keyword_variable: kNIL
                       result = nil
                     }
 
+     f_args_body: f_arglist bodystmt kEND
```
palkan, line 2518: @iliabylich Tried to implement it in a straightforward way (mostly by porting [this commit](https://github.com/ruby/ruby/commit/e8f53692ca7d65c92bde6d5bc6a1aea492915d9f) from Ruby). Got stuck with the "4 reduce/reduce conflicts" in the grammar 😞

The problem with `arg` terminal. Not sure what is the proper way to resolve it. Could you please give some hint?

doc/AST_FORMAT.md
```diff
@@ -802,6 +802,31 @@ Format:
  ~~~~~~~~~~~~~~~~~ expression
 ~~~
 
+### "Endless" method
+
+Format:
+
+~~~
+(def_e :foo (args) (int 42))
```
palkan, line 810: Not sure about the naming. Any suggestions?

lib/parser/ruby28.y
iliabylich: a few hints:
1. don't rename rules. I don't see `opt_f_arglist` in https://github.com/ruby/ruby/commit/e8f53692ca7d65c92bde6d5bc6a1aea492915d9f
2. copy all changes related to grammar. partially backported changes may contain conflicts
3. if you are curious and want to dig deeper (which is usually unnecessary) you can pass `--debug -O .racc_output` to `racc` and look into `.racc_output`. It has information about what should be on the stack to get a conflict.

doc/AST_FORMAT.md
iliabylich: I like it. Also could've been called something like "def one-liner", but as I understand you still can use `\` to break into multiple lines, so technically there's no such requirement.

Also, we need to emit some information about the `=` token.

lib/parser/source/map/definition.rb
```diff
@@ -9,13 +9,13 @@ class Map::Definition < Map
       attr_reader :name
       attr_reader :end
 
-      def initialize(keyword_l, operator_l, name_l, end_l)
+      def initialize(keyword_l, operator_l, name_l, body_l, end_l)
```
iliabylich, line 12: not sure if it's worth making this class so generic. Maybe let's create a new class? This way you could give it a very specific attribute for `=` character location.

doc/AST_FORMAT.md
palkan: I was thinking about these new nodes again, and now I'm not sure we need them at all: what about re-using `def` and `defs`? The absence of the `end` could indicate the endless-ness of the definition.

As well as the presence of the `=` (not sure how to call it, `assignment`?) attribute.

```
~~~
(def :foo (args) (int 42))
"def foo = 42"
 ~~~ keyword
     ~~~ name
         ^ assignment
 ~~~~~~~~~~~~ expression
~~~
```

iliabylich: that would break all the code expecting `end` to be non-nullable. I'm sure some libraries don't have `if` checks (because these checks don't make any sense atm), so it'd be a breaking change.

palkan: Good point. Thanks! 

What about the `assignment` name for the `=` attribute?

lib/parser/ruby28.y
palkan: Thanks for the advice! It helped 🙂

Made another attempt, and now it works!

lib/parser.rb
```diff
@@ -46,6 +46,7 @@ module Source
     require 'parser/source/map/variable'
     require 'parser/source/map/keyword'
     require 'parser/source/map/definition'
+    require 'parser/source/map/endless_definition'
```
iliabylich, line 49: I actually think that someday MRI developers may introduce endless classes/methods, so maybe we'll reuse it:
```rb
class A = def m = 42
module M = def m = 42
```

iliabylich:
@palkan feel free to merge it

iliabylich:
Thanks!

marcandre:
I really dislike that we're emitting a new node type for something that is not different in any way shape or form, only in it's syntax. The fact that it's written differently doesn't change that it is simply a `def`.

I personally do not consider a null location for a new Ruby syntax an actual breaking change and that emitting a new node type is more breaking, but if that's not the official stance, could we at least have an `emit_` setting?

I hope this will be reverted and resolve the issue.

iliabylich:
@marcandre I totally understand your point. There's only one reason why it's a new node: backward compatibility. By emitting this new syntax as `s(:def)` we would break all libraries that expect `def_node.end` to always be non-nullable. I'm pretty sure that some of them expect `def` node to always have `end` location.

For example, as I know rubocop ignores nodes that are not listed in its own readonly visitor, so if I have `def_e` node in my code it should be simply ignored. But if this new node is `def_e` I'd get an NPE and the whole file would be invalid.

What does it complicate on your side? If rubocop has all `if` checks for such cases you could try overriding [method `def_endless_method`](https://github.com/whitequark/parser/blob/master/lib/parser/builders/default.rb#L696) to emit `def` (something like `super.updated(:def, nil)`). Does it work for you?

marcandre:
>  I'm pretty sure that some of them expect def node to always have end location.

Possible. I'm pretty sure that some of them expect all def node to be `def` or `defs` (probably not as popular as rubocop though).

We'll have to agree to disagree that language syntax changes in major Ruby releases can not change anything to already emitted node types and that they all require a new node type, even if semantically the same.

Is that definition of "compatibility" actually stated anywhere in the docs?

> What does it complicate on your side?

It's simply feels *wrong* to emit a different node type. `a ? b : c` and `if a then b else c` are emitted as the same `s(:if`, so should `def foo` and `def: foo`. 

Basically: maintain compatibility with an `emit_` switch if you feel you have to, but I feel that `parser` with modernize options should output the *correct* meaningful ast.

iliabylich:
> Is that definition of "compatibility" actually stated anywhere in the docs?

No, but that's what you'd expect from any library, right? Parser doesn't follow semver, so we are not allowed to do any breaking changes because we have no major version bumps at all.

> `parser` with modernize options should output the *correct* meaningful ast.

I think it is correct and it is meaningful, but I'd say it's too verbose for your case.

If someone needs to rewrite `def_e` and `defs_e` nodes (like @palkan in his library) it can be explicitly written by looking at specific node types, not at the `end` location of generic `def`/`defs` nodes. For some use cases this level of verbosity works better, for some cases it only causes troubles.

marcandre:
> I think it is correct and it is meaningful

Could you expand on why you think this? What makes `def` and `def_e` semantically different types of nodes?

I'd be curious to know what @whitequark thinks of this

iliabylich:
> Could you expand on why you think this?

Well, maybe it's not a direct explanation, but I think that just AST (without source maps) should be enough to construct the same source code. Source maps are just a secondary data that gives you information about **how** the code is formatted, while AST represents **what** the code does.

Current format is correct because it gives **enough** data to do it. By merging `def` and `def_e` we lose this ability as we don't know which kind of method definition was initially used.

You had a very good example with ternary operator - I'm not sure if I'd emit it as a regular `if` node if I would implement it today. For example, IIRC babel/parser (for JS) also emits them as separate node types.

marcandre:
> Well, maybe it's not a direct explanation, but I think that just AST (without source maps) should be enough to construct the same source code. 

If by "same" you mean character by character, then I completely disagree with you and you clearly can't do it for the majority of nodes; good luck with `? :`, `unless`, `if`, but more simply just with `def`: use parenthesis or not for the argument list? how about spacing? This distinction is made by the https://github.com/mbj/unparser gem.

If by "same" you mean that they produce exact replica of the code will always behave entirely the same way under all circumstances, then I completely agree with you. That is exactly my point actually. I challenge you to find an example where I won't be able to produce equivalent code (without checking the actual source or `location` map) if `def_e` were emitted as `def`. All `def_e` can be written as 100% equivalent `def` and vice versa. `unparser` does not need a `def_e`. It's 100% pure syntactic sugar and therefore should not be treated as a different node type.

iliabylich:
You are right, all `def_e` nodes could be emitted as `def` nodes.

I didn't mean that current AST format is the-most-optimal and there are no nodes that could be replaced with equivalents. Of course we can get rid of things like `for` loop (by replacing it with `.each {}` just like MRI does under the hood) and `case` statement (by replacing it with a chain of `if`s with `===`). I guess my initial explanation was wrong from this POV. AST is definitely not only about un-parsing. I think it's important to give information about all kinds of nodes if possible, but you right, it's unclear when it should be a new node and when it's OK to just change source maps (if there's no difference between them from runtime perspective).

It's hard to make this decision (if possible at all) because it's completely unclear (at least for me) how Ruby grammar is going to change. Will there be any grammar features available only for one-line methods? Ruby has no RFCs and we have to put some assumptions like the one that was made in this PR.

What should we do if such new syntax will be introduce some day? Another compatibility flag? I don't like them and I don't want them to grow every time we change the format. For me it's a trade off between keeping AST format way too simple and adding flags every time we want to change it vs keeping it big and verbose, but forcing users to support more node types.

marcandre:
> Of course we can get rid of things like `for` loop (by replacing it with `.each {}` just like MRI does under the hood) 

I believe this is incorrect. Scoping is different. `for i in 1..10; end; p i` vs `(1..10).each { |i| }; p i` are different

> and `case` statement (by replacing it with a chain of `if`s with `===`).

Also incorrect (at least with `case expr`). You would have to create a local variable to hold the `expr`, which has a detectable impact on `binding` and would hold a reference to the object a different length of time.

I wrote "always behave entirely the same way under all circumstances" and I meant it. If they're the same under 99.99% of the circumstances, they are not entirely the same. 

> I guess my initial explanation was wrong from this POV.

I think your initial explanation was actually completely correct, you shouldn't discount your (actually correct) intuitive understanding because it doesn't yield the (actually incorrect) result you think it should 😄 

> What should we do if such new syntax will be introduce some day? Another compatibility flag? 

I agree completely that compatibility flags should be minimized.

Abstractly, `parser` is a set of mappings from all possible Ruby code to equivalent ASTs, one mapping for each supported Ruby versions.

To me, compatibility means that for all for any subset of those mappings, ruby code that are equivalent across those Ruby versions should have the same corresponding AST and that this AST never changes. I'm excluding ruby-head (2.8 here) as it can't offer compatibility until 2.8 is actually released.

If my Ruby 2.6 code produced a particular AST with `parser`, I expect that AST to not change (except for bug fixes). I also expect to get the same AST if using `Parser::Ruby27` and all future versions (assuming that the particular ruby code is still interpreted by Ruby the same way)

When support for Ruby 2.8 is added, running with `Parser::Ruby28` on the same 2.6-compatible code should produce the same AST. The possible resulting ASTs from `Parser::Ruby28`, *when considering Ruby 2.6 compatible code*, are the same as those of `Parser::Ruby26` and `Ruby27`. But running it on 2.8 code that is *not* equivalent or compatible with Ruby 2.6/2.7 should have no compatibility guarantee besides being an `AST::Node`.

It's clear to me that syntax changes in Ruby will sometimes change just the location (as it should here), sometimes introduce different node types and sometimes change expectations on children. As long as that doesn't impact older Ruby code, that is fine.

New node types should be reserved when the Ruby construct is actually introducing semantically new concepts (e.g. `...` which can't quite completely be replaced by other forwarding mechanisms, sadly).

whitequark:
If I recall correctly, then a single node type (like `def`) always has the same source map subclass, so that you can check which non-semantic variant you got by looking at which fields are `nil?`, not by checking `is_a? Parser::Source::Map::Whatever` which is an implementation detail. So if you introduce a new source map subclass with different fields it's a breaking change for the downstream tooling.

Another way to think about it is: suppose I write a tool that looks at the `end`s of methods. It expects every `def` to have one. If the tool breaks with updated Parser then this is of course the fault of `parser` and a violation of its backwards compatibility promise. So I think @iliabylich did the right thing here.

@marcandre 

> Abstractly, `parser` is a set of mappings from all possible Ruby code to equivalent ASTs, one mapping for each supported Ruby versions.

I didn't use any absolute statement like that as guidance when developing Parser. What I considered was, is it actually convenient for downstream tooling? Does it create cases where downstream tooling would have to do weird things to disambiguate? It shouldn't. AFAICT folding `def_e` into `def` would require downstream tooling to do weird things so I think it's a bad idea to do so.

marcandre:
Thanks for the reply.

Looks like I'm not convincing anybody (and I'm not getting convinced). That's ok 😄 

I have few more questions below, feel free to answer only those that you find interesting (if any). I've simplified by ignoring `defs` / `defs_e`.

1) We keep talking "backward compatiblity", but it's really ["forward compatibility"](https://en.wikipedia.org/wiki/Forward_compatibility), right?

2) The only reason I hear is that `def_e` is for compatiblity. So if this was the first ever release of `parser`, or if that syntax had been in Ruby since 1.8, do we agree that it would only emit `def`? So `parser` will emit "less than optimal" AST for compatibility? What are the other examples where `parser` does not emit the "optimal" AST for forward compatibility reasons?

3) @whitequark: do you regret having a single node type for `if`, `unless` and `?:`? I recall finding it very surprising. Now I find it ingenious.

4) If Ruby had introduced `%w[string other_string]` only today, would we create a new `str_something` to avoid having a `str` with no `begin` and `end` locations?

5) Why is having `irange` / `erange` accepting a `nil` child  (endless/beginless ranges) considered more forward compatible than having a `nil` end location? (For the record, I completely agree it was the right decision, but it seems like a less compatible change to me)

6) I probably should have started with this, but as far as compatiblity goes, introducing a `def_e` crashes all existing versions of `RuboCop` (because `Commissionner` expects knowing all node types, but that could be remedied). Even if that was/had been remedied (for example that `Commissionner` forwarded unknown calls to cops that respond to them, as it probably should), it would still break all cops relying on `VariableForce` (which analyses scopes of variables for them), all cops implementing `on_def` (and not `on_def_e`), all `NodePatterns` with a `"def"` in there, and crash the gem `unparser` (I imagine) and `DeepCover` (definitely).

By contrast, with a `def` with no end location, `RuboCop` works (all versions), `VariableForce` works, most cops implementing `on_def` too (*), all `NodePatterns` with a `"def"` work, the gem `unparser` and `DeepCover` work out of the box.

(*) I haven't verified which builtin cops rely on `location.end` for `def` nodes, but my assumption is that few do. `on_def` is pretty important as a scope, but I presume that the end location is important to very few cops; Clearly the Cop that checks the indentation of `def` vs `end` will break :-)
 Few cops will need to rewrite the whole `def`, or insert things begfore/after the `end`, and since `expression` is still valid they might be relying on that. 

Which known tools will work out of the box as designed with `def_e`? Any of those wouldn't work with `def` with no end location?

whitequark:
> 1\. We keep talking "backward compatiblity", but it's really ["forward compatibility"](https://en.wikipedia.org/wiki/Forward_compatibility), right?

I think what you call it depends on the point of view. I'd probably call it "AST stability" if I tried to be more clear.

> 2\. The only reason I hear is that `def_e` is for compatiblity. So if this was the first ever release of `parser`, or if that syntax had been in Ruby since 1.8, do we agree that it would only emit `def`? So `parser` will emit "less than optimal" AST for compatibility?

Correct.

> 3\. @whitequark: do you regret having a single node type for `if`, `unless` and `?:`? I recall finding it very surprising. Now I find it ingenious.

Nope, it was a deliberate design decision. Parser doesn't normalize everything (there's `kwbegin` for example) but it normalizes a great deal to make life easier for AST consumers.

> 4\. If Ruby had introduced `%w[string other_string]` only today, would we create a new `str_something` to avoid having a `str` with no `begin` and `end` locations?

Probably, yeah.

> 5\. Why is having `irange` / `erange` accepting a `nil` child  (endless/beginless ranges) considered more forward compatible than having a `nil` end location?

I'm no longer involved in day-to-day decisions, so ask @iliabylich. Arguably the decisions here should have been the same.

> 6\. I probably should have started with this, but as far as compatiblity goes, introducing a `def_e` crashes all existing versions of `RuboCop`

Okay, so adding `def_e` requires downstream tooling to do weird things after all. This means I was mistaken. I will follow my own principle and ask @iliabylich to revert this change and implement @marcandre's suggestion.

iliabylich:
@whitequark Now it's my turn to say "I don't follow".

Let's imagine that `def_e` is implemented as `def` with empty `end` location. Are there any guarantees that libraries that use parser implement all required support (i.e. they add `if` checks in appropriate places) before the next version of parser is released? That's great if so.

But what if there's no support? End users (of rubocop, let's say) start using `def foo: value` syntax and get errors.

On the other side, if it's implemented as a new `def_e` node it would be simply ignored (because there's no code that goes into these nodes). Which one is better? I always thought it's the latter.

Also, it makes old versions of rubocop incompatible with new parser, i.e. if users do `bundle update parser` the get same errors coming from the new syntax.

IMO endless (and beginningless) ranges [should've been implemented as new nodes too](https://github.com/rubocop-hq/rubocop/issues/6654), I think it was a mistake from me.

Who is the main consumer? End users or 3rd party libraries that use parser?

marcandre:
> Let's imagine that `def_e` is implemented as `def` with empty end location. Are there any guarantees that libraries that use parser implement all required support (i.e. they add if checks in appropriate places) before the next version of parser is released? That's great if so.

For many cases, there are no checks to make because there's rarely a need to check all the locations. What is that library doing that requires it to use the `end` location? Is there an example?

> But what if there's no support? End users (of rubocop, let's say) start using def foo: value syntax and get errors.

I'm not sure I understand. As I explained end users of current RuboCop will systematically get a crash with `def_e` even if running a single Cop (although this could be mitigated)

> On the other side, if it's implemented as a new def_e node it would be simply ignored (because there's no code that goes into these nodes).

Currently this is not true for `RuboCop`, will never be true for `DeepCover`, and I presume cannot be true for `unparser`.

> Also, it makes old versions of rubocop incompatible with new parser, i.e. if users do bundle update parser the get same errors coming from the new syntax.

With no `def_e`, it will be only for some cops (which can be disabled), and RuboCop provides nice error messages in those cases. With `def_e` you get a crash.

> IMO endless (and beginningless) ranges should've been implemented as new nodes too, I think it was a mistake from me.

As new nodes they would also have crashed RuboCop (but that could be mitigated).

As far as RuboCop is concerned, I'm 100% ok with RuboCop not running on previously unknown versions of Ruby. My main concern is to minimize the work involved with the compatibility of the very many cops existing (builtin or not). For this, I feel that an AST that is sticks close to our intuitive understanding is best. That's why I consider the AST (`type` + `children` = what is it) and `location` to be very secondary (`where is it written`).

palkan:
Hey, everyone!

I'm a bit late but as the original initiator and the author of this feature, I feel that I should jump into a discussion.

I also [proposed](https://github.com/whitequark/parser/pull/676#discussion_r412039130) using the absence of the `end` location as an endless method definition indicator; @iliabylich provided me with the compatibility argument and I agreed with him. And I still agree.

One thing that haven't been mentioned during the recent discussion, is another difference between `def` and `def_e`: endless def-s also have an assignment "operator" (`=`), and we have the corresponding `assignment` location in the new source map class. And endless method is syntactically different to a regular one not only due to the absence of the `end` keyword but by the presence of `=` which is not possible for regular methods.

>  introducing a def_e crashes all existing versions of RuboCop (because Commissionner expects knowing all node types, but that could be remedied)

IMO, that should be fixed at the RuboCop's side.

Generally speaking, I think, crashes due to new node types is a downstream software problem. New nodes are added regularly (especially in the recent years, Ruby syntax is evolving fast). Downstream software should be prepared for that, it should be able to handle new nodes gracefully.

On the other hand, when changing the existing nodes breaks downstream tools—that's more a Parser problem. For example, the already mentioned re-usage of `irange / erange` for endless ranges broke Unparser: https://github.com/mbj/unparser/pull/137 (adding a new node would also result in a crash but with a more clear exception: unsupported node type).

Anyway, I like the idea that @whitequark described:

 >  is it actually convenient for downstream tooling? Does it create cases where downstream tooling would have to do weird things to disambiguate?

That led me to the following idea: what if we come up with some kind of an informal (or even formal) process of modifying `AST_FORMAT.md`? For example, we can request reviews from major downstream software maintainers (to have a discussion like this before the feature is merged).

An alternative idea is to provide canary/edge/whatever releases with the features for the edge Ruby versions. That would make it possible to update the format without worrying about the compatibility while it's not a part of a stable release.

whitequark:
I have another proposal that would hopefully be a good compromise between the two we have now: having downstream clients opt into node types they support. For example, if old Rubocop doesn't support `def_e`, it would not opt into getting it, and then *Parser itself* would emit a nice diagnostic instead of emiting a `def_e` node and crashing the client.

whitequark:
For that matter you don't have to make the opt-in at node type granularity. You could opt into getting ranges with `nil`s in them just as well.

marcandre:
> On the other hand, when changing the existing nodes breaks downstream tools—that's more a Parser problem. For example, the already mentioned re-usage of `irange / erange` for endless ranges broke Unparser: [mbj/unparser#137](https://github.com/mbj/unparser/pull/137) (adding a new node would also result in a crash but with a more clear exception: unsupported node type).

My opinion is that we should not compromise on the AST we produce so that tools can try to be *forward compatible*. For endless node ranges, `unparser` broke. There was no other alternative anyways. Maybe my opinion is not shared, but that is not a problem.

FWIW, `DeepCover` [broke too](https://github.com/deep-cover/deep-cover/commit/4133d5fefbba62e3a58c3a674c9190bc9dd767db) mainly because we are super strict on the AST we process. Every single AST type has to be explicitly vetted, all its arguments (single or variable) is named and the possible types of these arguments is also specified:

```ruby
class Range
-  has_child from: Node
-  has_child to: Node
+  has_child from: [Node, nil]
+  has_child to: [Node, nil]
```
OTOH, `DeepCover` makes as few assumptions on location as possible. I originally thought that `location` was a kind of `OpenStruct`. Maybe I have the wrong mental image.

> That led me to the following idea: what if we come up with some kind of an informal (or even formal) process of modifying `AST_FORMAT.md`?

I'm very impressed by @iliabylich's dilligence in implementing all the syntax changes, and I should be more attentive. I'm fine to be pinged for these changes, but I am expecting to agree with @iliabylich the vast majority of times and I wouldn't want @iliabylich to have to wait for a lot of approvals, as long as there is openness to revisit a decision like in this case, in the months before the new syntax changes are released at X-mas.

marcandre:
> I have another proposal that would hopefully be a good compromise between the two we have now: having downstream clients opt into node types they support.

This seems like a lot of work for most clients (trivial for `DeepCover`, unless you want to include the locations in there in which case a lot of work) and a super complex API too. 

If you want to go this way, downstream clients could simply specify which Ruby version they are aware of. There's already a crude way for them to check if they are a bit out of their depth though, which is to not use `parser/current` 😅, or checking the result against a set of known parsers...

whitequark:
> This seems like a lot of work for most clients (trivial for `DeepCover`, unless you want to include the locations in there in which case a lot of work) and a super complex API too.

If you reject that, the question becomes: do we break the clients with (more) new nodes, or do we break the clients with (more) changes to existing nodes?

marcandre:
> If you reject that

I don't quite reject it. I say that the list of nodes a client is opting-in, with the possible children and locations doesn't need to be spelled out in full as it corresponds *exactly* to a given Ruby version.

As far as I'm concerned, a tool that is unaware of `Parser26` and uses it nevertheless might break when using it on code that `Parser25` could not process.  The tool is free to use `Parser25` (in which case `parser` will generate a nice syntax error on code using the new syntax), or try with `Parser26` and maybe it'll work, but probably not.

> do we break the clients with (more) new nodes, or do we break the clients with (more) changes to existing nodes?

My opinion is that the changes should reflect in the best way the actual Ruby syntax change. The closer we are to the semantics of the new Ruby code, the less difficult it will be for the clients to adapt, since what they are trying to do is directly related to what the code means.

As a consequence, changes that affect only the `locations` are not semantic and should be preferred to changes of AST children / types (like here). 

I sense that there might be a feeling that introducing a new type is the best way to avoid incompatibilities or is to be more expected than other changes but I believe it is the contrary. It doesn't actually help. Unless it's a different concept, there should not be a different AST type. Having `irange` with a `nil` child was the right decision because of that; it's not a new concept deserving of a new node type. Forward args (sadly) is a new concept and it is definitely out of the question to use `restarg` and say check the source 😅

whitequark:
> I say that the list of nodes a client is opting-in, with the possible children and locations doesn't need to be spelled out in full

I wasn't talking about that. I was talking about a mechanism similar to the existing one we have, e.g. `emit_def_e` like `emit_forward_arg`.

> As a consequence, changes that affect only the `locations` are not semantic and should be preferred to changes of AST children / types (like here).

Changes that affect locations *absolutely* are semantic, since the *whole fucking point* of me developing Parser in first place is to provide source locations. If you don't care about locations just use Ripper.

marcandre:
> I wasn't talking about that. I was talking about a mechanism similar to the existing one we have, e.g. `emit_def_e` like `emit_forward_arg`.

Oh, I stated before that I'm fine with `emit_def_e` if it's deemed the best solution.

>  the whole fucking point of me developing Parser in first place is to provide source locations.

I hope your use of language isn't a sign of exasperation. I only wish for an interesting discussion and the best `parser` gem possible. I will respect your decision and @iliabylich's, and even if it includes deciding to take a break or close the discussion if it becomes tedious to either of you.

> If you don't care about locations just use Ripper

I imagine you know, but in case you don't, Ripper produces a particularly horrible, unusable, unsemantic and even broken AST:

`foo(:bar)` vs `foo :bar` => different Sexp
`%i[foo]` vs `[:foo]` => different Sexp
`%i[foo]` vs `%w[foo]` => exactly the same Sexp!?!

It's actually not an [AST](https://en.wikipedia.org/wiki/Abstract_syntax) but a [parse tree](https://en.wikipedia.org/wiki/Parse_tree), and a broken one at that...

whitequark:
> I hope your use of language isn't a sign of exasperation.

Sorry, I had an awful day. It doesn't have much to do with this thread, and I should have just replied later.

Let me rephrase. The reason I started Parser when ruby_parser and Ripper existed was to have precise source locations. So for me, of course those are semantic--Parser is intended, and was written for, tooling that requires those!

marcandre:
> So for me, of course those are semantic--Parser is intended, and was written for, tooling that requires those!

By semantic, I meant "related to the *meaning* of the code". It's in that sense that I don't consider the locations to be semantic. `def foo = bar` and `def foo; bar; end`, at least in my mind, *mean* exactly the same thing, even if the syntax used is different. There is no doubt that locations can be quite helpful and together with a semantic AST make `parser` a fabulous gem ❤️ 

@bbatsov I hope things are going better for you today 😄

So where are we at? Possibilities are:
1) status quo
2) emit `def` with `nil` location while allowing users to define `def_endless_method` to emit a different AST if they prefer that, which is #716 
3) introduce an `emit_def_e` or similar that outputs `def` or `def_e`, with a default that remains to be determined.

I'll be happy with 2 or 3 (with either defaults) and I'll definitely survive with 1.
@iliabylich: thanks for #716. Do you prefer this to option 3 above? @palkan what about you?

iliabylich:
I'm not the best person to ask because I don't use parser (and if I'd use it starting from today there would be no difference). All 3 options are fine for me.

Also I'm actually slightly confused that new nodes cause exceptions. `Parser::AST::Processor` is the way to do traversing and it doesn't crash when there's an unknown node (`ruby-parse -L` uses it). rubocop, unparser and ruby-next - I thought all 3 easily handle new nodes by just ignoring them. rubocop can skip unknown nodes (and do no rewriting inside them), unparser/ruby-next can print `node.loc.expression.source`. At least it's better than runtime error IMO, but that's a completely separate topic to discuss. I was wrong initially as I thought that new nodes is the way to "save" other libraries from errors. Well, I guess I was wrong.

palkan:
Although ruby-next already uses `def_e`, I don't think a lot of projects in the wild rely on edge syntax, so I can take care of the change in the next patch release.

I'm OK with any option, maybe, except from the 3rd one: introducing a flag is more like postponing the final decision.

PR #680, opened 2020-04-25
- ruby27.y: reject invalid lvar in pattern matching

palkan:
- [x] Added lvar name check for all `match_var` (https://github.com/ruby/ruby/blob/501f2c44e6ae79c02a5c4d0f872fc7fa77258fcf/test/ruby/test_pattern_matching.rb#L1032-L1036)

- [x] Fixed strings interpolation check (https://github.com/ruby/ruby/blob/501f2c44e6ae79c02a5c4d0f872fc7fa77258fcf/test/ruby/test_pattern_matching.rb#L1092-L1097)

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -1239,8 +1239,10 @@ module Parser
 
     def match_var(name_t)
       name = value(name_t).to_sym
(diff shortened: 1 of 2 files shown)


iliabylich:
Thank you!

PR #681, opened 2020-04-25
rectify multiple assignment example

rahul404:
The commit corrects an example in [AST format document](https://github.com/whitequark/parser/blob/master/doc/AST_FORMAT.md) in the "Multiple Assignment" section.


iliabylich:
Thanks!

PR #683, opened 2020-04-29
Rewriter optimizations

marcandre:
This PR contains two independent optimizations for the `Source::TreeRewriter`.

The first one changes building the tree from `O(n^2)` to `O(n log(n))` where `n` is the number of rewriting actions.

The second one is of lesser importance and changes applying the tree from `O(n * s)` to `O(n + s)`, where `s` is the length of the source.

While I doubt users of `RuboCop` will notice the speedup but this will make a big difference for the conversion and annotation of source files into HTML by `DeepCover` which was ridiculously slow for long source files.

### DIFF
--- a/lib/parser/source/tree_rewriter.rb
+++ b/lib/parser/source/tree_rewriter.rb
@@ -221,20 +221,17 @@ module Parser
       #
       # @return [String]
       #
-      def process
-        source     = @source_buffer.source.dup
-        adjustment = 0
+     def process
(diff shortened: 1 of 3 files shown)


marcandre:
PS: A release when this is merged would be appreciated 😸 . Thanks

iliabylich:
@marcandre `2.7.1.2` has been released.

marcandre:
Thank you 😸

Issue #684, opened 2020-05-04
parser emitting warnings


mperham:
Can I request turning on warnings?  My test suite emits these warnings from parser:

/Users/mikeperham/.gem/ruby/2.7.1/gems/parser-2.7.1.2/lib/parser/source/tree_rewriter.rb:235: warning: mismatched indentations at 'end' with 'def' at 224
/Users/mikeperham/.gem/ruby/2.7.1/gems/parser-2.7.1.2/lib/parser/lexer.rb:10991: warning: assigned but unused variable - testEof

https://github.com/whitequark/parser/blob/master/Rakefile#L14

iliabylich:
The first warning definitely can be fixed. The second one comes from the generated code that we don't own.

I guess we could easily fix it by adding something like 
```ruby
if false
  testEof
end
```
At the bottom of `Lexer#advance`. It will have no runtime cost because Ruby starting from 2.5 has some basic DCE (at least for `if false/unless true`):
```
> ruby --dump=insns -e 'a = 1'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,5)> (catch: FALSE)
local table (size: 1, argc: 0 [opts: 0, rest: -1, post: 0, block: -1, kw: -1@-1, kwrest: -1])
[ 1] a@0
0000 putobject_INT2FIX_1_                                             (   1)[Li]
0001 dup
0002 setlocal_WC_0                a@0
0004 leave

> ruby --dump=insns -e 'a = 1; if false; a; end'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,23)> (catch: FALSE)
local table (size: 1, argc: 0 [opts: 0, rest: -1, post: 0, block: -1, kw: -1@-1, kwrest: -1])
[ 1] a@0
0000 putobject_INT2FIX_1_                                             (   1)[Li]
0001 setlocal_WC_0                a@0
0003 putnil
0004 leave
```

^ so there's an extra `putnil` that will go away because it will not be an implicit return value. Looks like a safe solution to me. I'll send a PR tomorrow.

PR #685, opened 2020-05-05
* fixed all warnings. tests are running in verbose mode now.

iliabylich:
1. added `if false; testEof; end` at the bottom of `Lexer#advance` to prevent warnings about unused variable `testEof`. It has 0 runtime cost.
2. fixed formatting in a few places
3. changed test task definition to run tests in verbose mode (like `ruby -w` would do)
4. For some reason `"" % {}` gives a warning (same with `Kernel#sprintf`, no idea why). Added a utility method to perform interpolation on messages only if there's something to interpolate.
5. added parentheses to all `m /regex/` calls
6. added `_` prefix to unused variables.

I'm getting 0 warnings locally when running `rake test`/`rake test_cov`

Closes https://github.com/whitequark/parser/issues/684

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -284,7 +284,7 @@ class Parser::Lexer
     # %
 
     # Ragel creates a local variable called `testEof` but it doesn't use
-    # it any assignment. This dead code is here to swallow the warning.
+    # it in any assignment. This dead code is here to swallow the warning.
     # It has no runtime cost because Ruby doesn't produce any instructions from it.
     if false
       testEof
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -283,6 +283,13 @@ class Parser::Lexer
     %% write exec;
     # %
 
+    # Ragel creates a local variable called `testEof` but it doesn't use
+    # it any assignment. This dead code is here to swallow the warning.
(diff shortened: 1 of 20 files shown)


lib/parser/messages.rb
```diff
@@ -93,4 +93,19 @@ module Parser
...
+    # Formats the message, returns a raw template if there's nothing to interpolate
+    #
+    # Code like `"" % {}` gives a warning, and so this method tries interpolating
+    # only if `arguments` hash is not empty.
+    #
+    # @api private
+    def self.format(reason, arguments)
+      template = MESSAGES[reason]
+      return template if Hash === arguments && arguments.empty?
+      template % arguments
```
bbatsov, line 108: I'd use `format` here. :-)

lib/parser/lexer.rl
```diff
@@ -283,6 +283,13 @@ class Parser::Lexer
     %% write exec;
     # %
 
+    # Ragel creates a local variable called `testEof` but it doesn't use
+    # it in any assignment. This dead code is here to swallow the warning.
+    # It has no runtime cost because Ruby doesn't produce any instructions from it.
```
mbj, line 288: Until ruby starts warning about `if false` as a redundant construct ;)

PR #686, opened 2020-05-06
Add argument_error to parser messages hash

bwsewell:
Essentially a duplicate of the `:invalid_regexp` message, yielding the raw message text with no accompanying text. Allows to creating a Diagnostic for an ArgumentError.

```
ArgumentError: wrong number of arguments (given 0, expected 1)
```

```ruby
begin
  # ...
rescue ArgumentError => e
  Parser::Diagnostic.new(
    :error,
    :argument_error,
    {
      message: e.message,
    },
    node.location.expression,
  )
end
```

### DIFF
--- a/lib/parser/messages.rb
+++ b/lib/parser/messages.rb
@@ -45,6 +45,7 @@ module Parser
     :invalid_assignment           => 'cannot assign to a keyword',
     :module_name_const            => 'class or module name must be a constant literal',
     :unexpected_token             => 'unexpected token %{token}',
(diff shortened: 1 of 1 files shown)


iliabylich:
Is this PR complete? These messages have to be used somewhere in the code.

If you want to introduce some custom messages for 3rd party library (for example, in a custom builder) you can do it there. There's no need to have a set of all possible errors that 3rd party libs may want.

bwsewell:
@iliabylich gotcha. They are not. Will close this fella.

PR #688, opened 2020-05-15
- fix Comment.associate for postfix conditions/loops

marcandre:
and also any node with children in inverted order.

See test file for example.

An faster alternative solution would be to instead figure out which nodes might have such a situation and visit them accordingly (are there any other than postfix `if/unless/while`?), but this slower solution seemed the most robust.

### DIFF
--- a/lib/parser/source/comment/associator.rb
+++ b/lib/parser/source/comment/associator.rb
@@ -107,6 +107,19 @@ module Parser
 
       private
 
+      POSTFIX_TYPES = Set[:if, :while, :while_post, :until, :until_post].freeze
+      def children_in_source_order(node)
+        if POSTFIX_TYPES.include?(node.type)
+          # All these types have either nodes with expressions, or `nil`
+          # so a compact will do, but they need to be sorted.
+          node.children.compact.sort_by { |child| child.loc.expression.begin_pos }
+        else
+          node.children.select do |child|
+            child.is_a?(AST::Node) && child.loc && child.loc.expression
+          end
+        end
+      end
+
       def do_associate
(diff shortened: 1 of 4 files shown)


marcandre:
For reference: the comment used to be associated to the `(send nil :some_condition)` node.

marcandre:
Now that I think of it, the tests suite of parser itself can be used to find those cases and make for a more efficient implementation. Stay tuned for an improved PR :-)

marcandre:
I've amended the PR to be more efficient. It uses the full test suite to double check that the efficiency isn't at the expense of correctness. I added a method `Node#children_in_source_order`; let me know if you'd prefer it to be hidden somewhere in the `CommentAssociator`.

lib/parser/ast/node.rb
```diff
@@ -34,6 +34,23 @@ def assign_properties(properties)
           @location = location
         end
       end
+
+      POSTFIX_TYPES = Set[:if, :while, :while_post, :until, :until_post].freeze
+      ##
+      # Returns children with an expression, in the order they appear in
+      # the source
+      #
+      def children_in_source_order
```
iliabylich, line 43: could you mark it as `@private` in YARD please? I don't think anyone except `parser` needs it.

marcandre: Sure. Probably best I isolate it elsewhere

marcandre:
Hopefully third time's a charm...
I refactored the test helper to make it cleaner to check something against all constructed nodes.
That made it easy to put the `children_in_source_order` as a private method of the associator and makes the tests nicer and isolated.

iliabylich:
@marcandre Thanks a lot!

PR #703, opened 2020-06-07
+ Source::TreeRewriter: Improved merging and representations

marcandre:
This PR provides `TreeRewriter` with:
* improved merging (allowing different ProcessedSource or offsetting)
* two representations of the rewriter (a basic one and a complete one)

*Background*:
I was looking at [erb-lint](https://github.com/Shopify/erb-lint) and they do some processing on some excerpts of erb files. Later, they want to combine the rewriting actions on the original erb (with an offset).
With the upcoming RuboCop 1.0 this would be tricky, and this made me realize that the two avenues I was thinking about were not available (transferring a `TreeRewriter` with an offset, or at least having a representation of the TreeRewriter so we could recreate the actions on a different source and with an offset). This PR allows both avenues.

I added a minor doc change too. Let me know if I should split this into different PRs

### DIFF
--- a/lib/parser/source/tree_rewriter.rb
+++ b/lib/parser/source/tree_rewriter.rb
@@ -261,6 +261,44 @@ module Parser
       end
 
       ##
+      # Returns a representation of the rewriter as an ordered list of replacements.
+      #
+      #     rewriter.as_replacements # => [ [1...1, '('],
+      #                                     [2...4, 'foo'],
+      #                                     [5...6, ''],
+      #                                     [6...6, '!'],
+      #                                     [10...10, ')'],
+      #                                   ]
+      #
+      # This representation is sufficient to recreate the result of `process` but it is
+      # not sufficient to recreate completely the rewriter for further merging/actions.
+      # See `as_nested_actions`
+      #
+      # @return [Array<Range, String>] an ordered list of pairs of range & replacement
+      #
+      def as_replacements
+        @action_root.ordered_replacements
+      end
+
+      ##
+      # Returns a representation of the rewriter as nested insertions (:wrap) and replacements.
+      #
+      #     rewriter.as_actions # =>[ [:wrap, 1...10, '(', ')'],
+      #                               [:wrap, 2...6, '', '!'],  # aka "insert_after"
+      #                               [:replace, 2...4, 'foo'],
+      #                               [:replace, 5...6, ''],  # aka "removal"
+      #                             ],
+      #
+      # Contrary to `as_replacements`, this representation is sufficient to recreate exactly
+      # the rewriter.
+      #
+      # @return [Array<(Symbol, Range, String{, String})>]
+      #
+      def as_nested_actions
+        @action_root.nested_actions
+      end
+
+      ##
       # Provides a protected block where a sequence of multiple rewrite actions
       # are handled atomically. If any of the actions failed by clobbering,
       # all the actions are rolled back. Transactions can be nested.
--- a/lib/parser/source/tree_rewriter.rb
+++ b/lib/parser/source/tree_rewriter.rb
@@ -129,6 +129,8 @@ module Parser
       ##
       # Merges the updates of argument with the receiver.
       # Policies of the receiver are used.
+      # This action is atomic in that it won't change the receiver
+      # unless it succeeds.
       #
       # @param [Rewriter] with
       # @return [Rewriter] self
--- a/lib/parser/source/tree_rewriter.rb
+++ b/lib/parser/source/tree_rewriter.rb
@@ -310,7 +310,7 @@ module Parser
 
       def check_range_validity(range)
         if range.begin_pos < 0 || range.end_pos > @source_buffer.source.size
-          raise IndexError, "The range #{range} is outside the bounds of the source"
+          raise IndexError, "The range #{range.to_range} is outside the bounds of the source"
         end
         range
       end
--- a/lib/parser/source/tree_rewriter.rb
+++ b/lib/parser/source/tree_rewriter.rb
@@ -155,6 +155,32 @@ module Parser
       end
 
       ##
+      # For special cases where one needs to merge a rewriter attached to a different source_buffer
+      # or that needs to be offset. Policies of the receiver are used.
+      #
+      # @param [TreeRewriter] rewriter from different source_buffer
+      # @param [Integer] offset
+      # @return [Rewriter] self
+      # @raise [IndexError] if action ranges (once offset) don't fit the current buffer
+      #
+      def import!(foreign_rewriter, offset: 0)
+        return self if foreign_rewriter.empty?
+
+        contracted = foreign_rewriter.action_root.contract
+        merge_effective_range = ::Parser::Source::Range.new(
+          @source_buffer,
+          contracted.range.begin_pos + offset,
+          contracted.range.end_pos + offset,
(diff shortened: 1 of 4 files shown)


lib/parser/source/tree_rewriter/action.rb
```diff
@@ -46,10 +46,52 @@ def ordered_replacements
         reps
       end
 
+      def nested_actions
+        reps = []
```
iliabylich, line 50: I guess `reps` stands for `replacements` from the method above. Maybe let's rename it so it makes more sense? Like `actions` or `result`

lib/parser/source/tree_rewriter/action.rb
```diff
@@ -46,10 +46,52 @@ def ordered_replacements
...
+        reps << [:wrap, @range, @insert_before, @insert_after] unless @insert_before.empty? &&
```
iliabylich, line 51: nit: could you invert `unless a && b` into `if` please? It's a bit hard to follow

lib/parser/source/tree_rewriter/action.rb
```diff
@@ -46,10 +46,52 @@ def ordered_replacements
...
+          begin_pos: children.first.range.begin_pos,
+          end_pos: children.last.range.end_pos,
+        )
+        with(range: range)
+      end
+
+      ##
+      # @return [Action] that has been moved to the given source_buffer and with the given offset
+      # No check is done on validity of resulting range.
+      def move(source_buffer, offset)
```
iliabylich, line 80: the name feels like it modifies original object (at least that's what move operator in c++ does 😄 , maybe it's just a wrong association from my side). WDYT about renaming it to `moved` (or something better if you have an idea) to indicate that it doesn't affect receiver.

100% up to you, there's a ton of similar methods in Ruby (like `Hash#merge` and many many others) that are non-destructive.

lib/parser/source/tree_rewriter/action.rb
```diff
@@ -46,10 +46,52 @@ def ordered_replacements
...
+        moved_range = ::Parser::Source::Range.new(
+          source_buffer,
+          @range.begin_pos + offset,
+          @range.end_pos + offset
+        )
+        self.class.new(
```
iliabylich, line 86: is it possible to use `with` here?

test/test_source_tree_rewriter.rb
```diff
@@ -260,4 +265,91 @@ def test_merge
...
+                    [ 5...6, ''],
+                    [ 6...6, '!'],
+                    [ 10...10, ')'],
+                  ],
+                  result.map {|r, s| [r.to_range, s]}
+                )
+  end
+end
+
+class TestSourceTreeRewriterMerge < Minitest::Test
```
iliabylich, line 303: I'd probably inherit `TestSourceTreeRewriterMerge` from `TestSourceTreeRewriter` (or there could be a single base class). Dependencies between test classes are unusual.

lib/parser/source/tree_rewriter.rb
```diff
@@ -234,6 +248,44 @@ def process
...
+      #     rewriter.as_actions # =>[ [:wrap, 1...10, '(', ')'],
+      #                               [:wrap, 2...6, '', '!'],  # aka "insert_after"
+      #                               [:replace, 2...4, 'foo'],
+      #                               [:replace, 5...6, ''],  # aka "removal"
+      #                             ],
+      #
+      # Contrary to `as_replacements`, this representation is sufficient to recreate exactly
+      # the rewriter.
+      #
+      # @return [Array<Symbol, Range, String{, String}/>]
```
iliabylich, line 283: I'm not sure if this return type is correct. Documentation of YARD says that to have a tuple you should use `()`, and I can't find any usages of slashes, so it should be something like
```
Array<(Symbol, Range, String{, String})>
```

Could you double-check it locally please?

lib/parser/source/tree_rewriter/action.rb
marcandre: Indeed

lib/parser/source/tree_rewriter/action.rb
marcandre: Sure

lib/parser/source/tree_rewriter/action.rb
marcandre: Agreed

lib/parser/source/tree_rewriter/action.rb
marcandre: Oh my, yes, I should have thought of that. Sooo much nicer too.

test/test_source_tree_rewriter.rb
marcandre: I don't know Minitest well (and don't intend to, I'm too much a fan of rspec), but that's what I did first and the tests of `TestSourceTreeRewriter` get executed twice...

lib/parser/source/tree_rewriter.rb
marcandre: Slash was a typo, `()` sound perfect

marcandre:
So I'm proposing to add `#import!` and kept `#merge!` as it was. I don't think that there's a need for `#import`. 

My first draft had a method a bit like `moved` that add `source_buffer` and `offset` as arguments, but also needed `policy` and had to call `self.class.new` which is never great as it complicates subclassing (which we do in RuboCop now). Easier to have something like `import!` which mutates an already created object.

What do you think?

marcandre:
Tests crash with TruffleRuby, with a nice cryptic error...

test/test_source_tree_rewriter.rb
marcandre: If you prefer I could simply add the extra setup to the first part. I miss RSpec!

iliabylich:
It says
```
Fatal error: java.lang.OutOfMemoryError: Could not allocate an aligned heap chunk
```

@eregon Could you take a look please? Are there are additional JVM options that we could set to fix it?

Also, does JVM (just like MRI) run GC when malloc (or some wrapper around it) fails to get memory? If so, does it mean that we have a test that consumes too much memory so it causes oom error on truffleruby? Not sure how's it possible.

Here's the link to the failed build - https://travis-ci.org/github/whitequark/parser/jobs/696562301, I'll rerun it to see if it's consistent.

marcandre:
There was also talk about an "implicit exception raised when an explicit one would be fatal" or something. That made my head spin 😅

test/test_source_tree_rewriter.rb
iliabylich: Right, minitest simply invokes all `/test_.*/` methods that are defined twice because of the inheritance (I'm sure it could to `instance_methods(false)` to drop inherited methods, no idea why it's designed this way). However, creating a base class with a `setup` method (and inheriting 2 test classes from it) could work. Let's keep it as is

iliabylich:
And it's green now. I guess I'll move triffleruby to `allow_failures` group once it happens again.

> So I'm proposing to add #import! and keep #merge! as it was. I don't think that there's a need for #import (which would just be dup.import!).

I agree.

Is there anything you'd like to change? If so I'll merge it

marcandre:
LGTM, thanks 😄 

Thanks for the review. I need to tell you that your reviews have been the most in-depth, precise and on-point I remember having 🙇

PR #710, opened 2020-06-11
+ Added `emit_forward_arg` compatibility flag.

iliabylich:
The reason is that Ruby 2.8 supports the following code:
  `def m(a, b, ...); end`

If set to false (the default):
  1. `def m(...) end` is emitted as
     ```
     s(:def, :m, s(:forward_args), nil)
     ```

  2. `def m(a, b, ...) end` is emitted as
     ```
     s(:def, :m,
       s(:args, s(:arg, :a), s(:arg, :b), s(:forward_arg)))
     ```

If set to true it uses a single format:
  1. `def m(...) end` is emitted as
     ```
     s(:def, :m, s(:args, s(:forward_arg)))
     ```

  2. `def m(a, b, ...) end` is emitted as
     ```
     s(:def, :m, s(:args, s(:arg, :a), s(:arg, :b), s(:forward_arg)))
     ```

This is necessary to backport https://github.com/ruby/ruby/commit/f8b4340fa2c254cd093ebc3bc70d2d0c46ea9997

I'll keep it open for a few days, feel free to leave comments/suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ below for explanation of `emit_*` calls):
     Parser::Builders::Default.emit_encoding            = true
     Parser::Builders::Default.emit_index               = true
     Parser::Builders::Default.emit_arg_inside_procarg0 = true
+    Parser::Builders::Default.emit_forward_arg         = true
 
 Parse a chunk of code:
 
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -1164,13 +1164,13 @@ s(:numblock,
 
 ## Forward arguments
 
-### Method definition accepting forwarding arguments
+### Method definition accepting only forwarding arguments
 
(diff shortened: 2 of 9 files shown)


marcandre:
Just curious, when is `s(:def, :m, s(:forward_args), nil)` desirable?

iliabylich:
Never. It was a wrong decision initially, but some code can rely on it, so we can't break it easily. Ideally users should enable the flag and use the new format.

marcandre:
Is a release planned for `emit_forward_arg`? I'd like to enable it by default in RuboCop

iliabylich:
@marcandre 2.7.1.4 has been released - https://rubygems.org/gems/parser/versions/2.7.1.4

marcandre:
❤️ Thank you @iliabylich ❤️

PR #715, opened 2020-06-23
- Fix locations for alias / undef nodes with internal symbols

marcandre:
```sh
$ ruby-parse -L -e "alias foo x"
s(:alias,
  s(:sym, :foo),
  s(:sym, :x))
alias foo x
~~~~~ keyword         
~~~~~~~~~~~ expression
s(:sym, :foo)
alias foo x
      ~ begin       # <<< error
      ~~~ expression
s(:sym, :x)
alias foo x
          ~ begin     # <<< error
          ~ expression
```

I couldn't find a (simple) way to add a test, but it looks like tests aren't necessarily precise to that level. For example I temporarily changed `%i[foo]` to erroneously generate a `begin` and no test fail. Since this hasn't been noticed in so many years, clearly it's not super important. I noticed it by pure chance.

Let me know if it's necessary to add a test (which would probably involve changing `assert_parses` to accept some kind of annotation like `! begin` or accept a block to run extra checks...

### DIFF
--- a/test/parse_helper.rb
+++ b/test/parse_helper.rb
@@ -49,15 +49,17 @@ module ParseHelper
     end
   end
 
-  def assert_source_range(begin_pos, end_pos, range, version, what)
-    assert range.is_a?(Parser::Source::Range),
-           "(#{version}) #{range.inspect}.is_a?(Source::Range) for #{what}"
-
-    assert_equal begin_pos, range.begin_pos,
-                 "(#{version}) begin of #{what}"
-
-    assert_equal end_pos, range.end_pos,
-                 "(#{version}) end of #{what}"
+  def assert_source_range(expect_range, range, version, what)
+    if expect_range == nil
+      # Avoid "Use assert_nil if expecting nil from .... This will fail in Minitest 6.""
+      assert_nil range,
+                   "(#{version}) range of #{what}"
+    else
+      assert range.is_a?(Parser::Source::Range),
+             "(#{version}) #{range.inspect}.is_a?(Source::Range) for #{what}"
+      assert_equal expect_range, range.to_range,
+                   "(#{version}) range of #{what}"
+    end
   end
 
   # Use like this:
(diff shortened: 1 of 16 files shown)


iliabylich:
Good catch. So you propose changing `.begin` to `nil` for such nodes, right? (Just like for symbol nodes inside `%i[]`). I agree, but it definitely requires some changes to assertions.

There are a few tests that verify `expression` of all kinds of arguments given to `alias`, but there are no tests asserting that `begin` is empty. And we have no tests asserting that individual parts of source maps are empty.

Adding something like `! begin` seems to be a good idea to me. Do you want to include it here? If not I can add it by myself in a separate PR.

Also `undef foo` is affected:
```sh
bin/ruby-parse --28 -L -e 'undef a'
s(:undef,
  s(:sym, :a))
undef a
~~~~~ keyword
~~~~~~~ expression
s(:sym, :a)
undef a
      ~ begin
      ~ expression
```

marcandre:
Yes, I mentionned `undef` in the title but not in the body...

I'll handle the test (and indeed I forgot ruby_motion / mac_ruby...)

marcandre:
PR amended.

test/parse_helper.rb
```diff
@@ -273,7 +282,7 @@ def parse_source_map_descriptions(descriptions)
 
       if (match = SOURCE_MAP_DESCRIPTION_RE.match(line))
         begin_pos        = match[1].length
-        end_pos          = begin_pos + match[2].length
+        end_pos          = begin_pos + match[2].length unless match[2] == '!'
```
iliabylich, line 285: 1. the pattern allows `!!!! begin (ast_path)` that makes no sense. `!` does not represent any source range
2. this very special `if` feels like a hack
3. the caller checks `end_pos`, but there's also `begin_pos` and it's unclear why only one of them is important

What do you think about yielding something like `source_map_type` (as the first argument) that is either `:assert_source_map` or `:assert_no_source_map` (not the best naming, but I hope you get the idea). This way you could do `case source_map_type ...` on the call site to perform either `assert_source_range(...)` or `assert_no_source_range(...)`

marcandre:
Ok, I refactored things a bit then. Improves failures a bit, e.g.:

```
TestParser#test_heredoc [test/test_parser.rb:261]:
(1.8) range of "        ~~~~~~~~ heredoc_body".
Expected: 8...16
  Actual: 7...15
```

Instead of comparing begin_pos and then end_pos.

test/parse_helper.rb
marcandre: Most definitely a hack 😅

iliabylich:
@marcandre Thanks!

PR #716, opened 2020-06-27
+ emit "endless method def" as `:def` node.

iliabylich:
(according to https://github.com/whitequark/parser/pull/676#issuecomment-650327931)

2.8.0-dev is not released, so I think we are free to change it.

### DIFF
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -829,7 +829,7 @@ Format:
 Format:
 
 ~~~
-(def_e :foo (args) (int 42))
+(def :foo (args) (int 42))
 "def foo() = 42"
  ~~~ keyword
      ~~~ name
@@ -843,7 +843,7 @@ Format:
 Format:
 
 ~~~
-(defs_e (self) :foo (args) (int 42))
+(defs (self) :foo (args) (int 42))
 "def self.foo() = 42"
  ~~~ keyword
           ~~~ name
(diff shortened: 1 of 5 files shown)


iliabylich:
@marcandre PTAL, for some reason GH doesn't allow me to request a review from you

palkan:
@iliabylich I tried to migrate to this version and found that it doesn't work in a way we it was discussed: `node.loc.end.nil?` couldn't be used, since `Map::EndlessDefinition` has no `#end` method. Not sure what is a proper way to fix this: drop `EndlessDefinition` and add `assignment` to `Definition` or inherit `EndlessDefinition` from `Definition` or simply duplicate `attr_accessor :end`. @marcandre Thoughts?

marcandre:
On Thu, Jul 2, 2020, 06:05 Vladimir Dementyev <notifications@github.com>
wrote:

> Drop EndlessDefinition and add assignment to Definition
>

Seems the most useful...

>

marcandre:
> @iliabylich I tried to migrate to this version and found that it doesn't work in a way we it was discussed: `node.loc.end.nil?` couldn't be used, since `Map::EndlessDefinition` has no `#end` method. Not sure what is a proper way to fix this: drop `EndlessDefinition` and add `assignment` to `Definition` or inherit `EndlessDefinition` from `Definition` or simply duplicate `attr_accessor :end`. @marcandre Thoughts?

#718 should be a good solution

PR #718, opened 2020-07-04
- unify locations for normal and endless method definition

marcandre:
This way one can ask `loc.end` or `loc.assignment` of all method definitions

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -1803,17 +1803,17 @@ module Parser
     end
 
     def definition_map(keyword_t, operator_t, name_t, end_t)
-      Source::Map::Definition.new(loc(keyword_t),
-                                  loc(operator_t), loc(name_t),
-                                  loc(end_t))
+      Source::Map::MethodDefinition.new(loc(keyword_t),
+                                        loc(operator_t), loc(name_t),
+                                        loc(end_t), nil, nil)
     end
 
     def endless_definition_map(keyword_t, operator_t, name_t, assignment_t, body_e)
       body_l = body_e.loc.expression
 
-      Source::Map::EndlessDefinition.new(loc(keyword_t),
-                                         loc(operator_t), loc(name_t),
-                                         loc(assignment_t), body_l)
+      Source::Map::MethodDefinition.new(loc(keyword_t),
+                                        loc(operator_t), loc(name_t), nil,
+                                        loc(assignment_t), body_l)
     end
 
     def send_map(receiver_e, dot_t, selector_t, begin_t=nil, args=[], end_t=nil)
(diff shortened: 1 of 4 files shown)


marcandre:
Note: this keeps locations untouched for `Module`

lib/parser/builders/default.rb
```diff
@@ -1803,17 +1803,17 @@ def module_definition_map(keyword_t, name_e, operator_t, end_t)
...
     def definition_map(keyword_t, operator_t, name_t, end_t)
-      Source::Map::Definition.new(loc(keyword_t),
-                                  loc(operator_t), loc(name_t),
-                                  loc(end_t))
+      Source::Map::MethodDefinition.new(loc(keyword_t),
+                                        loc(operator_t), loc(name_t),
+                                        loc(end_t), nil, nil)
     end
 
     def endless_definition_map(keyword_t, operator_t, name_t, assignment_t, body_e)
```
iliabylich, line 1811: let's rename this method to `method_definition_map`

marcandre: Sure. But we'll need to keep an alias for compatibility, right?

iliabylich: `endless_definition_map` is (was) used only in `ruby28.y` that is still in development. I think it's ok to rename without any aliases

marcandre: Oh, sorry, I thought you wanted `definition_map` => `method_definition_map`.

Would you like to rename `endless_definition_map` => `method_definition_map` and that this is used for standard and endless method definitions?

My PR uses `definition_map` for standard methods, and `endless_definition_map` for endless ones. This way `definition_map` remains as a valid API and the implementation is updated, but maybe there's a better way to go.

marcandre: (seems it would be confusing if standard methods used `definition_map` and endless ones used `method_definition_map`)

iliabylich: nvm, I think it's better to keep it as is.

EDIT: just got a notification from GH that you reverted it. Yes, let's keep it as is, otherwise it sounds like it handles all method definitions. 👍

marcandre: Yes, I had misunderstood your comment (although you clearly made it on the `def endless_definition_map`, I should have paid more attention)

test/test_parser.rb
```diff
@@ -1856,6 +1856,7 @@ def test_def
       %q{def foo; end},
       %q{~~~ keyword
         |    ~~~ name
+        |! assignment
```
marcandre, line 1859: @iliabylich I'm glad you suggested we introduce this notation, didn't think I'd use it so soon 🙇‍♂️

iliabylich:
@marcandre Do you need a release?

marcandre:
Not really, but maybe @palkan would like one

palkan:
No rush, I'm good, thanks) Just want to make sure the tests pass against master.

PR #722, opened 2020-07-22
- Parser#parse returns nil instead of false if error is thrown

marcandre:
In some circumstances (when reaching the EOF too early), [`racc` erroneously returns `false` instead of `nil`](https://github.com/ruby/racc/pull/136).

Even though it's an upstream issue, the fix is so simple that I thought it was worthwhile to circumvent it.

This PR also improves the documentation for non-fatal cases. There are no other tests with `all_errors_are_fatal` set to `false`.

### DIFF
--- a/lib/parser/base.rb
+++ b/lib/parser/base.rb
@@ -177,16 +177,16 @@ module Parser
     end
 
     ##
-    # Parses a source buffer and returns the AST.
+    # Parses a source buffer and returns the AST, or `nil` in case of a non fatal error.
     #
     # @param [Parser::Source::Buffer] source_buffer The source buffer to parse.
-    # @return [Parser::AST::Node]
+    # @return [Parser::AST::Node, nil]
     #
     def parse(source_buffer)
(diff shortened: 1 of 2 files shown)


lib/parser/base.rb
```diff
@@ -210,8 +210,9 @@ def parse_with_comments(source_buffer)
 
     ##
     # Parses a source buffer and returns the AST, the source code comments,
-    # and the tokens emitted by the lexer. If `recover` is true and a fatal
-    # {SyntaxError} is encountered, `nil` is returned instead of the AST, and
+    # and the tokens emitted by the lexer. In case of a fatal error, a {SyntaxError}
```
iliabylich, line 213: ha, it took me a while to understand that `recover` here is responsible for "swallowing the error", but not "error recovery". Not sure if it's worth renaming it, current name sounds good as a part of the public API (you "recover" from VM error, not from the parsing error). Thanks for changing this doc too 👍

iliabylich:
@marcandre Thanks!

PR #724, opened 2020-07-24
+ Add Source::Range#line_span

marcandre:
This PR introduces a shortcut for `first_line..last_line` or `first_line...last_line`.

While the decision for `to_range` and `column_range` to be exclusive feels is always right, when mapping a range to lines I find this less clear, so I added a `exclude_end` keyword parameter.

Background: I'm still in a RuboCop refactor (definitely yak shaving, see last PR 😆 ) and noticing some occurrences of this pattern (a dozen inclusive, 3 exclusive). There's a method `line_range` for this in RuboCop, but it would be nicer as a building method and the name is unfortunate as it is easy to confuse with `SourceBuffer#line_range`.

I believe the concept is general enough to be worthwhile adding to `parser` directly, but if not this can be closed and I'll open a different PR for the doc fix I'm sneaking in the first commit 😅

### DIFF
--- a/lib/parser/source/range.rb
+++ b/lib/parser/source/range.rb
@@ -87,6 +87,27 @@ module Parser
       alias_method :first_line, :line
 
       ##
+      # The range of line numbers spanned by the range.
+      # The range will be exclusive or not depending on the keyword argument;
+      # default is inclusive.
+      #
+      # Assume that `def_node` is the Node corresponding to the following example:
+      #
+      #    def foo
+      #      do_something
+      #      42
+      #    end
+      #
+      # Then:
+      #
+      #    def_node.loc.end.line_span                           # => 4..4
+      #    def_node.loc.expression.line_span(exclude_end: true) # => 1...4
+      #
+      def line_span(exclude_end: false)
+        ::Range.new(first_line, last_line, exclude_end)
+      end
+
+      ##
       # @return [Integer] zero-based column number of the beginning of this range.
       #
       def column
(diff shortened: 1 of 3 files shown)


lib/parser/source/range.rb
```diff
@@ -86,6 +86,27 @@ def line
...
+      #      do_something
+      #      42
+      #    end
+      #
+      # Then:
+      #
+      #    def_node.loc.end.line_span                           # => 4..4
+      #    def_node.loc.expression.line_span(exclude_end: true) # => 1...4
+      #
+      def line_span(exclude_end: false)
```
iliabylich, line 106: It looks a bit confusing to me that it returns different ranges because of this boolean argument. I'd expect it to return:

+ `1..4` for `exclude_end = false`
+ `1...5` for `exclude_end = true`

What's the reason to add this method? Are there examples in rubocop that you could share? `exclude_end` doesn't "exclude" the `end` keyword in general (because there can be (a) single line method def and (b) not all statements have `end` keyword, this method doesn't make sense for oneliners like lvar assign)

marcandre: Right, this can be confusing. I was not referring to `end` keyword, or `location.end`, but to `::Range#exclude_end?`. 

For `1...4`, the thinking is to exclude `4` as the `def` doesnt' include that line's ending, but maybe it's just not clean enough overall.

For RuboCop, the typical use is to deal with comments; whenever we replace a particular range, then we need to take care to swallow any comments within that range (`exclude_end == true`), or if we want to know if it that range is commented (`exclude_end == false`).

Maybe it's best to keep it within RuboCop then?

iliabylich: Yes, it seems to be very Rubocop -specific

marcandre: Ok, thanks for your time 👍

PR #729, opened 2020-08-31
* ruby28.y -> ruby30.y

palkan:
Looks like it's time: https://github.com/ruby/ruby/commit/21c62fb670b1646c5051a46d29081523cd782f11

PR #731, opened 2020-09-01
- test_current.rb: 2.8 -> 3.0

palkan:
Follow-up for #729

### DIFF
--- a/test/test_current.rb
+++ b/test/test_current.rb
@@ -22,8 +22,8 @@ class TestCurrent < Minitest::Test
       assert_equal Parser::Ruby26, Parser::CurrentRuby
     when /^2\.7\.\d+/
       assert_equal Parser::Ruby27, Parser::CurrentRuby
-    when /^2\.8\.\d+/
-      assert_equal Parser::Ruby28, Parser::CurrentRuby
+    when /^3\.0\.\d+/
+      assert_equal Parser::Ruby30, Parser::CurrentRuby
     else
       flunk "Update test_current for #{RUBY_VERSION}"
     end


koic (approved):
👍

palkan:
@iliabylich Maybe, we shouldn't allow failures for ruby-head?

Another idea is to migrate to Github Actions and see separate statuses for different versions in PRs. And GA will likely be faster than Travis.

I have a working configuration [here](https://github.com/ruby-next/parser/tree/master/.github/workflows) which could be ported to this repo.

iliabylich:
> Maybe, we shouldn't allow failures for ruby-head?

Development branch of any repo on GH can be broken anytime. It shouldn't block us.

> migrate to Github Actions and see separate statuses

I look at ruby-head/rubocop statuses only when I cut a release or if I feel that PR has some potentially breaking changes. Otherwise it's useless (again, because both can be broken)

> And GA will likely be faster than Travis

I think it's fast enough. truffleruby is slow, but MRI tests take 1.5 minutes to run. Is truffleruby faster on GH actions? 👍 if so

palkan:
> Is truffleruby faster on GH actions? 

Looks like slow is not the only problem 😀 `Fatal error: java.lang.OutOfMemoryError` https://travis-ci.org/github/whitequark/parser/jobs/723110359

Currently, I do not test against Truffle. I will check.

iliabylich:
I've restarted the job and now it's green (this is the 2nd failure that I've seen so far).

There are a few more changes that should be done, I'll make them in a separate PR.

Thanks, merging!

palkan:
>  Is truffleruby faster on GH actions?

~7-8 minutes: https://github.com/ruby-next/parser/runs/1057046471

I will try to provide a PR a bit later.

iliabylich:
Ok, let's try it. Do you want to send a PR?

palkan:
> Do you want to send a PR?

Yeah, I will try to do that tomorrow.

PR #735, opened 2020-09-02
moved truffleruby to allow_failures group in travis.yml

iliabylich:


eregon:
Sorry about that, I will try to investigate why we run in a fatal OutOfMemoryError here: https://travis-ci.org/github/whitequark/parser/jobs/727303862

eregon:
For whatever reason it seems I cannot reproduce this in 3 runs: https://travis-ci.org/github/eregon/parser/builds

iliabylich:
right, it fails randomly, but I've noticed that the error comes on the last step of the test suite where we check that all AST nodes that are created are known for our AST visitor.

Here's the `for_each_node` method - https://github.com/whitequark/parser/blob/7c5558df1d1842857ba8883219916c96528858d6/test/helper.rb#L70, there are a few places in tests that call it. Basically, we record all nodes and do a couple of post-checks in the `Minitest.after_run {}` block.

PR #737, opened 2020-09-03
meta: root is no longer used

akimd:
Last use was replaced by zsuper in cbbb6bbd28cac20dc1a618c7d4779dce12f5f749.

meta also mentions 'ident'.  I think it is used only temporarily by the parser and never shows in final ASTs.  Is that right?


iliabylich:
Thanks!

> it is used only temporarily by the parser and never shows in final ASTs. Is that right?

Yes

PR #743, opened 2020-09-24
Exclude .gitignore from the package to support vendoring.

iliabylich:
If .gitignore is a part of the .gem it's impossible to include source of
the locally installed parser (using bundle install --path ./vendor) because git
automatically excludes all gitignore-d files (including all parsers and lexer).

### DIFF
--- a/parser.gemspec
+++ b/parser.gemspec
@@ -35,7 +35,10 @@ Gem::Specification.new do |spec|
                           lib/parser/ruby30.rb
                           lib/parser/macruby.rb
                           lib/parser/rubymotion.rb
+                       ) - %w(
+                          .gitignore
                        )
+
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.test_files    = spec.files.grep(%r{^test/})
   spec.require_paths = ['lib']


parser.gemspec
```diff
@@ -35,7 +35,10 @@ Gem::Specification.new do |spec|
                           lib/parser/ruby30.rb
                           lib/parser/macruby.rb
                           lib/parser/rubymotion.rb
+                       ) - %w(
+                          .gitignore
```
mbj, line 39: @iliabylich OT: But thought triggered by your commit, I think parser should not use `git` at all to find packaged files.

See: https://github.com/mbj/mutant/issues/1052 where I improved the packagability for debian and others.

iliabylich: wait, is there a debian package of `parser`? This field is used to build `.gem` file locally and push it to rubygems.org. I agree that it requires `git` locally to build it (which is kinda bad if you only to build a package), but at the same time it offers freedom of using any test files that are added to global `.gitignore` (I have files like `test<1-10>.rb` there).

I'd prefer to keep it as is if nobody needs it. Feel free to send a PR that fixes it though, I'll merge it (and I'll do cleanup of my local garbage 😄 ), but if it's just a "good practice" that nobody needs I'd like to ignore it 👀

mbj: @utkarsh2102 feel free to weigh in here.

utkarsh2102: Eee, hello 👋🏻 

I am the maintainer of `parser` in Debian. It's actively maintained and is synced with the latest version! 😄 
(so first of all, thanks for your work, really! 💯)

Yep, thanks @mbj for pointing this out, using `git` is a bit troublesome and it's best if avoided!
The complete rationale is here: https://docs.rubocop.org/rubocop-packaging/cops_packaging.html#gemspec-git-rationale

I'll be v happy to send a PR to drop `git` from gemspec! 🌮

iliabylich: > I'll be v happy to send a PR to drop git from gemspec! 🌮

Yes, please. I think `{bin,lib}/**/*.rb` pattern should be enough

mbj: This will also drop the `.y` which are not needed for a bundled gem. Should be a nice size reduction.

utkarsh2102: Okay, great! Will open a PR later today; completing my school homework for now! 😭

PR #745, opened 2020-09-25
Drop `git` in gemspec

utkarsh2102:
Hi @iliabylich,

As discussed in #743, here's the PR, dropping `git` in gemspec.
Rationale at: https://docs.rubocop.org/rubocop-packaging/cops_packaging.html#gemspec-git-rationale

This helps in maintaining this in Debian and other downstream 💫 

Signed-off-by: Utkarsh Gupta <<utkarsh@debian.org>>

### DIFF
--- a/parser.gemspec
+++ b/parser.gemspec
@@ -20,27 +20,8 @@ Gem::Specification.new do |spec|
     'source_code_uri' => "https://github.com/whitequark/parser/tree/v#{spec.version}"
   }
 
-  spec.files         = `git ls-files`.split + %w(
-                          lib/parser/lexer.rb
-                          lib/parser/ruby18.rb
-                          lib/parser/ruby19.rb
-                          lib/parser/ruby20.rb
-                          lib/parser/ruby21.rb
-                          lib/parser/ruby22.rb
-                          lib/parser/ruby23.rb
-                          lib/parser/ruby24.rb
-                          lib/parser/ruby25.rb
-                          lib/parser/ruby26.rb
-                          lib/parser/ruby27.rb
-                          lib/parser/ruby30.rb
-                          lib/parser/macruby.rb
-                          lib/parser/rubymotion.rb
-                       ) - %w(
-                          .gitignore
-                       )
-
+  spec.files         = Dir['bin/*', 'lib/**/*.rb', 'parser.gemspec']
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
-  spec.test_files    = spec.files.grep(%r{^test/})
   spec.require_paths = ['lib']
 
   spec.required_ruby_version = '>= 2.0.0'
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@ Gemfile.lock
 InstalledFiles
 _yardoc
 coverage
-lib/bundler/man
+lib/bundler/man/*
 pkg
 rdoc
 yardoc


parser.gemspec
```diff
@@ -20,27 +21,26 @@ Gem::Specification.new do |spec|
...
-                          lib/parser/ruby25.rb
-                          lib/parser/ruby26.rb
-                          lib/parser/ruby27.rb
-                          lib/parser/ruby30.rb
-                          lib/parser/macruby.rb
-                          lib/parser/rubymotion.rb
-                       ) - %w(
-                          .gitignore
-                       )
+  spec.files         = Rake::FileList['{bin,lib}/**/*'].exclude(*File.read('.gitignore').split)
```
iliabylich, line 24: This is a bit hard to follow. Also it forces `.gitignore` to have entries that are valid for `FileList` (this I why you had to change it I suppose)

Does it work if you do this?

```ruby
Dir['bin/*'] + Dir['lib/**/*.rb'] + 'parser.gemspec'
```

You can check it by doing

```
$ gem uninstall parser
$ gem uninstall ast
$ rake build
$ gem install pkg/parser-2.7.1.4.gem
$ ruby-parse -e '1 + 1'
```

iliabylich: also `spec/test_files` seems to be optional, and we don't want to include test files into release build, could you remove this assignment too, please?

utkarsh2102: Perfect, anything that's pure Ruby works for us! Just not `git` 😄 
I'll fix this right away! 😄

utkarsh2102: Does it work for you now?

parser.gemspec
```diff
@@ -20,27 +20,24 @@ Gem::Specification.new do |spec|
...
-                          lib/parser/ruby26.rb
-                          lib/parser/ruby27.rb
-                          lib/parser/ruby30.rb
-                          lib/parser/macruby.rb
-                          lib/parser/rubymotion.rb
-                       ) - %w(
-                          .gitignore
-                       )
+  spec.files         = Dir['bin/*'] + Dir['lib/**/*.rb'] + 'parser.gemspec' + %w(
+                           lib/parser/lexer.rb
```
iliabylich, line 24: `lib/parser/lexer.rb` and `lib/parser/ruby*.rb` are twice in the list. The last part `+ %w( ... )` should be removed.

parser.gemspec
```diff
@@ -59,4 +56,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'simplecov', '~> 0.15.1'
 
   spec.add_development_dependency 'gauntlet'
+  spec.add_development_dependency 'rubocop-packaging'
```
iliabylich, line 59: do we need it?

iliabylich (changes requested):

parser.gemspec
utkarsh2102: Eee, right. My bad.

parser.gemspec
utkarsh2102: It helps to ensure that the library is easily maintainable downstream (not only Debian, but other distributions as well!).
The `Packaging` extension was explicitly created to help downstream maintainers without compromising the experience of upstream maintainers. 

Let me know what you think?

iliabylich: it's a dependency that `parser` doesn't need. could you remove it please? it doesn't affect development process and it's not even required anywhere. it will only slow down CI builds for no practical reason

utkarsh2102: fine, removed.

iliabylich:
@utkarsh2102 my bad, should be `+ ['parser.gemspec']` (because we concatenate arrays)

utkarsh2102:
> @utkarsh2102 my bad, should be `+ ['parser.gemspec']` (because we concatenate arrays)

I've taken all together at once now. Should be good to go.

iliabylich (approved):
👍

iliabylich:
@utkarsh2102 Thanks!

utkarsh2102:
Hi @iliabylich,
I dropped you a mail, could you take a look at it, whenever you have time? 😄

parser.gemspec
```diff
@@ -20,27 +20,8 @@ Gem::Specification.new do |spec|
...
-                          lib/parser/ruby26.rb
-                          lib/parser/ruby27.rb
-                          lib/parser/ruby30.rb
-                          lib/parser/macruby.rb
-                          lib/parser/rubymotion.rb
-                       ) - %w(
-                          .gitignore
-                       )
-
+  spec.files         = Dir['bin/*', 'lib/**/*.rb', 'parser.gemspec']
```
mbj, line 23: BTW, and this is not a bad thing: before `parser` would ship anything which was not gitignored explicity (or in the whitelist) this means that the `parser` gem would not ship tests anymore. Which is IMO a good thing.

utkarsh2102: Yep, happy to have it this way!

Issue #742, opened 2020-09-24
NoMethodError in parsing numparams with ruby 2.6 parser.


mbj:
Tested with `parser-2.7.1.4` the following crashes the parser with a `NoMethodError`. I'd expect a `Parser::SyntaxError` instead.

```ruby
require 'parser/ruby26'

Parser::Ruby26.parse(<<~'RUBY')
  proc do
    _1 = nil
  end
RUBY
```

Exception backtrace:

```
NoMethodError: undefined method `>' for nil:NilClass
  /home/mutant-dev/.gem/ruby/2.6.6/gems/parser-2.7.1.4/lib/parser/max_numparam_stack.rb:21:in `has_numparams?'
  /home/mutant-dev/.gem/ruby/2.6.6/gems/parser-2.7.1.4/lib/parser/builders/default.rb:1569:in `check_assignment_to_numparam'
  /home/mutant-dev/.gem/ruby/2.6.6/gems/parser-2.7.1.4/lib/parser/builders/default.rb:592:in `assignable'
  /home/mutant-dev/.gem/ruby/2.6.6/gems/parser-2.7.1.4/lib/parser/ruby26.rb:3960:in `_reduce_102'
  /home/mutant-dev/.rubies/ruby-2.6.6/lib/ruby/2.6.0/racc/parser.rb:259:in `_racc_do_parse_c'
  /home/mutant-dev/.rubies/ruby-2.6.6/lib/ruby/2.6.0/racc/parser.rb:259:in `do_parse'
  /home/mutant-dev/.gem/ruby/2.6.6/gems/parser-2.7.1.4/lib/parser/base.rb:189:in `parse'
[snip callstack from unparser tests]
```

Note that this 2.7 syntax example is actually in the test corpus of parser itself. But as an 2.7 example. I wounder if tests for Ruby version N should be run by testsfor >N as "expect syntax error" examples.

So this issue proposes 2 things:

A) Fix the `NoMethodError` and raise a `SyntaxError` instead.
B) Re-use examples for newer syntax as expect syntax error examples for older parsers.

I'll take a stab ad A) later. But B) is open for discussion. IMO its a very useful property to re-use the test corpus this way.

iliabylich:
Interesting. I can't reproduce it on the latest master, looks like I (accidentally) fixed it here https://github.com/whitequark/parser/pull/725/files#diff-0a90f5f53d48f851460a59d188b639c1R21.

I'll cut a release in a few hours, thanks for report.

[...]

PR #746, opened 2020-09-26
test older parsers don't throw non-syntax errors on new syntax

iliabylich:
raising an error in the `else` branch makes 42 tests to fail (i.e. there are 42 breaking MRI changes that we are aware of)

@mbj PTAL, do you have anything to add?

### DIFF
--- a/test/parse_helper.rb
+++ b/test/parse_helper.rb
@@ -84,6 +84,24 @@ module ParseHelper
       parser.instance_eval { @lexer.force_utf32 = true }
       try_parsing(ast, code, parser, source_maps, version)
     end
+
+    # Also check that it doesn't throw anything
+    # except (possibly) Parser::SyntaxError on other versions of Ruby
+    with_versions(ALL_VERSIONS - versions) do |version, parser|
+      begin
+        source_file = Parser::Source::Buffer.new('(assert_older_rubies)', source: code)
+        parser.parse(source_file)
+      rescue Parser::SyntaxError
+        # ok
+      rescue StandardError
+        # unacceptable
+        raise
+      else
+        # No error means that `code` is valid for `version`, but has a different meaning.
+        # Sometimes Ruby has breaking changes (like numparams)
+        # that re-use constructions from previous versions.
+      end
+    end
   end
 
   def try_parsing(ast, code, parser, source_maps, version)


mbj:
> PTAL, do you have anything to add?

This prevents regressions like the ones I reported in #742. I think it qualifies to the idea we had in the discussions around that issue.

Thanks.

marcandre:
Cool stuff. I imagine that a test that the AST produced is different could be added (in the no error case)?

mbj:
> Cool stuff. I imagine that a test that the AST produced is different could be added (in the no error case)?

Yes, that makes sense. May require some whitelisting for new syntax added that is just sugar for existing AST. But likely a good default.

OT: I really like these "static-property-tests". Where you take a corpus of test inputs and try to find as many correctness signals as possible.

Another OT: @iliabylich I've started to re-use parsers test suite in `unparser` (found tons of issues in unparser) and I'm currently resorting to this hack: https://github.com/mbj/unparser/pull/148/files#diff-6e44895a7a81a0ca94584e0d4d9ebc1f. Basically I implement a "fake" minitest to capture the examples.

I wounder if you'd support changing most of the tests to be defined in a data structure that is walked by minitest. This way use cases like unparser (and likely other tools that need to be exposed to syntax edge cases) can require that data without having to resort to "drive by capturing" as I do over at that unparser branch.

If you like the idea I'd be happy to start a PR on that.

Note we could still generate the exactly same minitest surface. Just the minitest `test_foo` a-like methods would be generated from that data structure.

iliabylich:
> Cool stuff. I imagine that a test that the AST produced is different could be added

Agree, maybe we'll find that some `SINCE_X_Y` rules are set incorrectly.

> Another OT: @iliabylich I've started to re-use parsers test suite in unparser (found tons of issues in unparser) and I'm currently resorting to this hack: mbj/unparser#148 (files). Basically I implement a "fake" minitest to capture the examples.

haha, I did the same multiple times (https://github.com/sorbet/sorbet/blob/master/tools/scripts/import_whitequark.rb for example)

> I wounder if you'd support changing most of the tests to be defined in a data structure that is walked by minitest

Could you provide an example of how it might look like? I'm not sure if I fully understand what you want to be simplified, existing test suite looks quite simple to me for exporting

PR #747, opened 2020-09-30
Simplify code

marcandre:
I presume this dates from before Ruby 2.0 as minimum target.

### DIFF
--- a/lib/parser/meta.rb
+++ b/lib/parser/meta.rb
@@ -7,7 +7,7 @@ module Parser
     # All node types that parser can produce. Not all parser versions
     # will be able to produce every possible node.
     NODE_TYPES =
-      %w(
+      %i(
         true false nil int float str dstr
         sym dsym xstr regopt regexp array splat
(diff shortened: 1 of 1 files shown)


bbatsov:
Guess so. For a very long time parser supported Ruby 1.8.

bbatsov:
Btw, I've noticed that the README doesn't mention that Ruby 2+ is required (you can only see this in the gemspec/CI), so we should probably mention this as well.

marcandre:
> Btw, I've noticed that the README doesn't mention that Ruby 2+ is required

It does actually: "runs on MRI >=2.0.0"

iliabylich:
Thanks!

bbatsov:
> It does actually: "runs on MRI >=2.0.0"

My bad, I didn't think to check "Features" for this. :-) Usually there's some "Requirements/Compatibility" section about this, so it's easier to find. Anyways, not a big deal.

PR #756, opened 2020-10-28
Source buffer freeze

marcandre:
Builds on #755 and makes `SourceBuffer` freezable. Currently attempting to freeze before caches are build will crash.

### DIFF
--- a/lib/parser/source/buffer.rb
+++ b/lib/parser/source/buffer.rb
@@ -299,6 +299,12 @@ module Parser
         line_begins.size + @first_line - 2
       end
 
+      # :nodoc:
+      def freeze
+        source_lines; line_begins; source_range # build cache
+        super
+      end
+
       private
 
       # @returns [0, line_begin_of_line_1, ..., source.size + 1]
--- a/test/test_source_buffer.rb
+++ b/test/test_source_buffer.rb
@@ -164,4 +164,12 @@ class TestSourceBuffer < Minitest::Test
     @buffer.source = "foo\nbar"
     assert_equal ['foo', 'bar'], @buffer.source_lines
   end
+
+  def test_freeze
+    @buffer.source = "1\nfoo\nbar\n"
+    @buffer.freeze
+    @buffer.source_lines
+    @buffer.source_range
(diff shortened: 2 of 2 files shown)


lib/parser/source/buffer.rb
```diff
@@ -303,27 +296,49 @@ def source_range
       # @return [Integer]
       #
       def last_line
-        line_begins.size + @first_line - 1
+        line_begins.size + @first_line - 2
+      end
+
+      # :nodoc:
+      def freeze
```
iliabylich, line 303: I don't know if I agree with this. I mean, I understand the rationale behind it, but I'm just afraid that MRI can implicitly call this method in different places without going through dynamic dispatch, i.e. it can call `rb_freeze` (of how's it called) to set a bitflag directly.

It's better than what's currently in master, but it's a source of potential confusion.

If these ivars are set in most cases we could compute them in the constructor. If not - I guess I'm ok with overriding `.freeze`.

Also, side question: why do you want buffer to be frozen?

marcandre: > I'm just afraid that MRI can implicitly call this method in different places without going through dynamic dispatch

AFAIK, there is no such circumstance currently; Ruby never sets the freeze flag directly. I've been advocating so that there isn't, so far so good and `Ractor.make_shareable` calls user-defined `freeze` methods (and then proceeds to deep freeze by calling `#freeze` on its unfrozen components).

> why do you want buffer to be frozen?

Currently, there is no good reason to. 
I understand that it's typical in Ruby to not "care" about `freeze` (or `initialize_copy` for mutable), but it is part of the public API, and currently if someone calls it, it may crash.

An alternative implementation would be to store the caches inside a struct. This way we could `freeze` the buffer when it is completed with a source, and not need to prebuild the caches (unless the struct itself gets frozen, a sign that we are being deep-frozen). This would reflect the fact that, once the source is given, the buffer is immutable (`frozen? # => true`).

iliabylich: > if someone calls it, it may crash

I'd not expect anyone to freeze objects that they don't own, but I agree with your point.

iliabylich (approved):
👍

PR #757, opened 2020-10-28
Make `SourceBuffer` Ractor-shareable. Basic test for `Node`.

marcandre:
Build on #756 so that `SourceBuffer` is `Ractor`-shareable.

Since `Node`, `Range` and `Map` are typically frozen, there's probably no issue with them.
`SourceBuffer` is the only non-frozen element in the chain.

`Ractor` is still evolving, maybe you'll want to wait until X-mas to merge this...

### DIFF
--- a/lib/parser/source/buffer.rb
+++ b/lib/parser/source/buffer.rb
@@ -323,7 +323,11 @@ module Parser
 
       # @returns 0-based line index of position
       def line_index_for_position(position)
-        @line_index_for_position[position] ||= bsearch(line_begins, position) - 1
+        @line_index_for_position[position] || begin
+          index = bsearch(line_begins, position) - 1
+          @line_index_for_position[position] = index unless @line_index_for_position.frozen?
+          index
+        end
       end
 
       if Array.method_defined?(:bsearch_index) # RUBY_VERSION >= 2.3
--- a/test/test_base.rb
+++ b/test/test_base.rb
@@ -28,4 +28,11 @@ class TestBase < Minitest::Test
     assert_nil ast.loc.dup.node
     Parser::AST::Node.new(:zsuper, [], :location => ast.loc)
   end
+
+  def test_node_ractor
+    ast = Parser::CurrentRuby.parse('1')
+    ::Ractor.make_shareable(ast)
+    assert ::Ractor.shareable?(ast)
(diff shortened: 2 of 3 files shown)


iliabylich:
> so that `SourceBuffer` is `Ractor`-shareable.

What's the value of making it shareable? If you want to parallelize rubocop it's easier to split the work per files, no? Then you need to make string (i.e. filepaths) shareable, what's already done in MRI.

marcandre:
> > so that `SourceBuffer` is `Ractor`-shareable.
> 
> What's the value of making it shareable? If you want to parallelize rubocop it's easier to split the work per files, no?

Sorry I wasn't clear. The value is not in making shareable `SourceBuffer` per say , but `Node` or `Range`. Attempting to send either to a Ractor without making it shareable first would result in a deep-copy of everything (including the source and all other nodes, ranges, maps, etc). Making it shareable first could yield a potential crash down the line if a cache for a particular position is then built.

I think it can be useful to send `Nodes` and `Range` to Ractors. Not sure for RuboCop today, but for example I have a site to [apply a RuboCop node-pattern to Ruby code](https://nodepattern.herokuapp.com/). Think `regexp` but for Ruby AST nodes. I'd like to add a tab to search for a particular node pattern in a collection of popular gems. I'd like the results to start appearing as soon as they are found. I'm thinking of a `Ractor` that I can give a pattern and a path to and it sends back whatever nodes it finds that matches, as it finds them. I can then format this in HTML or Json and send that down to the web socket. I could do the formatting within the Ractor and send the formatted result back, but it feels like the wrong design; I'd rather send the found `Node`s.

I would ask the reverse: what value do you see in keeping it "may-crash-or-not-if-made-shareable"?

iliabylich:
Ok, I get it now. One thing I forgot is that node locations store references to `source_buffer`, and so to make nodes shareable we also need to make source buffer shareable. Sending just a buffer still looks useless to me (but I understand that it's just a consequence of the current design).

👍 , Travis is lagging atm, I'll merge all 3 PRs once they are green

iliabylich:
@marcandre could you rebase please?

marcandre:
Rebased 👍

iliabylich:
Thanks!

Issue #753, opened 2020-10-26
ruby/ruby@52c630d


iliabylich:
https://github.com/ruby/ruby/commit/52c630da004d9273e8e5fc91c6304e9eed902566

PR #760, opened 2020-10-30
+ ruby30.y: use `=>` for one-line pattern matching.

iliabylich:
This commit tracks upstream commit ruby/ruby@52c630d.

Closes https://github.com/whitequark/parser/issues/753

@palkan Could you confirm that you don't use `Builder#rassign`/`multi_rassign` please? This API is a part of the 3.0 parser, so I guess only you can use it 😄

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -655,14 +655,6 @@ module Parser
         binary_op_map(lhs, eql_t, rhs))
     end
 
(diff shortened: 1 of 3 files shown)


palkan:
> Could you confirm that you don't use Builder#rassign/multi_rassign please? 

My parser still uses them but that's OK. I believe no one except me uses edge features) Let's merge and I'll upgrade it.

Issue #754, opened 2020-10-27
ruby/ruby@bdd1d17


iliabylich:
https://github.com/ruby/ruby/commit/bdd1d17ac2c14ed937944860e1ec6b361a2fda6d

PR #765, opened 2020-11-23
+ ruby30.y: allow endless method without arglist.

iliabylich:
This commit tracks upstream commit ruby/ruby@bdd1d17.

Closes https://github.com/whitequark/parser/issues/754

I noticed that we have an exception to our rule that "all nodes have expression_l" - that's not true from the very beginning:

```sh
$ bin/ruby-parse --30 -L -e 'def m; end'
s(:def, :m,
  s(:args), nil)
def m; end
~~~ keyword
    ~ name
       ~~~ end
~~~~~~~~~~ expression
s(:args)
```

^ `args` node has absolutely no source maps. `def foo = 42` also emits a dummy `args` node.

@marcandre @mbj @palkan Side question for this PR (I'm going to merge it anyway as it doesn't break anything), but what's your hunch on how many things are going to break if we make `s(:args)` nullable?

### DIFF
--- a/lib/parser/ruby30.y
+++ b/lib/parser/ruby30.y
@@ -847,7 +847,7 @@ rule
                       result = @builder.ternary(val[0], val[1],
                                                 val[2], val[4], val[5])
                     }
-                | defn_head f_paren_args tEQL arg
+                | defn_head f_opt_paren_args tEQL arg
                     {
                       _def_t, name_t = val[0]
                       endless_method_name(name_t)
@@ -861,7 +861,7 @@ rule
                       @context.pop
                       @current_arg_stack.pop
                     }
-                | defn_head f_paren_args tEQL arg kRESCUE_MOD arg
+                | defn_head f_opt_paren_args tEQL arg kRESCUE_MOD arg
                     {
                       _def_t, name_t = val[0]
                       endless_method_name(name_t)
@@ -881,7 +881,7 @@ rule
                       @context.pop
                       @current_arg_stack.pop
                     }
-                | defs_head f_paren_args tEQL arg
+                | defs_head f_opt_paren_args tEQL arg
                     {
                       _def_t, _recv, _dot_t, name_t = val[0]
                       endless_method_name(name_t)
@@ -895,7 +895,7 @@ rule
                       @context.pop
                       @current_arg_stack.pop
                     }
-                | defs_head f_paren_args tEQL arg kRESCUE_MOD arg
+                | defs_head f_opt_paren_args tEQL arg kRESCUE_MOD arg
                     {
                       _def_t, _recv, _dot_t, name_t = val[0]
                       endless_method_name(name_t)
@@ -2604,6 +2604,12 @@ keyword_variable: kNIL
                 | # nothing
                     {
                       result = nil
+                    }
+
+f_opt_paren_args: f_paren_args
+                | none
+                    {
+                      result = @builder.args(nil, [], nil)
                     }
 
    f_paren_args: tLPAREN2 f_args rparen
--- a/test/test_parser.rb
+++ b/test/test_parser.rb
@@ -9682,21 +9682,6 @@ class TestParser < Minitest::Test
     Parser::Builders::Default.emit_forward_arg = true
   end
 
-  def test_endless_method_without_brackets
-    assert_diagnoses(
-      [:error, :unexpected_token, { :token => 'tEQL' }],
-      %Q{def foo = 42},
-      %q{        ^ location},
-      SINCE_3_0)
-
-    assert_diagnoses(
-      [:error, :unexpected_token, { :token => 'tEQL' }],
-      %Q{def obj.foo = 42},
-      %q{            ^ location},
-      SINCE_3_0
-    )
-  end
-
   def test_endless_method_with_rescue_mod
     assert_parses(
       s(:def, :m,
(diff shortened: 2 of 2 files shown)


marcandre:
> how many things are going to break if we make `s(:args)` nullable

In `rubocop`, it would be easy to not break anything; `DefNode#arguments` could probably return `[]` (or `s(:args)`) and avoid any break.

If we keep the `nil` result, then 17 buitin cops break, not that bad.

That being said, my personal preference is for keeping it as is; I find it easier in general to deal with "stable" ASTs structures. For example one thing I dislike in the current AST is that the body of a `def` node has 3 different "structures": `nil`, a single node like `s(:send ...)`, or `s(:begin, ...)`. Call me strange, but I would much prefer that in all cases there would be a `s(:begin, ...)`, respectively with 0, 1 or many children 😅

Finally, what about `def foo(); ...`; you'd still have an `args` node in this case, right? Not that I want to rehash this debate, but imo AST should reflect the semantics and be thought independently of locations. `def foo; ...` is the same with or without parenthesis as far as Ruby is concerned, the AST should be the same.

mbj:
> but what's your hunch on how many things are going to break if we make `s(:args)` nullable?

Unparser and mutant will break, hard. But thats easy to overcome. Should not block improvement.

iliabylich:
Ok, then let's keep it as is.

mbj:
@iliabylich My comment was meant the other way round: I'm for this change. The `s(:args)` carries no important meaning in the mentioned case.

I'm fine to break thinks in the interest of a cleaner AST. A fix on my side is ~5min of work.

mbj:
Plus we could do the change with the 3.x release of the `parser` gem which is an indirect semver. Unlikely we'll break existing unparser/mutant usage as both constrain to `2.7` releases of `parser`.

iliabylich:
Yes, I understand it.

The reason why I wanted to keep it is that I think that if it breaks abstract libraries depending on `parser`, what's gonna happen with libraries where things are hardcoded? (or just tightly coupled with the current AST structure).

> Plus we could do the change with the 3.x release of the parser gem which is an indirect semver

That's a very good point. Ok, I'll create an issue for changing it.

> I would much prefer that in all cases there would be a `s(:begin, ...)`, respectively with 0, 1 or many children

I agree, and 3.0 is a good moment to fix it. `begin` can have empty `begin_l/end_l`

mbj:
There are possibly others we may want to fix. Example there are `dstr` nodes without a dynamic component. And possibly the chance to normalize some `dstr` nodes in a way they do not depend on the actual syntax that was used anymore while yielding different AST that is semantically equivalent on evaluation.

iliabylich:
Could you provide an example please?

mbj:
First class:

```
mutant-dev@mbj ~/devel/mutant (master*) $ bundle exec ruby-parse -e '%()'                                          <<<
(dstr)
mutant-dev@mbj ~/devel/mutant (master*) $ bundle exec ruby-parse -e '%(foo)'
(str "foo")
```

To me: This is inconsistent. We only should ever emit `dstr` nodes if there is actually a dynamic component.

I've got more painful ones in the unparser test suite. Let me open a separate issue for these should we agree on improving the AST format. I'm busy right now and you'd have to wait till the end of this week to get a consistent list.

iliabylich:
ah, right, currently we "flatten" an AST if there's exactly one child (but not 0)

mbj:
BTW: I think the dstr changes I'd like to see are not breaking the AST format. But they'd porbably fit well into the 3.x format.

mbj:
@iliabylich I found more examples I already have elaborated on: https://github.com/mbj/unparser/issues/196#issuecomment-723717238

marcandre:
I'm confused: so will `(args)` become nullable?
And we're contemplating a switch so that the body of a `def` & al will always be in a `begin`, (plus other changes like dstr)? Thinking about it more, the worst case is the case with 1 child; wrapping it in a `begin` is a really good idea. `nil` for no children is not that bothersome, but yeah, I still think I prefer "variable length AST types" (like `args` and `begin`) to be always present.

If we're looking for issues with the current format, I can't think of much else that's bothersome. I'd could add the discrepancy between `:foo` and `:foo=` in:
```
$ ruby-parse -e 'self.foo = 0; self.foo ||= 0; self.foo, bar = 0'
(begin
  (send
    (self) :foo=
    (int 0))
  (or-asgn
    (send
      (self) :foo)   # would be best as `:foo=`
    (int 0))
  (masgn
    (mlhs
      (send
        (self) :foo=)
      (lvasgn :bar))
    (int 0)))
```

The case with variable always has `lvasgn` and is fine:
```
$ ruby-parse -e 'foo = 0; foo ||= 0; foo, bar = 0'
(begin
  (lvasgn :foo
    (int 0))
  (or-asgn
    (lvasgn :foo)  # it's not `lvar`, good
    (int 0))
  (masgn
    (mlhs
      (lvasgn :foo)
      (lvasgn :bar))
    (int 0)))
```

iliabylich:
> I'm confused: so will (args) become nullable?

Yes, correct me if I'm wrong but the rule "all nodes have source maps" is extremely valuable. Without it we can end up creating "virtual" nodes everywhere (for flags and other dirty things).

> And we're contemplating a switch so that the body of a def & al will always be in a begin

Yes, here's a ticket - https://github.com/whitequark/parser/issues/767

> (plus other changes like dstr)?

Right, but I need more details (except for the case mentioned above)

> `nil` for no children is not that bothersome, but yeah, I still think I prefer "variable length AST types" (like args and begin) to be always present.

Body of `def m; end` has no source maps, it has to be `nil`

Issue #761, opened 2020-11-05
AST for keyword arguments


marcandre:
It just dawned on me that the AST for `foo(a: 42)` should probably be different than for `foo({a: 42})` in 3.0+, since they mean different things. Even in 2.7 they aren't completely the same (the first one issues a warning).

I imagine that replacing the type `:hash` with `:keyargs` or similar might work?

```
(send nil :foo
  (hash
    (pair
      (sym :a)
      (int 42)))
  (keyargs
    (pair
      (sym :a)
      (int 42))))
```

We'll need a legacy switch for this; RuboCop for example will probably not care much about this distinction and can rely on locations to determine if the hash is an actual literal or keyword arguments whenever need be. `unparser` will need a different AST though, right @palkan ?

palkan:
We should probably ping @mbj :) But yeah, Unparser relies on node types, so, it must be aware of it.

mbj:
Yes, unparser has the main constraint to *never* look at source maps "ever". Also unparser does have a logic to emit kwargs over hash literals. And this logic would break in 3.0.

Hence I support this change.

@marcandre Are you using unparser? If so: For what?

marcandre:
> @marcandre Are you using unparser?

I've not had that chance yet. I do like the idea of the gem though, call me a virtual fan 😆.

mbj:
> I've not had that chance yet. I do like the idea of the gem though, call me a virtual fan .

Hah, thx. So lets exploit the fan status:

I asked as unparser needs more testing, I get weird cases reported, like https://github.com/mbj/unparser/issues/189, and wish to find them all. If you wanted please install unparser and run `bundle exec unparser .` it should find and round trip (semantically under AST equivalency) all ruby files. Than reduce and report all reported errors.

iliabylich:
Oof, we need to find a good name.

`kwargs` doesn't really fit semantics of this new node (it conflicts with method arguments)

Is everyone ok with the name `keyargs`?

How about `kwparams`? That's exactly what this hash is about: it's a parameter (that is passed to a method call), not an argument (that is taken by a method)

iliabylich:
> We'll need a legacy switch for this

Agree, it affects stable `parser/ruby27`.

Also, this is possible, right?

```ruby
a(b: 1, **c, d: 2)
```

Should we wrap both pairs into a new node?

marcandre:
> `kwargs` doesn't really fit semantics of this new node (it conflicts with method arguments)

[...]

PR #769, opened 2020-11-25
+ builder: emit implicit hash passed to a method call as kwargs

iliabylich:
Closes https://github.com/whitequark/parser/issues/761.

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -130,6 +130,54 @@ module Parser
 
     class << self
       ##
+      # AST compatibility attribute; Starting from Ruby 2.7 keyword arguments
+      # of method calls that are passed explicitly as a hash (i.e. with curly braces)
+      # are treated as positional arguments and Ruby 2.7 emits a warning on such method
+      # call. Ruby 3.0 given an ArgumentError.
+      #
+      # If set to false (the default) the last hash argument is emitted as `hash`:
+      #
+      # ```
+      # (send nil :foo
+      #   (hash
+      #     (pair
+      #       (sym :bar)
+      #       (int 42))))
+      # ```
+      #
+      # If set to true it is emitted as `kwargs`:
+      #
+      # ```
+      # (send nil :foo
+      #   (kwargs
+      #     (pair
+      #       (sym :bar)
+      #       (int 42))))
+      # ```
+      #
+      # Note that `kwargs` node is just a replacement for `hash` argument,
+      # so if there's are multiple arguments (or a `kwsplat`) all of them
+      # are wrapped into `kwargs` instead of `hash`:
+      #
+      # ```
+      # (send nil :foo
+      #   (hash
+      #     (pair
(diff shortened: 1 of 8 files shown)


iliabylich:
@marcandre Could you take a look, too, please? (Kind find you in a "Reviewers" dropdown)

lib/parser/builders/default.rb
```diff
@@ -128,6 +128,54 @@ class << self
...
+      #       (int 42))))
+      # ```
+      #
+      # Note that `kwargs` node is just a replacement for `hash` argument,
+      # so if there's are multiple arguments (or a `kwsplat`) all of them
+      # are wrapped into `kwargs` instead of `hash`:
+      #
+      # ```
+      # (send nil :foo
+      #   (hash
```
marcandre, line 164: ^hash^kwargs

marcandre (commented):
The implementation surprises me. It was too difficult to emit the right node type from the get go?

iliabylich:
> It was too difficult to emit the right node type from the get go?

Yes and no, it's relatively simple. We rewrite either the last arg (if it's a hash) or the 2nd last (if it's a hash AND the last argument is a block). Rules are not trivial, and so the implementation takes 5 lines. The rest of the diff is about docs/tests.

(just in case) If you are talking about rewriting it in the `associate` method (that constructs a `hash` node) that's simply incorrect, not all hashes are kwargs. And there's no node for `parameters` (that'd be very helpful here)

marcandre:
I meant that looking in `ruby30.y` there are 4 rules that call `associate`; I was expecting those in `primary` and `aref_args` to remain unchanged, and those in  `opt_call_args` and in `call_args` to change to `associate_kwargs` (or whatever you'd want to call them). Would that not work?

iliabylich:
Yes, I had a similar idea, but there's a single exception - `indexasgn`:
```
             lhs: ...
                | primary_value tLBRACK2 opt_call_args rbracket
                    {
                      result = @builder.index_asgn(val[0], val[1], val[2], val[3])
                    }
```

Here `opt_call_args` can't be `kwargs` because this `foo[bar: 42] = baz` doesn't take kwargs at all.

Also by looking deeper at the grammar I've realized that it would affect `super(foo: 42)`, and we need to modify it too 😄 . I'll push an update

iliabylich:
Or maybe it's safer to rewrite it at the very beginning and un-rewrite in indexasgn. Yes, it seems to be the only exception, all other branches end up doing `call_method`/`super`/`yield`

iliabylich:
I give up on moving it to the grammar level. On 1.8 (huh) `aref_args` rule is a part of indexing (that does support kwargs) and regular array literals at the same time (that don't support it). Making 1.8 special would require us to rewrite a ton of tests to separate 1.8/SINCE_1_9 examples.

Added `kwargs` to `super` as mentioned above, going to merge it as is, feel free to propose better ideas.

Issue #772, opened 2020-12-02
Reduce dstr complexity on equivalent semantics but different AST.


mbj:
As discussed in other issues here, I'd like to make `parser` emit less "noise" on dynamic strings.

Lets discuss some of the cases that make it hard for me to maintain unparser. The first case is a class of strings that express the same semantics but the AST is unique for squiggly heredocs:

```ruby
<<~HEREDOC.
  #{}
  #{}
HEREDOC

"#{}
#{}
"
```

Both dynamic strings express the same runtime behavior. But the AST structure returned by `parser` is uniquely bound to the concrete syntax:

```
s(:begin,
  s(:dstr,
    s(:str, ""),
    s(:begin),
    s(:str, "\n"),
    s(:str, ""),
    s(:begin),
    s(:str, "\n")),
  s(:dstr,
    s(:begin),
    s(:str, "\n"),
    s(:begin),
    s(:str, "\n")))
<<~HEREDOC
~~~~~~~~~~... expression
s(:dstr,
  s(:str, ""),
  s(:begin),
  s(:str, "\n"),
  s(:str, ""),
  s(:begin),
  s(:str, "\n"))
<<~HEREDOC
~~~~~~~~~~ expression
  #{}
~~~~~... heredoc_body
HEREDOC
~~~~~~~ heredoc_end
s(:str, "")
  #{}
~~ expression
s(:begin)
  #{}
    ~ end
  ~~ begin
  ~~~ expression
s(:str, "\n")
  #{}
     ... expression
s(:str, "")
  #{}
~~ expression
s(:begin)
  #{}
    ~ end
  ~~ begin
  ~~~ expression
s(:str, "\n")
  #{}
     ... expression
s(:dstr,
  s(:begin),
  s(:str, "\n"),
  s(:begin),
  s(:str, "\n"))
"#{}
~ begin
~~~~... expression
"
~ end
s(:begin)
"#{}
   ~ end
 ~~ begin
 ~~~ expression
s(:str, "\n")
"#{}
    ... expression
s(:begin)
#{}
  ~ end
~~ begin
~~~ expression
s(:str, "\n")
#{}
   ... expression
```

Proposed change: Eliminate all `s(:str, "")` nodes inside `dstr` from the squiggly heredoc case. Note I do NOT want to change this case:

```
bundle exec ruby-parse --locate -e '"" ""'
s(:dstr,
  s(:str, ""),
  s(:str, ""))
"" ""
~~~~~ expression
s(:str, "")
[...]

PR #774, opened 2020-12-09
* do not emit truncated parts of squiggly heredoc

iliabylich:
Closes https://github.com/whitequark/parser/issues/772

So now
```
$ /bin/cat test.rb
<<~HEREDOC
  #{}
  #{}
HEREDOC

"#{}
#{}
"

"" ""
```

is parsed as
```
$ bin/ruby-parse --30 test.rb
(begin
  (dstr
    (begin)
    (str "\n")
    (begin)
    (str "\n"))
  (dstr
    (begin)
    (str "\n")
    (begin)
    (str "\n"))
  (dstr
    (str "")
    (str "")))
```

(two first cases are equivalent, the third case is not affected, we still emit all empty strings)

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -359,18 +359,23 @@ module Parser
       if !dedent_level.nil?
         dedenter = Lexer::Dedenter.new(dedent_level)
 
-        if node.type == :str
+        case node.type
+        when :str
           str = node.children.first
           dedenter.dedent(str)
-        elsif node.type == :dstr || node.type == :xstr
-          node.children.each do |str_node|
+        when :dstr, :xstr
+          children = node.children.map do |str_node|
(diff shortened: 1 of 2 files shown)


lib/parser/builders/default.rb
```diff
@@ -359,18 +359,23 @@ def dedent_string(node, dedent_level)
...
         dedenter = Lexer::Dedenter.new(dedent_level)
 
-        if node.type == :str
+        case node.type
+        when :str
           str = node.children.first
           dedenter.dedent(str)
-        elsif node.type == :dstr || node.type == :xstr
-          node.children.each do |str_node|
+        when :dstr, :xstr
```
mbj, line 366: This is so nice for me in the unparser case as I now can potentially collapse the xstr and dstr handling against each other.

Issue #775, opened 2020-12-13
ruby/ruby@88f3ce1


iliabylich:
https://github.com/ruby/ruby/commit/88f3ce12d32ffbef983b0950743c20253ea2d0c6

iliabylich:
I guess I have to rename `in_match` node:

```
foo in pattern
 => true/false

foo => pattern
 => nil/exception
```

Honestly I have absolutely no idea how to name them. @marcandre @mbj @palkan Any thoughts? `match_pattern/match_pattern_with_error`?

marcandre:
How about `match_pattern` for `=>`  and `match_pattern_p` (as in predicate) for `in`? I'd say the simplest name could be the raising version, as it is the same as `case-in`.

iliabylich:
Oh, that's a great name. Thanks, I'll use it (unless someone else comes with a better proposal 😄 )

palkan:
Another idea: `in_match` and `assoc_match`. I'm not sure AST should reflect the underlying behavior (I mean, `bool`).

marcandre:
> I'm not sure AST should reflect the underlying behavior

Good question. I'm not sure of the answer. Maybe behavior is actually more important than how they're written. If the behavior was not different, we wouldn't need a different type. Or if (god forbid 😅) there's one day a new way to write one or the other, say `~~` === `in`, it should use the same AST.

palkan:
> there's one day a new way to write one or the other, say ~~ === in, it should use the same AST.

Similarly, `true/false` could become `truthy/nil` in the future, so, making `bool` a bit confusing.

And I think a parser should care more about the _language_ and not the _implementation_. And the language has no knowledge of the possible return values.

iliabylich:
I'm going to go with `match_pattern`/`match_pattern_p`. Yes, currently `in` works as a predicate, `=>` throws an error, but I have a bad feeling that it's not a final version and that it can be changed someday. It's been announced that pattern matching is (mostly) stable, but it's a feature that is used by only a small percentage of users, and so yeah, I expect changes to be made in the near future. Semantics (I hope) will be the same, but specific constructions can be changed. If `=>` and `in` are swapped we will not change AST, it still would be `match_pattern`/`match_pattern_p` (but with different `node.keyword.source`: `in` vs `=>`)

@palkan thank you for your proposal

PR #777, opened 2020-12-16
+ ruby30.y: reintroduce `expr in pat`

iliabylich:
This commit tracks upstream commit ruby/ruby@88f3ce1.

Closes https://github.com/whitequark/parser/issues/775

So, if compatibility flag is disabled:

+ 2.7 emits `foo in bar` as `in_match`
+ 3.0.. emits `foo in bar` as `match_pattern_p`, `foo => bar` as `match_pattern`

If enabled:

+ 2.7 emits `foo in bar` as `match_pattern` (because it throws)
+ 3.0.. emits `foo in bar` as `match_pattern_p`, `foo => bar` as `match_pattern`

This flag doesn't affect 3.0 at all. I think it's incorrect to emit `foo in bar` as `in_match` on 3.0. On 2.7 this node has a different meaning (throw vs true/false), and so it must be a different node.

It's a bit chaotic with the flag being disabled, but
1. There's no way to misinterpret throwing and bool versions (even when multiple grammars are used).
2. It doesn't break 2.7 when disabled
3. Just enable the flag to have `match_pattern` vs `match_pattern_p` on all versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ below for explanation of `emit_*` calls):
     Parser::Builders::Default.emit_arg_inside_procarg0 = true
     Parser::Builders::Default.emit_forward_arg         = true
     Parser::Builders::Default.emit_kwargs              = true
+    Parser::Builders::Default.emit_match_pattern       = true
 
 Parse a chunk of code:
 
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -1910,15 +1910,56 @@ Format:
 
 ## Pattern matching
 
-### Using `in` modifier
+### Using `in` operator
 
-Format:
+Ruby 2.7 throws a `NoMatchingPatternError` for `foo in bar` if given value doesn't match pattern.
+
+Format when `emit_match_pattern` compatibility attribute is disabled (the default):
 
 ~~~
 (in-match
   (int 1)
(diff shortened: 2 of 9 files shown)


iliabylich:
@marcandre @mbj @palkan Could you take a look please? Does it work well for you?

mbj:
@iliabylich I'm okay with your reasoning, but have to admit my mental model for pattern matches is incomplete at this point. 

Mostly because its a big "self closed" set of features I'll not be able to use in my daytime Ruby for the next 2 years, given I've to support versions without this extension.


Once I get to make unparser support for latest parser changes for Ruby 3 I can say more.

marcandre:
LGTM

That being said, I'm wondering if it isn't going too far, being compatible with an experimental feature. I.e. if Ruby itself reserves the right to change the future about a language construct, it would be hard to fault `parser` for doing the same. In this case it doesn't cost much I guess...

iliabylich:
> That being said, I'm wondering if it isn't going too far, being compatible with an experimental feature.

Is it valid to say that it's experimental for `parser` too and that we can also change anytime we want 😄 ? Just kidding. Some people need it, and so I think everything should be backported, even if it's not stable yet. And common rules of "don't break depending libraries" should work here too.

PR #781, opened 2020-12-25
ruby-3.0 is released

krushi123:
ruby 3.0.0 is released and no need to check for dev version

### DIFF
--- a/lib/parser/current.rb
+++ b/lib/parser/current.rb
@@ -84,7 +84,7 @@ module Parser
     CurrentRuby = Ruby27
 
   when /^3\.0\./
-    current_version = '3.0.0-dev'
+    current_version = '3.0.0'
     if RUBY_VERSION != current_version
       warn_syntax_deviation 'parser/ruby30', current_version
     end


lib/parser/current.rb
```diff
@@ -84,7 +84,7 @@ def warn_syntax_deviation(feature, version)
     CurrentRuby = Ruby27
 
   when /^3\.0\./
```
koic, line 86: It looks like the `else` branch also needs to be updated.

```diff
  when /^3\.0\./
-   current_version = '3.0.0-dev'
+   current_version = '3.0.0'
    if RUBY_VERSION != current_version
      warn_syntax_deviation 'parser/ruby30', current_version
    end

    require 'parser/ruby30'
    CurrentRuby = Ruby30

  else # :nocov:
    # Keep this in sync with released Ruby.
-   warn_syntax_deviation 'parser/ruby27', '2.7.x'
-   require 'parser/ruby27'
-   CurrentRuby = Ruby27
+   warn_syntax_deviation 'parser/ruby30', '3.0.x'
+   require 'parser/ruby30'
+   CurrentRuby = Ruby30
  end
```

cf: https://github.com/whitequark/parser/commit/47111d191f6e989db562d2a08e92fa9da87c4435

iliabylich:
@krushi123 Thanks!

@koic I'll fix the rest

Issue #785, opened 2021-01-13
highlight line should handle case where size == 0


rubys:
Reproduction instructions:

```
irb(main):003:0> Parser::CurrentRuby.parse('foo++')
(string):1:6: error: unexpected token $end
(string):1: foo++
(string):1:
Traceback (most recent call last):
...
```

Note the blank line where there should be either a caret or a tilde indicating the spot of the error.

The code in question is in [Parser::Diagnostic::render_line](https://github.com/whitequark/parser/blob/6337d7bf676f66d80e43bd9d33dc17659f8af7f3/lib/parser/diagnostic.rb#L124-L128).

I'd submit a pull request, but it isn't clear to me whether the code should be putting out a caret or a tilde in this case.  I have verified that `column_range.begin_pos` is the spot immediately after the end of the line in this case.

iliabylich:
`$end` is a special token that has `begin = end = EOF`, this is why it has zero source range.

Some other languages also don't render it:
```
$ node -e 'foo**'
[eval]:1
foo**


SyntaxError: Unexpected end of input
    at new Script (vm.js:88:7)
    at createScript (vm.js:263:10)
    at Object.runInThisContext (vm.js:311:10)
    at Object.<anonymous> ([eval]-wrapper:10:26)
    at Module._compile (internal/modules/cjs/loader.js:1138:30)
    at evalScript (internal/process/execution.js:94:25)
    at internal/main/eval_string.js:23:3

$ perl -e '++'
syntax error at -e line 1, at EOF
Execution of -e aborted due to compilation errors.
```

but some languages do:
```
$ python -c 'foo++'
  File "<string>", line 1
    foo++
        ^
SyntaxError: invalid syntax

$ echo "class ++" > test.java && javac test.java
test.java:1: error: <identifier> expected
class ++
     ^
test.java:1: error: reached end of file while parsing
class ++
        ^
2 errors
```

I think it's ok to change it. A few  options:

1. we could emit `$eof` token with a EOF...EOF+1 range, but I can't estimate implications of this change. Potentially it can break libraries that depend on `parser`.
2. we could check for `error_token_id == 0` in `Parser::Base#on_error` and if it's zero add +1 to `location.end`, but I guess it's not a part of the Racc stable API, i.e. it can change anytime.
3. we could emit a different error message for `$eof` token, but it seems to be a breaking change.

[...]

PR #786, opened 2021-01-24
emit $eof token with eof...eof+1 range to properly render diagnostic

iliabylich:
Closes https://github.com/whitequark/parser/issues/785


iliabylich:
Oof, looks like it breaks something in rubocop

Issue #728, opened 2020-07-30
Summarize `inspect`


marcandre:
The current `inspect` methods for most classes are not very usable:

```ruby
  @corrector=
   #<Parser::Source::TreeRewriter:0x00007fd280304148
    @action_root=
     #<Parser::Source::TreeRewriter::Action:0x00007fd280d73d90
      @children=[],
      @enforcer=
       #<Method: RuboCop::Cop::Corrector(Parser::Source::TreeRewriter)#enforce_policy(event) /Users/mal/parser/lib/parser/source/tree_rewriter.rb:384>,
      @insert_after="",
      @insert_before="",
      @range=#<Parser::Source::Range /Users/mal/rubocop/lib/rubocop/cop/lint/redundant_cop_disable_directive.rb -1...10139>,
      @replacement=nil>,
    @diagnostics=
     #<Parser::Diagnostic::Engine:0x00007fd2803040a8
      @all_errors_are_fatal=false,
      @consumer=#<Proc:0x00007fd280d73b88 /Users/mal/rubocop/lib/rubocop/cop/corrector.rb:26 (lambda)>,
      @ignore_warnings=false>,
    @enforcer=
     #<Method: RuboCop::Cop::Corrector(Parser::Source::TreeRewriter)#enforce_policy(event) /Users/mal/parser/lib/parser/source/tree_rewriter.rb:384>,
    @in_transaction=false,
    @policy={:crossing_deletions=>:accept, :different_replacements=>:raise, :swallowed_insertions=>:raise},
    @source_buffer=
     #<Parser::Source::Buffer:0x00007fd284028358
      @column_for_position=
       {0=>0,
        31=>0,
        102=>0,
        166=>0,
        268=>6,
        343=>6,
        421=>6,
        492=>6,
        571=>6,
        646=>6,
        722=>6,
   #...
    @line_for_position= {
   # ...
```

`Parser::Source::Range` has nice `inspect` and `to_s`, but `Parser::Source::Buffer` and `Parser::Source::TreeRewriter` (at a minimum) would benefit from a summarized version...

I hope to propose a PR one day...

PR #788, opened 2021-03-07
+ `Source::TreeRewriter#inspect` [#728]

marcandre:
This provide a short `inspect` for TreeRewriter. I thought it might be useful to add a summarized summary of the actions of the rewriter, either "empty", or a mix of additions (`+"insert me"@42`), deletions (`-55...60`) and replacements (`^"replacement"@77...88), but I can change of remove if that's too cryptic. The main goal was to avoid an `inspect` that was really very long.

### DIFF
--- a/lib/parser/source/tree_rewriter.rb
+++ b/lib/parser/source/tree_rewriter.rb
@@ -330,6 +330,11 @@ module Parser
         @in_transaction
       end
 
+      # :nodoc:
+      def inspect
+        "#<#{self.class} #{source_buffer.name}: #{action_summary}>"
+      end
+
       ##
       # @api private
       # @deprecated Use insert_after or wrap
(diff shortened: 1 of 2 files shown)


lib/parser/source/tree_rewriter.rb
```diff
@@ -361,6 +366,28 @@ def insert_after_multi(range, text)
...
+        replacements = as_replacements
+        case replacements.size
+        when 0 then return 'empty'
+        when 1..3 then #ok
+        else
+          replacements = replacements.first(3)
+          suffix = '…'
+        end
+        parts = replacements.map do |(range, str)|
+          if str.empty?
```
iliabylich, line 379: Not sure if I understand what happens in this if/elsif logic? Could you explain it please?

marcandre: The `replacements` method returns a range and the string that replaces it. The if/elsif/ logic breaks these into deletions, additions, and replacements.

iliabylich: got it, thanks. is there a better way to identify these replacements?

iliabylich: actually, I think it's enough to have these comments, 👍  if there's no other way to do it (without rewriting)

iliabylich (approved):

lib/parser/source/tree_rewriter.rb
marcandre: Sorry, not sure what you mean

iliabylich: the way how you jump into if/else branches is not very clean and explicit. Technically it's based on undocumented behavior, and it can break once insertion has a non-empty `range` (for absolutely any reason). I was thinking about using some better check to identify replacements.

iliabylich:
@marcandre thanks

lib/parser/source/tree_rewriter.rb
marcandre: Internally, these are all treated as replacements. It's simply that humans will see a replacement of a range corresponding to "hello" with an empty string as a deletion. But technically there is no distinction between `delete(range)` and `replace(range, '')`. Same idea of straight insertion vs  replacement.
In short: this can not break, because a straight insertion is always computed as a string replacing an empty range, by definition.

marcandre:
My pleasure 👍
Thanks for the near immediate feedback!

PR #791, opened 2021-04-05
* Bump maintenance branches to 3.0.1, 2.7.3, 2.6.7, and 2.5.9

koic:
These Rubies has been released.

- https://www.ruby-lang.org/ja/news/2021/04/05/ruby-3-0-1-released/
- https://www.ruby-lang.org/en/news/2021/04/05/ruby-2-7-3-released/
- https://www.ruby-lang.org/en/news/2021/04/05/ruby-2-6-7-released/
- https://www.ruby-lang.org/en/news/2021/04/05/ruby-2-5-9-released/

## Ruby 3.0 branch

Bump 3.0 branch from 3.0.0 to 3.0.1
https://github.com/ruby/ruby/compare/v3_0_0...v3_0_1

There seems to be no change to the syntax.

## Ruby 2.7 branch

Bump 2.7 branch from 2.7.2 to 2.7.3
https://github.com/ruby/ruby/compare/v2_7_2...v2_7_3

There is the leading arguments support to arguments forwarding change to the syntax.
Thas PR is backport #712 to Ruby 2.7 branch.

cf. https://github.com/ruby/ruby/commit/27fca66207f2c35f2f44f6a7cbbe6fd153546082

## Ruby 2.6 branch

Bump 2.6 branch from 2.6.6 to 2.6.7
https://github.com/ruby/ruby/compare/v2_6_6...v2_6_7

There seems to be no change to the syntax.

## Ruby 2.5 branch

Bump 2.5 branch from 2.5.8 to 2.5.9
https://github.com/ruby/ruby/compare/v2_5_8...v2_5_9

There seems to be no change to the syntax.

PR #792, opened 2021-04-06
+ ruby31.y: branch parser

koic:
ruby31.y is based on ruby30.y as Ruby 3.0.1. And Ruby 3.0.0 and 3.0.1 are the same (#791) .


lib/parser/ruby31.y
```diff
@@ -0,0 +1,3077 @@
...
+end
+
+---- header
+
+require 'parser'
+
+---- inner
+
+  def version
+    30
```
iliabylich, line 3066: This line should be updated too

lib/parser/current.rb
```diff
@@ -92,6 +92,15 @@ def warn_syntax_deviation(feature, version)
     require 'parser/ruby30'
     CurrentRuby = Ruby30
 
+  when /^3\.1\./
+    current_version = '3.1.0'
```
iliabylich, line 96: Also, could you change it to `3.1.0-dev` please?

koic:
@iliabylich Thanks for your review! I updated this PR.

iliabylich:
Thanks!

PR #794, opened 2021-04-13
vertical bar is not a valid character in an lvar name

rubys:
I just happened to notice this via visual inspection (this regular expression happens to be problematic for Opal, see https://github.com/opal/opal/issues/2195 for workaround)


iliabylich:
@rubys Thanks and sorry for delay

Issue #795, opened 2021-04-20
Fix Parser::SyntaxError error with endless method for object equality


bkuhlmann:
## Overview

Discovered when attempting to write an endless method for object equality that I'd get a syntax error. Here's a snippet for illustration:

``` ruby
class Example
  def ==(other) = hash == other.hash
end
```

In this case, I'm using [Reek](https://github.com/troessner/reek) to illustrate the stack dump in case it's helpful:

<details>
  <summary>Reek Stack Dump</summary>

``` 
Inspecting 1 file(s):
        !!!
        Source '/Users/bkuhlmann/Engineering/Misc/snippet.rb' cannot be processed by Reek due to a syntax error in the source file.

        This is a problem that is outside of Reek's scope and should be fixed by you, the
        user, in order for Reek being able to continue.

        Things you can try:
        - Check the syntax of the problematic file
        - If the file is not in fact a Ruby file, exclude it in your .reek.yml file

        Exception message:

        #<Parser::SyntaxError: setter method cannot be defined in an endless method definition>

        Original backtrace:

        /Users/bkuhlmann/.gem/ruby/3.0.1/gems/parser-3.0.1.0/lib/parser/diagnostic/engine.rb:72:in `process'
	/Users/bkuhlmann/.gem/ruby/3.0.1/gems/parser-3.0.1.0/lib/parser/base.rb:274:in `diagnostic'
	/Users/bkuhlmann/.gem/ruby/3.0.1/gems/parser-3.0.1.0/lib/parser/ruby30.rb:27:in `endless_method_name'
	/Users/bkuhlmann/.gem/ruby/3.0.1/gems/parser-3.0.1.0/lib/parser/ruby30.rb:5156:in `_reduce_249'
	(eval):3:in `_racc_do_parse_c'
	(eval):3:in `do_parse'
	/Users/bkuhlmann/.gem/ruby/3.0.1/gems/parser-3.0.1.0/lib/parser/base.rb:190:in `parse'
	/Users/bkuhlmann/.gem/ruby/3.0.1/gems/parser-3.0.1.0/lib/parser/base.rb:207:in `parse_with_comments'
	/Users/bkuhlmann/.gem/ruby/3.0.1/gems/reek-6.0.3/lib/reek/source/source_code.rb:117:in `parse'
	/Users/bkuhlmann/.gem/ruby/3.0.1/gems/reek-6.0.3/lib/reek/source/source_code.rb:53:in `syntax_tree'
[...]

PR #796, opened 2021-04-20
- fix a false positive for endless method definition

koic:
Fixes: #795.

This PR fixes false positives that comparison methods recognizes as a setter method.

### DIFF
--- a/lib/parser/ruby30.y
+++ b/lib/parser/ruby30.y
@@ -3071,7 +3071,7 @@ require 'parser'
   end
 
   def endless_method_name(name_t)
-    if name_t[0].end_with?('=')
+    if !%w[=== == != <= >=].include?(name_t[0]) && name_t[0].end_with?('=')
       diagnostic :error, :endless_setter, nil, name_t
     end
   end
--- a/test/test_parser.rb
+++ b/test/test_parser.rb
@@ -10107,6 +10107,19 @@ class TestParser < Minitest::Test
       SINCE_3_0)
   end
 
+  def test_endless_comparison_method
+    %i[=== == != <= >= !=].each do |method_name|
+      assert_parses(
+        s(:def, method_name,
+          s(:args,
+            s(:arg, :other)),
+          s(:send, nil, :do_something)),
+        %Q{def #{method_name}(other) = do_something},
+        %q{},
+        SINCE_3_0)
+    end
+  end
+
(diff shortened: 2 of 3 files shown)


lib/parser/ruby30.y
```diff
@@ -3071,7 +3071,7 @@ require 'parser'
   end
 
   def endless_method_name(name_t)
-    if name_t[0].end_with?('=')
+    if !%w[== === >= <= !=].include?(name_t[0]) && name_t[0].end_with?('=')
```
iliabylich, line 3074: `[]=` should also be in this list (and who knows what else):
``` ruby
> ObjectSpace.each_object(Class).flat_map(&:instance_methods).uniq.select { |mid| mid.to_s.end_with?('=') && !(mid.to_s[-2] =~ /\w/) }
 => [:===, :==, :!=, :<=, :>=, :[]=]
```

test/test_parser.rb
```diff
@@ -10107,6 +10107,73 @@ def test_endless_setter
       SINCE_3_0)
   end
 
+  def test_endless_comparison_method
+    assert_parses(
```
iliabylich, line 10111: could you rewrite it into a loop that iterates over method names please? Also, source maps are redundant here.

lib/parser/ruby30.y
iliabylich: wait, or is it a setter? 😄

koic: `[]=` will also result in the error on MRI :-)

```ruby
% ruby -ve "def []=(index, value) = 42"
ruby 3.0.1p64 (2021-04-05 revision 0fb782ee38) [x86_64-darwin19]
-e:1: setter method cannot be defined in an endless method definition
def []=(index, value) = 42
```

```ruby
% ruby -ve "def []=(index, value) = 42"
ruby 3.1.0dev (2021-04-06T07:03:20Z master 31ba817887) [x86_64-darwin19]
-e:1: setter method cannot be defined in an endless method definition
def []=(index, value) = 42
```

koic: Oh, you're right. It's a setter, so it's an error and it's expected. I updated this PR again.

iliabylich:
@koic Thanks!

koic:
Thanks too!

PR #801, opened 2021-05-13
+ ruby31.y: allow "command" syntax in endless method definition

koic:
Fixes https://github.com/whitequark/parser/issues/799.

### DIFF
--- a/lib/parser/ruby31.y
+++ b/lib/parser/ruby31.y
@@ -261,6 +261,74 @@ rule
                                     val[0], val[1], val[2]),
                                   val[3], val[4])
                     }
+                | defn_head f_opt_paren_args tEQL command
+                    {
+                      _def_t, name_t = val[0]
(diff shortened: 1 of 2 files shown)


iliabylich:
@koic could you also backport tests for `command kRESCUE_MOD arg` cases please? I'm talking about [these guys](https://github.com/ruby/ruby/commit/31794d2e733e081e4e221f27aff6380393981681#diff-88c80426a639b67117eb2cb10026a95fd1982a22d399442d9303a330e537b34bR1473-R1474).

Also this new syntax is a pure hack, `def foo = bar baz` is a command assignment that is reduced in a different way 🔥

koic:
Thank you for your review. I've added missing backporting tests.

iliabylich (approved):
👍

iliabylich:
@koic thanks!

PR #818, opened 2021-09-21
+ ruby31.y: Allow value omission in Hash literals

### LINKED ISSUE #817: ruby/ruby@c60dbcd
https://github.com/ruby/ruby/commit/c60dbcd1c55cd77a24c41d5e1a9555622be8b2b8

skryukov:
Closes #817.

This commit tracks upstream commit https://github.com/ruby/ruby/commit/c60dbcd

### DIFF
--- a/test/test_parser.rb
+++ b/test/test_parser.rb
@@ -10054,6 +10054,22 @@ class TestParser < Minitest::Test
       SINCE_3_1)
   end
 
+  def test_value_omission
+    assert_parses(
+      s(:hash,
+        s(:pair, s(:sym, :a), s(:send, nil, :a)),
+        s(:pair, s(:sym, :b), s(:send, nil, :b))),
+      %q{{a:, b:}},
+      %q{^ begin
+        |       ^ end
+        |  ^ operator (pair)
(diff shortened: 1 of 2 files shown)


test/test_parser.rb
```diff
@@ -10054,6 +10054,15 @@ def test_private_endless_method_command_syntax
       SINCE_3_1)
   end
 
+  def test_value_omission
+    assert_parses(
+      s(:hash,
+        s(:pair, s(:sym, :a), s(:send, nil, :a)),
+        s(:pair, s(:sym, :b), s(:send, nil, :b))),
+      %q{{a:, b:}},
```
iliabylich, line 10062: There's a `source_maps` argument that has to be passed explicitly (`%q{}`).
Also, could you please add source maps checks for both key and value?

iliabylich (changes requested):
Thanks! Initially I was thinking about a new node type (as always 🤦 ), but I like this approach a lot. Once the test is fixed I'd be glad to merge it.

iliabylich (approved):
👍

iliabylich:
@skryukov Thanks!

PR #820, opened 2021-09-24
- ruby31.y: handle local variables as hash labels with omitted values

iliabylich:
There's a new rule in 3.1 that allows hash value to be omitted if it's equal to key:
```sh
$ ./miniruby -ve 'p({ rand: })'
ruby 3.1.0dev (2021-09-24T13:56:38Z master 40a65030e5) [x86_64-darwin20]
{:rand=>0.6152290626750991}
```

We backported it in https://github.com/whitequark/parser/pull/818, but there's one case that we missed - local variables can also be used:
```sh
./miniruby -ve 'foo = 1; p({ foo: })'
ruby 3.1.0dev (2021-09-24T13:56:38Z master 40a65030e5) [x86_64-darwin20]
{:foo=>1}
```

This PR amends this logic.

And of course, there are implications like circular arguments validation that was missing:
```
$ /bin/cat ../ruby/test.rb
def m(a = { a: })
end

$ bin/ruby-parse --31 ../ruby/test.rb
../ruby/test.rb:1:13: error: circular argument reference a
../ruby/test.rb:1: def m(a = { a: })
../ruby/test.rb:1:             ^

$ ../ruby/miniruby -cv ../ruby/test.rb
ruby 3.1.0dev (2021-09-24T13:56:38Z master 40a65030e5) [x86_64-darwin20]
../ruby/test.rb:1: circular argument reference - a
```

@marcandre @koic @mbj @palkan Can I get a review of this API please? Are there any issues with multiple nodes having the same source location?

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -518,6 +518,20 @@ module Parser
       n(:pair, [ key, value ], pair_map)
     end
 
+    def pair_label(key_t)
+      key_l = loc(key_t)
+      value_l = key_l.adjust(end_pos: -1)
+
+      label = value(key_t)
+      value =
+        if label =~ /\A[[:lower:]]/
+          n(:ident, [ label.to_sym ], Source::Map::Variable.new(value_l))
+        else
+          n(:const, [ nil, label.to_sym ], Source::Map::Constant.new(nil, value_l, value_l))
+        end
+      pair_keyword(key_t, accessible(value))
+    end
+
     def kwsplat(dstar_t, arg)
       n(:kwsplat, [ arg ],
         unary_op_map(dstar_t, arg))
@@ -608,18 +622,28 @@ module Parser
       when :ident
         name, = *node
 
-        if @parser.static_env.declared?(name)
-          if name.to_s == parser.current_arg_stack.top
-            diagnostic :error, :circular_argument_reference,
-                       { :var_name => name.to_s }, node.loc.expression
-          end
+        if %w[? !].any? { |c| name.to_s.end_with?(c) }
+          diagnostic :error, :invalid_id_to_get,
+                     { :identifier => name.to_s }, node.loc.expression
+        end
 
-          node.updated(:lvar)
-        else
-          name, = *node
-          n(:send, [ nil, name ],
+        # Numbered parameters are not declared anywhere,
+        # so they take precedence over method calls in numblock contexts
+        if @parser.version >= 27 && @parser.try_declare_numparam(node)
+          return node.updated(:lvar)
+        end
+
+        unless @parser.static_env.declared?(name)
+          return n(:send, [ nil, name ],
             var_send_map(node))
         end
+
+        if name.to_s == parser.current_arg_stack.top
+          diagnostic :error, :circular_argument_reference,
+                     { :var_name => name.to_s }, node.loc.expression
+        end
+
+        node.updated(:lvar)
 
       else
         node
@@ -682,6 +706,17 @@ module Parser
         @parser.static_env.declare(name)
 
         node.updated(:lvasgn)
+
+      when :match_var
+        name, = *node
+
+        var_name = node.children[0].to_s
+        name_loc = node.loc.expression
+
+        check_assignment_to_numparam(var_name, name_loc)
+        check_reserved_for_numparam(var_name, name_loc)
+
+        node
 
       when :nil, :self, :true, :false,
            :__FILE__, :__LINE__, :__ENCODING__
--- a/doc/AST_FORMAT.md
+++ b/doc/AST_FORMAT.md
@@ -266,6 +266,42 @@ Format:
  ~~~~~~~~~~ expression (pair)
 ~~~
 
+#### With local variable
+
+Format:
+
+~~~
+(pair (sym :foo) (lvar :foo))
+"{foo:}"
+     ^ operator (pair)
+  ~~~ expression (sym)
+  ~~~ expression (lvar)
+~~~
+
+#### With constant
+
+Format:
+
+~~~
+(pair (sym :foo) (const nil :foo))
+"{FOO:}"
+     ^ operator (pair)
+  ~~~ expression (const)
+  ~~~ expression (lvar)
+~~~
+
+#### With method call
+
+Format:
+
+~~~
+(pair (sym :puts) (send nil :puts))
+"{puts:}"
+      ^ operator (pair)
+  ~~~~ expression (sym)
+  ~~~~ expression (send)
+~~~
+
 #### Plain
 
 Format:
--- a/lib/parser/messages.rb
+++ b/lib/parser/messages.rb
@@ -76,6 +76,7 @@ module Parser
     :duplicate_variable_name      => 'duplicate variable name %{name}',
     :duplicate_pattern_key        => 'duplicate hash pattern key %{name}',
     :endless_setter               => 'setter method cannot be defined in an endless method definition',
+    :invalid_id_to_get            => 'identifier %{identifier} is not valid to get',
 
     # Parser warnings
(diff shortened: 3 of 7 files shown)


palkan:
> Are there any issues with multiple nodes having the same source location?

As soon as it's safe to use the fact that source locations are equal for both key and value nodes to detect _implicit_ hashes, that would work for me.

(In [my implementation](https://github.com/ruby-next/parser/commit/31b45cfab93a3fbe32866433329f36387df0ccb3), I used a separate node type; but I guess that's not an option for Parser 🙂).

marcandre:
Tricky.
Was it considered to have the location be `nil` instead?
The reservation I have with reusing the location (instead of `nil` makes it so close to the normal form that it makes it very difficult to distinguish them.
Writing a method to know if a particular `lvar` is of such a form, one has to use the parent (not actually available), check that it is a `pair` and that the `lvar` in question is the second child, and finally compare the location with that of the first child.
This non-locality of the information doesn't seem ideal. I don't feel strongly about this, but I would intuitively think that the `sym` should have a location but the `lvar`/`send`/... shouldn't.

marcandre:
To comment further, in RuboCop we do have access to the `parent`, so whichever way this ends up, we can implement a method to check this (albeit ugly). Only risk factor is that a cop would rewrite the source without realizing this new case, and instead of crashing, and the resulting generated code could be incompatible. Maybe @koic or @dvandersluis can think of other impacts for RuboCop?

iliabylich:
@marcandre There's a rule that all nodes have some location (at least `expression_l`), I'm pretty sure that changing it would have bigger impact than sharing location between two nodes.

There's always an alternative - creating a new node type 😄

marcandre:
Ah ah, new mode type sounds really bad indeed. 😬

Empty expression (just after the colon) would be an alternative.

iliabylich:
> Empty expression (just after the colon) would be an alternative.

That also would be a breaking change (node without `expr_l`). Also I don't know how I feel about "dummy" nodes in general, mostly because it splits nodes into two groups and it feels like this separation sits on top of node-type groups (i.e. this way we get two groups of types, `dummy(typeA, typeB, ...)` and `real(typeC, typeD, ...)`). I'd like to keep it flat.

I did quick research in other languages that support shorthand object field syntax.

**JavaScript (Babel/parser)** - [the doc](https://github.com/babel/babel/blob/main/packages/babel-parser/ast/spec.md#objectexpression) says that `Object` node has a list of members where each member is a sum type of `ObjectProperty | ObjectMethod | SpreadElement` where `ObjectProperty` has

```
key: Expression
shorthand: boolean;
value: Expression
```

The doc has `T | null` in many places, so I guess `key` and `value` are non-nullable, which means they share location. The flag is just a nice addition.

**Rust (syn crate)** - AST has a type called [FieldValue](https://docs.rs/syn/1.0.77/syn/struct.FieldValue.html) that represents `field` in `Foo { field }` struct "instantiation", and it has the following fields:

```
pub member: Member,
pub colon_token: Option<Colon>,
pub expr: Expr,
```

[Here's how it's used internally](https://github.com/dtolnay/syn/blob/d195790a8c3d5797a7c0230cb6c5b671643cc943/src/expr.rs#L2614) - `if` branch handles explicit (i.e. `Foo { field: value }`) key-value pairs, `else` branch handles `Foo { field }` syntax, and `syn` also copies location of the `key` to `value`.

At least we are not inventing anything new in this PR.

Another alternative that I see (and dislike a lot but still would like it be mentioned) is making `pair.value` nullable. I suppose that'd be a huge change for `parser` users, right?

I'm going to keep this PR open for several days, maybe someone can suggest a better approach.

pocke:
FYI, I found two differences between MRI and parser gem other than lvars. I'll create a new issue if you need it.

First, MRI accepts a constant as key/value but parser gem doesn't.

```bash
$ ruby -ve 'p({String:})' 
ruby 3.1.0dev (2021-09-24T23:41:02Z master 39a6bf5513) [x86_64-linux]
{:String=>String}

$ bin/ruby-parse --31 -e 'p({String:})' 
(send nil :p
  (hash
    (pair
      (sym :String)
      (send nil :String))))
```



Second, MRI rejects a label with `!` or `?`, but parser gem doesn't.

```bash
$ ruby -e p({foo!:})
-e:1: identifier foo! is not valid to get

$ ruby -e p({foo?:})
-e:1: identifier foo? is not valid to get

$ bin/ruby-parse --31 -e 'p({foo!:})'
(send nil :p
  (hash
    (pair
      (sym :foo!)
      (send nil :foo!))))

$ bin/ruby-parse --31 -e 'p({foo?:})'
(send nil :p
  (hash
    (pair
      (sym :foo?)
      (send nil :foo?))))
```

lib/parser/builders/default.rb
```diff
@@ -518,6 +518,15 @@ def pair_quoted(begin_t, parts, end_t, value)
       n(:pair, [ key, value ], pair_map)
     end
 
+    def pair_label(key_t)
+      key_l = loc(key_t)
+      value_l = key_l.adjust(end_pos: -1)
+
+      ident = n(:ident, [ value(key_t).to_sym ], Source::Map::Variable.new(value_l))
+      value = accessible(ident)
+      result = pair_keyword(key_t, value)
```
pocke, line 527:
```suggestion
      pair_keyword(key_t, value)
```

I think this assignment should be removed because it displays an unused variable warning under `RUBYOPT=-w` environment.

mbj:
> Can I get a review of this API please? Are there any issues with multiple nodes having the same source location?

Unparser by definition does not use source locations and instead tries its best to generate concrete syntax "just" from the structure of the AST nodes. So the source location aspect does not concern me - at this point.

iliabylich:
ok, time to finish it 😄 

I've spent some time investigating what `gettable` function does in MRI and it's quite tricky because of how obfuscated it is:
1. MRI calls `TOK_INTERN` macro (or a wrapper/helper function `tokenize_ident`) to attach `id` to a token. `ID` is basically a Ruby `Symbol`.
2. this macro calls `rb_intern3` to construct an `ID` that is internally represented as a tagged pointer (i.e. a pointer or a value depending on leading bytes that is `&`-checked every time it's dereferenced, constant known-at-compile-time symbols can be compared by value). This whole process is performed only for tokens that have variadic values like `tCONSTANT`, `tIDENTIFIER`, `tLABEL` and so on. Tokens like `kCLASS` don't need it because their value are constant and can be inferred based on a token type that is simply a flat number.
3.  `rb_intern3` is defined in `symbol.c` and it calls `intern_str` -> `rb_str_symname_type` -> `rb_enc_symname_type` (and also -> `enc_synmane_type_leading_chars`) that attaches additional bitflags:
	+ `ID_GLOBAL` for `$`-prefixed values
    + or `ID_INSTANCE` for `@`-prefixed values
    + or `ID_CLASS` for `@@`-prefixed values
    + or `ID_CONST` for  values starting with capital letter
    + or `ID_LOCAL` otherwise
4. Later MRI uses these flags to check what's the type of the `token.val`

We do the same thing in many places in `Builder`, I think it makes sense to mirror these checks as separate methods to simplify backporting in the future.

This `gettable` methods is called `Builder::assignable` in `parser`, so I'm going to add missing checks there.

Also I've noticed that this method checks for assignment to numparams in numblocks and errors if there are any, but we check it directly in `.y` file. I've found a bug:

```ruby
proc { 1 in _1 }
``` 

MRI rejects this code but we accept it. It will be fixed by moving numparam check to the builder (wrapped with `@version` check, assignment to numparam is valid for Ruby < 27)

iliabylich:
@marcandre @koic @mbj @palkan @pocke Can I get a review please? Do you see any missing tests cases?

iliabylich:
Merging to unblock myself, feel free to post thoughts/requests here.

PR #822, opened 2021-10-08
- lexer.rl: Fix handling of beginless ranges at start of line

### LINKED ISSUE #814: Beginless ranges and line breaks
This code seems to be parsed incorrectly:

```ruby
def bar
  foo = 2
  ..foo
end
```

Calling `bar` returns `..2`, but Parser parser this as:
```
s(:def, :bar,
  s(:args),
  s(:lvasgn, :foo,
    s(:irange,
      s(:int, 2),
      s(:lvar, :foo))))
```

I would expect this to be parser identically to
```ruby
def bar
  foo = 2;
  ..foo
end
```

This version is parsed by Parser as:
```
s(:def, :bar,
  s(:args),
  s(:begin,
    s(:lvasgn, :foo,
      s(:int, 2)),
    s(:irange, nil,
      s(:lvar, :foo))))
```

mvz:
This fixes an issue where beginless ranges at the start of a line would be interpreted as continuing an expression starting on the previous line. Unlike leading dots for method calls, `..` and `...` at the start of a line do not continue an expression from the previous line.

Fixes #814.

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -2513,6 +2513,28 @@ class Parser::Lexer
         end
       };
 
+      c_space* '..'
+      => {
+        emit(:tNL, nil, @newline_s, @newline_s + 1)
+        if @version < 27
+          fhold; fnext line_begin; fbreak;
+        else
+          emit(:tBDOT2)
+          fnext expr_beg; fbreak;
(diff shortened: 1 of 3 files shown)


iliabylich:
Thanks!

Issue #830, opened 2021-11-26
`error: end pattern with unmatched parenthesis: /(/` occurs when using Ruby 3.1 parser


koic:
An error occurs when parsing `/\(/` regular expression with Ruby 3.1 parser.

## Reproduction context

```console
% ruby-parse -v
ruby-parse based on parser version 3.0.3.0

% ruby -v
ruby 3.1.0dev (2021-11-26T13:32:26Z master dff8d12226) [x86_64-darwin19]
```

The same error occurs in the stable Ruby versions. (e.g. Ruby 3.0.3)

## Expected behavior (using Ruby 3.0 parser)

No errors.

```console
% ruby-parse --30 -e '/\(/'
(regexp
  (str "\\(")
  (regopt))
```

## Actual behavior (using Ruby 3.1 parser)

`error: end pattern with unmatched parenthesis: /(/` is raised.

```console
% ruby-parse --31 -e '/\(/'
(fragment:0):1:1: error: end pattern with unmatched parenthesis: /(/
(fragment:0):1: /\(/
(fragment:0):1: ^~~~
```

iliabylich:
oof, thanks a bug in https://github.com/whitequark/parser/commit/24d2f682c56319266602a3b3115e087fd81d667e.

right, there are still "escape sequences" that should be treated without escaping (i.e. as a sequence of independent characters).

I'll take a look tomorrow.

PR #831, opened 2021-11-29
fix escaping of ctrl- and meta-chars in regexes

iliabylich:
the bug was introduced in [24d2f68](https://github.com/whitequark/parser/commit/24d2f682c56319266602a3b3115e087fd81d667e) and affects only 3.1-dev

Closes https://github.com/whitequark/parser/issues/830


koic:
Thanks! ❤️

PR #834, opened 2021-12-03
warn_syntax_deviation: call `warn` only once

casperisfine:
Calling `warn` 3 times is very annoying for warning management tools that define `Warning.warn`, they're seen as 3 distinct warning instead of a single one.

cc @paracycle

### DIFF
--- a/lib/parser/current.rb
+++ b/lib/parser/current.rb
@@ -3,9 +3,9 @@
 module Parser
   class << self
     def warn_syntax_deviation(feature, version)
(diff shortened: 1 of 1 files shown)


casperisfine:
Thanks @iliabylich !

iliabylich:
Thanks. Please, let me know if you'd like it to be released now. Otherwise it'll be a part of 3.2 post-Christmas release

casperisfine:
Nah, it has been like this for years, it can wait the next update.

Issue #837, opened 2021-12-14
ruby/ruby@aa7c4c3


iliabylich:
https://github.com/ruby/ruby/commit/aa7c4c37d7752af4897567dec1b0745873b0457a

iliabylich:
This commit is a bit problematic to backport. I'm working on it, but it will take some time. Unfortunately, it blocks `3.1.0.0` release.

PR #839, opened 2022-01-02
+ ruby31.y: accept forward arg without parentheses.

iliabylich:
This commit tracks upstream commit ruby/ruby@aa7c4c3.

Also there's a change to Context API that caused a massive rewrite in all grammars.
However, it doesn't affect any logic before 3.1, it's mostly a refactoring.

Closes https://github.com/whitequark/parser/issues/837.

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -714,7 +714,7 @@ module Parser
         node.updated(:gvasgn)
 
       when :const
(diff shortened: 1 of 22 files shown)


iliabylich:
Tested on a corpus of 5476 files with 944280 LOC, there's no diff between this branch and master. Merging.

PR #844, opened 2022-01-27
+ ruby-parse: add ability to read fragment from stdin

jstanley0:
use - as the filename to read from stdin

this is useful when parsing multiple lines or code containing characters that would otherwise need to be escaped

### DIFF
--- a/lib/parser/runner.rb
+++ b/lib/parser/runner.rb
@@ -166,6 +166,7 @@ module Parser
       @option_parser.parse!(options)
 
       # Slop has just removed recognized options from `options`.
+      @fragments << $stdin.read if options.delete('-')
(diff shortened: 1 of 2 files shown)


jstanley0:
another option is to read a fragment from stdin where we're currently printing "need something to parse" and exiting (line 180 in runner.rb); I am happy with either approach

iliabylich:
So this change allows running `echo "foo" | ruby-parse -`, right?

jstanley0:
exactly - or just `ruby-parse -` and then paste some ruby code, followed by Ctrl+D

iliabylich:
Thanks! Let me know if you need a release.

Issue #845, opened 2022-02-15
ruby/ruby@c0651b4


iliabylich:
https://github.com/ruby/ruby/commit/c0651b4ae1426f8ad9322a320d22035223c77537

iliabylich:
Also https://github.com/ruby/ruby/commit/e6b537e605381b6f1c9518adbeb6813c5e579df7

PR #848, opened 2022-04-05
Fix using p_kwnorest in f_no_kwarg

palkan:
Also, enable testing for 3.2 (that's why we haven't caught this yet).

`p_kwnorest` is resolved into `s(:match_nil_pattern)`, which we can not use in `f_no_kwarg`.

Ref https://github.com/whitequark/parser/commit/820cc04b8d176c6f46ccc9a010c55f6113f4b2a

### DIFF
--- a/lib/parser/ruby32.y
+++ b/lib/parser/ruby32.y
@@ -2252,11 +2252,14 @@ opt_block_args_tail:
 
       p_kwnorest: kwrest_mark kNIL
                     {
-                      result = [ @builder.match_nil_pattern(val[0], val[1]) ]
+                      result = val
                     }
 
     p_any_kwrest: p_kwrest
                 | p_kwnorest
+                    {
+                      result = [ @builder.match_nil_pattern(val[0][0], val[0][1]) ]
+                    }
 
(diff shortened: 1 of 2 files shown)


iliabylich:
And then we should re-open https://github.com/whitequark/parser/issues/845, right? I guess correct fix is to change `p_kwnorest` rule to be reduced into `[val[0], val[1]]` and move `s(:match_nil_pattern)` construction to the place that needs it. From what I understand low-level `p_kwnorest` derivation constructs a node that is too high-level.

But I think it's ok to merge this PR and re-open a tracking issue.

iliabylich:
> Also, enable testing for 3.2 (that's why we haven't caught this yet).

my bad 👀

palkan:
> And then we should re-open https://github.com/whitequark/parser/issues/845, right?

Yep.

> I think it's ok to merge this PR and re-open a tracking issue.

Let's wait a bit; I will try to provide a proper fix (as you suggested).

palkan:
@iliabylich Done!

iliabylich:
@palkan Thank you

PR #880, opened 2022-10-26
add tests for multiple pattern matches in a single body.

iliabylich:
This commit tracks upstream commit ruby/ruby@db0e0da.

This is another piece that have been properly implemented in parser since the very beginning. I'm not going to backport the bug.

Closes https://github.com/whitequark/parser/issues/869.


iliabylich:
That was the last 3.2 change as of now. Feel free to ping me if you need a release.

PR #883, opened 2022-10-26
disable failing test on TruffleRuby

iliabylich:
TruffleRuby throws a RegexpError with a different message and so assertion fails:

```ruby
-{:message=>"invalid Unicode escape: /foo-\\u-bar/"}
+{:message=>"invalid Unicode escape"}
```


iliabylich:
OK, CI is fully green 🟢

PR #893, opened 2022-12-13
* lexer.rl: extract `extend_string_escaped` to a separate method

iliabylich:
cc @eregon


eregon (approved):
This looks great, I think we should merge it

Issue #871, opened 2022-10-13
The generated Parser::Lexer#advance method is too big for JIT compilers


eregon:
I have been investigating why `parser` is so slow on TruffleRuby.
I had a suspicion it's a too big generated method and indeed that's now clearly the case from this analysis: https://github.com/oracle/truffleruby/issues/2290#issuecomment-1277796067
This has been found before, and notably I found https://github.com/whitequark/parser/issues/524 from a quick search.

To summarize, the `Parser::Lexer#advance` method is currently 13525 lines long which is humongous.
I would think most if not all Ruby JITs cannot compile such a big method (or if they do they will compile it in a very suboptimal manner).
And so that method needs to be run in the interpreter, and that's typically slower the more advanced the JIT is due to more profiling and other reasons.
Hence parser is slow on anything but CRuby (on MJIT it hangs, on YJIT it's a bit faster like 15% (unsure if it compiles `advance` or not, I'd guess not)). I used `ruby -Ilib -rparser/current -rbenchmark -e 'code=File.read("test/test_lexer.rb"); 10.times { p Benchmark.realtime { Parser::CurrentRuby.parse(code) } }'`.
So on TruffleRuby we can see it tried to compile it, saw it was way too big and bailed out, and then executed the `advance` method all the time in interpreter.

It's easy to reproduce, for example in the parser repo:
```
$ truffleruby --engine.TraceCompilation -Ilib -rparser/current -e 'loop { Parser::CurrentRuby.parse(File.read("test/helper.rb")) }' |& grep 'Lexer#advance'      
[engine] opt failed id=6106  Lexer#advance                                      |Tier 1|Time  778(-24199569+24200347)ms|Reason: org.graalvm.compiler.truffle.compiler.GraphTooBigBailoutException: Graph too big to safely compile. Node count: 113880. Graph Size: 150001. Limit: 150000.|Timestamp 24200346847485|Src lexer.rb:11303
[engine] opt failed id=6106  Lexer#advance                                      |Tier 1|Time  635(-24200352+24200987)ms|Reason: org.graalvm.compiler.truffle.compiler.GraphTooBigBailoutException: Graph too big to safely compile. Node count: 113879. Graph Size: 150001. Limit: 150000.|Timestamp 24200987352654|Src lexer.rb:11303
```
(The `Graph Size: 150001` just means it aborted after reaching the limit)

Interesting it does compile when lexing very small and simple files
```
[...]

PR #894, opened 2022-12-15
+ lexer.rl: use Ragel -F0 on non-CRuby

eregon:
* F0 generates a much smaller Parser::Lexer#advance method which is much better for JITs.
* See https://github.com/whitequark/parser/issues/871

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,8 @@ tmp
 *.output
 .ruby-version
 .ruby-gemset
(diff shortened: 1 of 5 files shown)


iliabylich:
Thanks!

headius:
This change is significantly slower on JRuby and only marginally faster on TruffleRuby without experimental options: #897

PR #896, opened 2022-12-17
Reduce the number of allocated Ranges

fatkodima:
While optimizing rubocop, I identified that `parser` creates ~100Mb of unneeded `Range` objects while checking just 660 ruby files.

Additionally, made 2 more tiny changes which were producing 1-2% of execution and made them a little faster. 

See https://github.com/rubocop/rubocop/pull/11300 for more details.


iliabylich:
Thanks!

PR #902, opened 2022-12-21
Move blocks outside Lexer#advance

eregon:
* This gives big compilation size and time benefit, because the local variables of Lexer#advance are no longer captured by those blocks (even if unused due to Ruby semantics and `Proc#binding`).
* See https://github.com/whitequark/parser/issues/871#issuecomment-1362218586

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -116,6 +116,19 @@ class Parser::Lexer
         []
       end
 
+    @emit_integer = lambda { |chars, p| emit(:tINTEGER,   chars); p }
+    @emit_rational = lambda { |chars, p| emit(:tRATIONAL,  Rational(chars)); p }
+    @emit_imaginary = lambda { |chars, p| emit(:tIMAGINARY, Complex(0, chars)); p }
+    @emit_imaginary_rational = lambda { |chars, p| emit(:tIMAGINARY, Complex(0, Rational(chars))); p }
+    @emit_integer_re = lambda { |chars, p| emit(:tINTEGER,   chars, @ts, @te - 2); p - 2 }
+    @emit_integer_if = lambda { |chars, p| emit(:tINTEGER,   chars, @ts, @te - 2); p - 2 }
+    @emit_integer_rescue = lambda { |chars, p| emit(:tINTEGER,   chars, @ts, @te - 6); p - 6 }
+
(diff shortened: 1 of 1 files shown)


eregon:
It would be good to add a test to ensure `Lexer#advance` remains without block/proc/lambda/lambda-literal inside it, as it matters quite a bit for performance and compilation size/time.
I'm not able to do that now though (it's already past midnight here and I'm traveling tomorrow).

lib/parser/lexer.rl
```diff
@@ -833,6 +834,20 @@ class Parser::Lexer
     KEYWORDS_BEGIN[keyword] = KEYWORDS[keyword] = :"k#{keyword.upcase}"
   end
 
+  EMIT_INTEGER = lambda { |lexer, chars| lexer.emit(:tINTEGER,   chars) }
+  EMIT_RATIONAL = lambda { |lexer, chars| lexer.emit(:tRATIONAL,  Rational(chars)) }
+  EMIT_IMAGINARY = lambda { |lexer, chars| lexer.emit(:tIMAGINARY, Complex(0, chars)) }
+  EMIT_IMAGINARY_RATIONAL = lambda { |lexer, chars| lexer.emit(:tIMAGINARY, Complex(0, Rational(chars))) }
+  EMIT_INTEGER_RE = lambda { |lexer, chars| lexer.emit(:tINTEGER,   chars, @ts, @te - 2); p -= 2 }
```
iliabylich, line 841: `p` is a local variable from `Lexer#advance`

eregon: Oh, right.

eregon:
The failing test shows these lambda also access some ivars, so they can't be constants, I'll move them in `initialize` then.

lib/parser/lexer.rl
eregon: Fixed.

eregon:
Done, CI should be green now.

iliabylich:
Thanks!

eregon:
I missed that https://github.com/whitequark/parser/pull/901/commits/5b2135eda12035ab700f52efc995f11de37fd8ff#diff-894f90ef5e566bf82a7b031fd82df1aece88eaac96cae6483feba8c020a63761R884 also did this partially in a slightly different way, which also works but has the slight perf disadvantage of allocating a Proc per such method call vs one per Lexer instance.
So, my apologies about the potential conflicts (should be just a few lines though).

PR #904, opened 2023-01-05
+ Ruby 3.3 support

olefriis:
I've basically mimicked #841.

* Copied `lib/parser/ruby32.y` to `lib/parser/ruby33.y`, editing the version number on the first line. Since Ruby 3.2 was released a couple of weeks ago, the grammars are probably identical for now (?).
* Other changes similar to #841.
* Added a changelog entry. As I understand the version number specs for this repo, adding Ruby 3.3 support means bumping a "minor" version. I have _not_ bumped the version in `lib/parser/version.rb`, as I suppose this will be part of the actual release process. Also, since I do not know when the version will be released, I haven't added a timestamp in the changelog.


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+v3.3.0.0
```
iliabylich, line 4: Could you revert this part please?

lib/parser/current.rb
```diff
@@ -110,6 +110,15 @@ def warn_syntax_deviation(feature, version)
     require 'parser/ruby32'
     CurrentRuby = Ruby32
 
+  when /^3\.3\./
+    current_version = '3.3.0'
```
iliabylich, line 114: Should be `3.3.0-dev`, we'll change it to `3.3.0` once it's released (in a year)

lib/parser/ruby33.y
```diff
@@ -0,0 +1,3238 @@
...
+end
+
+---- header
+
+require 'parser'
+
+---- inner
+
+  def version
+    32
```
iliabylich, line 3171: This should be 33

iliabylich (changes requested):
Thanks! I've left a few comments.

Also there are two missing pieces:

1.  `parse_helper.rb`, `ALL_VERSIONS` constants should include `3.3`
2. `parse_helper.rb`, `parser_for_ruby_version` method needs to handle `3.3`

CHANGELOG.md
olefriis: Definitely! Wasn't sure if this had to be part of the PR anyway. 👍

lib/parser/current.rb
olefriis: Cool! Fixed.

lib/parser/ruby33.y
olefriis: Ouch, I didn't notice that one. Thanks! And fixed.

olefriis:
Thanks for the extremely quick feedback, @iliabylich. I've applied the suggested changes, except the one about the `parser_for_ruby_version` needing to handle `3.3`. That should already be there, right?

iliabylich:
Thanks!

nickborromeo:
👋 @iliabylich can I ask when this change is going to be scheduled to be released as a patch version for the gem? 🙇

iliabylich:
I don't plan to release until Ruby 3.2.1 (i.e. with release of the next Ruby version) or in case of a bugfix (that would be `parser` 3.2.0.1).

Why do you need this change?

nickborromeo:
We have a running branch every week to test Ruby and the sha that we are targeting is [9726736006b3f74635fd8af05814fe0908e2cf84](https://github.com/github/ruby) which contains this [change for the versions](https://github.com/ruby/ruby/blob/cc15963aa30ed41d4e5263233134d275de832683/include/ruby/version.h#L59-L76)

It looks like there are other gems like `rubocop-ast` and `rubocop` that rely on this, right now we are just trying to get a timeline of things (which you have already provided since you are planning to release when Ruby 3.2.1 comes out.

PR #910, opened 2023-03-03
Fixed default builder invalid subexp warning

bkuhlmann:
## Overview

Hello. :wave: I was getting Ruby warnings with the latest version of this gem when running RuboCop so this will help fix the issue.

## Screenshots/Screencasts

Was seeing the following when running RuboCop:

2023-03-03_11-35-40-iTerm2

## Details

Necessary to resolve the following Ruby warnings from showing up in the console:

    parser-3.2.1.0/lib/parser/builders/default.rb:2249: warning: invalid subexp call

If the `source` is properly converted to a string then these warnings go away and keeps your console and logs clean.

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -2246,7 +2246,7 @@ module Parser
         source
       end
 
-      Regexp.new(source, (Regexp::EXTENDED if options.children.include?(:x)))
+      Regexp.new(source.to_s, (Regexp::EXTENDED if options.children.include?(:x)))
     end
 
     def static_regexp_node(node)


iliabylich:
Thinks. Just our of curiosity, do you have any code example that triggers this warning? This is the first time I see it.

lib/parser/builders/default.rb
```diff
@@ -2246,7 +2246,7 @@ def static_regexp(parts, options)
         source
       end
 
-      Regexp.new(source, (Regexp::EXTENDED if options.children.include?(:x)))
+      Regexp.new(source.to_s, (Regexp::EXTENDED if options.children.include?(:x)))
```
iliabylich, line 2249: Please correct if I'm wrong, but I think `source` is always a `String` on this line, and so calling `.to_s` should be a noop.

bkuhlmann: You're correct -- at least when I put a breakpoint before Line 2249 -- I can see it's a string but for some reason it doesn't cast unless you message `#to_s`.

iliabylich:
Hmmm, first of all, calling `to_s` doesn't fix the warning for me.

Also, this code triggers the warning:

```ruby
3.0.0 :022 > Regexp.new('\gi')
/Users/ilyabylich/Work/parser-fork/lib/parser/builders/default.rb:22: warning: invalid subexp call: /.+\gi/
 => /\gi/
3.0.0 :023 > Regexp.new('\g')
 => /\g/
```

According to [Regexp documentation](https://ruby-doc.org/3.2.1/Regexp.html#top) `\g` syntax is used as `\g<name>`, so the warning makes sense.

This warning comes just from constructing a Regexp from its source, and running `ruby -c demo.rb` also prints the same warning. A proper fix would be setting `$VERBOSE` to `nil`, constructing a Regexp and restoring `$VERBOSE`.

However `parser` doesn't silence any warnings from Ruby, so this shouldn't even be fixed. There are many other Regexp-related warnings that we let Ruby print during parsing. Closing as there's nothing to fix.

bkuhlmann:
Thanks and sorry about this. I didn't realize my regex was invalid to begin with but the error message was confusing and tripping me up. I've fixed my regular expression and the issue is resolved.

PR #948, opened 2023-10-12
Replace whitequark blog links with archive URLs

backus:
Unfortunately, whitequark's amazing blog has been down since early 2022 as far as I can tell. These blog posts are great resources still for understanding parts of this gem, so I've replaced the broken blog links in the README with the most recent working Internet Archive links.


iliabylich:
Thanks!

PR #954, opened 2023-11-10
* Bump Racc to 1.7.3

koic:
Racc 1.7.3 has been released and this PR bumps Racc to 1.7.3.
https://github.com/ruby/racc/releases/tag/v1.7.3

`racc` is specified in the gemspec, it might be possible to remove the `racc` dependency from the Gemfile, although a specific version is still being specified, maintaining previous practices.


iliabylich:
Thanks.

Issue #951, opened 2023-10-23
ruby/ruby@3b05238


iliabylich:
https://github.com/ruby/ruby/commit/3b05238289b973af3d5ff73d750ba243c0c2fa3f

PR #976, opened 2023-12-28
+ ruby33.y: extract p_assoc and p_in rules

iliabylich:
This commit tracks upstream commit ruby/ruby@3b05238.

Closes https://github.com/whitequark/parser/issues/951.

Racc doesn't have access to a result of a `p_assoc` rule in the last mid-rule of the diff (probably because of a lookahead?) and so it can't read `$ctxt.in_kwarg`. If that's true for Lrama then it's a very nasty bug that I don't see exactly how to reproduce :(. It could be something like this:

```ruby
def foo a: case 1; in 1; then 1; end, b:
  p 42
end
```

that gets rejected (this particular snippet is accepted by MRI which is correct). It could even be a UB if generated code reads uninitialised memory and transmutes it to `ctxt` (then it becomes a heisenbug) but I hope Ruby CI runs ASAN for all of its tests.

### DIFF
--- a/lib/parser/ruby33.y
+++ b/lib/parser/ruby33.y
@@ -332,39 +332,35 @@ rule
                     {
                       result = @builder.not_op(val[0], nil, val[1], nil)
                     }
-                | arg tASSOC
+                | arg p_assoc
                     {
                       @lexer.state = :expr_beg
                       @lexer.command_start = false
                       @pattern_variables.push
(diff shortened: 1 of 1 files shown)


iliabylich:
cc @nobu, I hope it's a false alarm and Lrama does everything right but could you PTAL? Do all mid-rules have access to results of previous mid-rules in the scope of the same derivation?

PR #985, opened 2024-01-06
Fix incorrect required_ruby_version

Watson1978:
This gem has used String#delete_suffix which was introduced at Ruby 2.5. https://github.com/whitequark/parser/blob/95d26953ebe823211845fb84cb93c18cd501f561/lib/parser/lexer/literal.rb#L250

Ref. https://github.com/ruby/ruby/blob/ruby_2_5/NEWS

So, even Ruby 2.0 can be installed now, however it will not work.


iliabylich:
Thanks!

PR #986, opened 2024-01-06
- Supports Ruby 2.0+ runtime

koic:
This is an alternative approach to the changes in #985. The Parser gem may be required to operate across as many Ruby versions as possible. The use of `delete_suffix` starts from Parser version 3.3.0.0. Therefore, by replacing `delete_suffix` with `sub`, and releasing Parser 3.3.0.1, supporting Ruby 2.0 would be feasible.

### DIFF
--- a/lib/parser/lexer/literal.rb
+++ b/lib/parser/lexer/literal.rb
@@ -247,7 +247,7 @@ module Parser
         # E
         # because there are not enough leading spaces in the closing delimiter.
         delimiter.end_with?(@end_delim) &&
-          delimiter.delete_suffix(@end_delim).bytes.all? { |c| c == SPACE }
+          delimiter.sub(/#{@end_delim}\z/, '').bytes.all? { |c| c == SPACE }
       elsif @indent
         @end_delim == delimiter.lstrip
       else
--- a/parser.gemspec
+++ b/parser.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
-  spec.required_ruby_version = '>= 2.5.0'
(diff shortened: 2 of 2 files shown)


iliabylich:
Ruby 2.0.0 has reached EOL at `2016-02-24`, what's the reason to keep supporting it? Not saying that it's bad or wrong, but it doesn't let us use any new language features.

koic:
For example, Ruby Next is used in a transpiler that allows for using new Ruby syntax on older Ruby runtimes. It seems to support Ruby 2.2+.
https://github.com/ruby-next/ruby-next/blob/v1.0.0/ruby-next.gemspec#L28

There are also users who cannot necessarily move away from EOL versions. Considering that the Parser gem is arguably the most utilized Ruby parser implementation, I think there is worth in supporting older Ruby runtimes as much as possible.

On the other hand, I deeply regret that this comes at the cost of not being able to use newer language features 🙇

iliabylich:
Ok, fair enough

iliabylich:
I've released `3.3.0.1`, just in case anyone needs it

koic:
Thank you for considering this matter.

marcandre:
It's not clear to me why someone with a very old Ruby would want to use later versions of `parser` (earlier versions still work and were fine parsing old Ruby code)

Nevertheless, if this comes up from time to time, consider using `backports` to bring in the methods needed like `delete_suffix` (or `ruby-next/core`, but I don't think there's an API for using just a few particular methods)

mbj:
> It's not clear to me why someone with a very old Ruby would want to use later versions of `parser` (earlier versions still work and were fine parsing old Ruby code)
> 
> Nevertheless, if this comes up from time to time, consider using `backports` to bring in the methods needed like `delete_suffix` (or `ruby-next/core`, but I don't think there's an API for using just a few particular methods)

these users can also do core extensions to do a "single method" backport if needed. I personally would not regress my own project for these users.

Issue #989, opened 2024-01-12
Syntax error on heredoc inside block indented with tabs


bquorning:
[An error report in rubocop-rspec](https://github.com/rubocop/rubocop-rspec/issues/1774) led me all the way via rubocop-ast into this gem. I wrote the following test case for test/test_parser.rb, and it fails (seemingly in _all_ ruby versions and not just 3.3.0 as in the linked issue) with `Parser::SyntaxError: (2.3) unterminated string meets end of file`

```rb
  def test_parser_of_heredoc_inside_block_with_tab_indentation
    assert_parses(
      s(:block,
        s(:send, nil, :foo),
        s(:args),
        s(:str, "bar\n")),
      %Q{foo do!\t<<~DESC!\t\tbar!\tDESC!end}.gsub('!', "\n"),
      %q{},
      SINCE_2_3)
  end
```

AlexWayfer:
With Ruby 3.2 RuboCop gives such output (not breaking itself at least):

```
> bundle exec rubocop
Inspecting 8 files
...F...F

Offenses:

lib/static_files_toys/template.rb:10:18: F: Lint/Syntax: unterminated string meets end of file
(Using Ruby 3.0 parser; configure using TargetRubyVersion parameter, under AllCops)
		GREP_OPTIONS = <<~STR.chomp
                 ^
static_files_toys.gemspec:12:23: F: Lint/Syntax: unterminated string meets end of file
(Using Ruby 3.0 parser; configure using TargetRubyVersion parameter, under AllCops)
	spec.description   = <<~DESC
                      ^

8 files inspected, 2 offenses detected
```

iliabylich:
Thanks for reporting, that's definitely a regression introduced in `3.3.0.0`, I'll make a fix soon.

AlexWayfer:
I can confirm that version `3.3.0.3` fixes the problem. Great thank you.

iliabylich:
parser `3.3.0.3` has been released (and it seems to fix the issue, thanks @AlexWayfer )

bquorning:
Thank you for the quick fix ❤️

PR #990, opened 2024-01-12
- lexer.rl: accept tabs before closing heredoc delimiter

iliabylich:
Closes https://github.com/whitequark/parser/issues/989.

The bug is not related to blocks specifically but example from the report can also be parsed now:

```
$ /bin/cat test.rb
foo do
	<<~DESC
		bar
	DESC
end

$ bin/ruby-parse --33 test.rb
(block
  (send nil :foo)
  (args)
  (str "bar\n"))
```


iliabylich:
Turns out the latest minitest [also defines `BASE_DIR`](https://github.com/minitest/minitest/commit/e410bade1723478cffb234faedd725cb0a599ec0) that conflicts with our global variable 😞

PR #991, opened 2024-01-15
Register a Ruby 3.4 parser

casperisfine:
To avoid failing when running as part of ruby-head test builds.

I did the same as previous years, just initialized 3.4 as a copy of 3.3.


koic:
Rakefile lacks the following change:

```diff
diff --git a/Rakefile b/Rakefile
index 778e61db..1afcf209 100644
--- a/Rakefile
+++ b/Rakefile
@@ -38,6 +38,7 @@ GENERATED_FILES = %w(lib/parser/lexer-F0.rb
                      lib/parser/ruby31.rb
                      lib/parser/ruby32.rb
                      lib/parser/ruby33.rb
+                     lib/parser/ruby34.rb
                      lib/parser/macruby.rb
                      lib/parser/rubymotion.rb)
```

Ref: https://github.com/whitequark/parser/pull/904

casperisfine:
@koic good catch. I updated my PR.

koic (approved):

casperisfine:
> Honestly I was hoping for some hype around Prism to slowly deprecate `parser` for 3.4+ Rubies.

I highly suspect 3.4 will be the last one, but yeah, right now lots of tooling hasn't moved (and will probably take a while because `prism` doesn't support older rubies (yet?)

> Do you need a release?

It would be nice, but no rush.

casperisfine:
Much appreciated.

PR #999, opened 2024-03-10
Raise a more specific error when encountering an unknown magic comment encoding

### LINKED ISSUE #998: Parsing code with an invalid encoding raises
```ruby
Parser::CurrentRuby.parse("# encoding: utf")

# /home/earlopain/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/parser-3.3.0.5/lib/parser/source/buffer.rb:69:in `find': unknown encoding name - utf (ArgumentError)
#
#          Encoding.find(result[3] || result[4] || result[6])
```

Ruby does raise an error when trying to execute the file: 
```
[earlopain@DESKTOP-PC whitequark-parser]$ ruby test.rb
test.rb:1: unknown encoding name: utf (ArgumentError)
```

It doesn't look like a `SyntaxError`, not quite sure what this actually is, where it's coming from. However I believe the parser gem here shouldn't raise an `ArgumentError` in this case. This makes tools like RuboCop raise as well when trying to analyze these invalid files.

Earlopain:
Closes #998

Let me know if I should tweak the docs for the exception in some way.

I also know that this library still targets Ruby 2.0. CI only seems to test against somewhat recent versions. I used the old begin/end syntax since I know that omitting the begin was added somewhere between but other than that I haven't tested on older versions.

### DIFF
--- a/lib/parser.rb
+++ b/lib/parser.rb
@@ -59,6 +59,7 @@ module Parser
 
   require 'parser/syntax_error'
   require 'parser/clobbering_error'
+  require 'parser/unknown_encoding_in_magic_comment_error'
   require 'parser/diagnostic'
   require 'parser/diagnostic/engine'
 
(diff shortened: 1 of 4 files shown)


iliabylich:
> Let me know if I should tweak the docs for the exception in some way.

Looks good to me. Thank you very much!

PR #1003, opened 2024-03-17
* Use `require_relative` in the Parser codebase

koic:
If there are many searches in the `$LOAD_PATH` in the user environment, require will perform unnecessary searches that are not needed. In contrast, `require_relative` is efficient because it uses a relative path.

Parser gem requires Ruby 2.0+, it is possible to use `require_relative`, which was introduced in Ruby 1.9.
https://github.com/whitequark/parser/blob/v3.3.0.5/parser.gemspec#L27


iliabylich:
Thanks!

PR #1013, opened 2024-05-17
Supress warnings during parsing

Earlopain:
User input can contain arbitrary regexes.
Something like `Parser::CurrentRuby.parse("/[\177\01\1778]/")` shows a warning:
> character class has duplicated range: /[\x7f\x01\x7f8]/

I'm interested in getting the warning count in the RuboCop test suite down and since it has a few regex cops there are quite a few warnings of that kind as well. They are unactionable, RuboCop instead emulates these warnings by inspecting the ast parsing the regexp with the `regexp_parser` gem.

-------

Also noticed the same for floats:
`Parser::CurrentRuby.parse("9.9999e999")`
> Float 9.9999e999 out of range

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -2263,7 +2263,12 @@ module Parser
         source
       end
 
-      Regexp.new(source, (Regexp::EXTENDED if options.children.include?(:x)))
+      begin
+        old_verbose, $VERBOSE = $VERBOSE, nil
+        Regexp.new(source, (Regexp::EXTENDED if options.children.include?(:x)))
+      ensure
+        $VERBOSE = old_verbose
+      end
     end
 
     def static_regexp_node(node)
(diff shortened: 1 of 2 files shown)


iliabylich:
@koic Does it look good for you? I'm not sure if these warnings are a part of some implicit contract that other libraries might rely on. If not I believe we should merge it.

Earlopain:
I've ignored these warnings in RuboCop itself so this is not strictly necessary. If you feel there is some risk associated here then feel free to close (though it would be nice not having to ignore these warnings)

dvandersluis:
I ran into this with https://github.com/rubocop/rubocop/pull/13385 too. I'm trying to add a RuboCop for `warning: regular expression has ']' without escape`, but parser outputs that same warning while parsing.

koic:
It appears that these warnings come not from diagnostic processing by the Parser gem, from Ruby implementation code. This change seems reasonable.

PR #1022, opened 2024-06-08
- lexer.rl: handle CLRF as a line separator

### LINKED ISSUE #1020: Offsets with \r\n in source
When you parse source that contains \r\n in the source, they are automatically converted into \n, as per https://github.com/whitequark/parser/blob/8bd5ec3416d58c28ab230b6e34a3bba1b58a8f06/lib/parser/source/buffer.rb#L190. The issue is that this can really throw off source locations. For example:

```ruby
Parser::CurrentRuby.parse("1\r\n2\r\n3").children[2].loc
# => #<Parser::Source::Map::Operator:0x000000010b47c3d0 @expression=#<Parser::Source::Range (string) 4...5>, @node=s(:int, 3), @operator=nil>
```

This is saying the source range is `4...5`, which is one of the `\r` characters.

For prism's purposes it's okay if the locations are different, I'll just make it so that it doesn't compare locations for files that contain `\r\n`. My issue is that I use the source buffer to parse with both parsers (https://github.com/ruby/prism/blob/90d570aa50bfff43c66e5f6c600370a61c091329/test/prism/ruby/parser_test.rb#L188-L208) but the source has already been modified internally in the buffer with no way to retrieve the original.

For a solution, I'm wondering if either:

(a) The `\r\n` gsub can be removed (and therefore have the parser instead of the buffer replace \r\n when necessary)
(b) The `Buffer` class could support an `auto_clrf` parameter (or something) that would disable that behavior

iliabylich:
Closes https://github.com/whitequark/parser/issues/1020.

```ruby
3.3.1 :004 > Parser::CurrentRuby.parse("1\r\n2\r\n3").children[2].loc
 => #<Parser::Source::Map::Operator:0x00000001222f0f80 @expression=#<Parser::Source::Range (string) 6...7>, @node=s(:int, 3), @operator=nil>

3.3.1 :005 > Parser::CurrentRuby.parse("1\r\n2\r\n3").children[2].loc.expression.source
 => "3"
```

A few notes:

1. If `\r\n` is a line separator `parser` still emits `tNL` token with location of the `\n` character
2. `tSTRING_CONTENT` tokens now have proper locations, but the content doesn't include `\r` part of `\r\n` (because `eval(%{"\r\n"})` is just `"\n"`), so `.source` of their locations doesn't match string content. I guess it's fine, the same happens with all escape sequences anyway.

If it doesn't break Rubocop's test suite I guess it's safe to merge it as is.

@kddnewton Could you take a look at this please? Does it fix Prism's translator?

### DIFF
--- a/lib/parser/lexer.rl
+++ b/lib/parser/lexer.rl
@@ -545,7 +545,7 @@ class Parser::Lexer
     @newline_s = p
   }
 
-  c_nl       = '\n' $ do_nl;
+  c_nl       = ('\r\n' | '\n') $ do_nl;
   c_space    = [ \t\r\f\v];
   c_space_nl = c_space | c_nl;
 
@@ -1995,7 +1995,7 @@ class Parser::Lexer
 
       # Here we use '\n' instead of w_newline to not modify @newline_s
       # and eventually properly emit tNL
-      (c_space* w_space_comment '\n')+
+      (c_space* w_space_comment ('\r\n' | '\n'))+
       => {
         if @version < 27
           # Ruby before 2.7 doesn't support comments before leading dot.
--- a/lib/parser/lexer/literal.rb
+++ b/lib/parser/lexer/literal.rb
@@ -208,7 +208,7 @@ module Parser
       end
 
       unless @buffer.empty?
-        emit(:tSTRING_CONTENT, @buffer, @buffer_s, @buffer_e)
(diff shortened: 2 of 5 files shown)


kddnewton:
This gets close, but runs into issues with escaped `\r` and literal `\n` then getting grouped, as in:

```ruby
<<EOS
foo\rbar
baz\r
EOS
```

(There are regular newline characters after each line.) In this PR, it groups that last `\r\n` because the gsub is happening after escape sequences are resolved.

PR #1028, opened 2024-07-19
Suppress Ruby 3.4's warning

koic:
Starting with Ruby 3.4, there is a gradual plan to freeze strings: https://bugs.ruby-lang.org/issues/20205#note-35

This PR suppresses the following Ruby 3.4's warning:

```console
/Users/koic/.rbenv/versions/3.4-dev/lib/ruby/gems/3.4.0+0/gems/parser-3.3.4.0/lib/parser/source/buffer.rb:97:
warning: literal string will be frozen in the future
```


iliabylich:
Thanks!

PR #1030, opened 2024-07-24
Create SECURITY.md

Desertsmogtech:


iliabylich:
Sorry but I'm not sure if I get the purpose of this addition. What kind of security issues can be in a library that doesn't do anything except for in-memory byte manipulation using a safe language?

PR #1038, opened 2024-08-29
Remove `benchmark` dependency from the runner

Earlopain:
Ruby wants to bundle it: https://github.com/ruby/ruby/pull/11492, which means that the next ruby release will warn and the one after that will error, unless it is added to the gemspec. Usage is trivial and can simply be inlined.

The benchmark gem does it like this for 14 years already so I don't imagine compatibility problems with older ruby versions: https://github.com/ruby/benchmark/blame/0383673a9089d6a74a73ef37c67d5afc15d66d71/lib/benchmark.rb#L277-L287

Test with `bin/ruby-parse lib/parser/ruby34.rb --bench`.

### DIFF
--- a/lib/parser/runner.rb
+++ b/lib/parser/runner.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'benchmark'
 require 'find'
 require 'optparse'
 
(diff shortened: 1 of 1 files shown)


iliabylich:
Thanks. Do you know whether `Process.times.utime` uses monotonic time? If not could you replace it with `Process.clock_gettime(Process::CLOCK_MONOTONIC)` please?

Earlopain:
I believe it is but `CLOCK_MONOTONIC` seems more correct anyways. You basically want the cumulative of these https://ruby-doc.org/3.3.4/Process.html#method-c-times. I don't think it matters much in this context butI switched it over :+1:

iliabylich:
Thank you very much!

PR #1044, opened 2024-10-31
* Bump maintenance branches to 3.2.6

### LINKED ISSUE #1043: New 3.2.6 Release
I noticed that [3.2.6](https://www.ruby-lang.org/en/news/2024/10/30/ruby-3-2-6-released/) is out. Can we get a release to suppress the typical incompatible version warning that I expect would arise?

As a note, would it be possible to make the warning less sensitive to patch releases?

koic:
Closes #1043.

Ruby 3.2.6 has been released:
https://www.ruby-lang.org/en/news/2024/10/30/ruby-3-2-6-released/

Bump 3.2 branch from 3.2.5 to 3.2.6:
https://github.com/ruby/ruby/compare/v3_2_5...v3_2_6

There seems to be no change to the syntax.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,14 +17,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["3.0.7", "3.1.6", "3.2.5", "3.3.5", "jruby-9.2"]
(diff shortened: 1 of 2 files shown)


iliabylich:
Thanks, `3.3.5.1` has been released

PR #1047, opened 2024-12-17
- Fix `ruby-parse` with a folder ending in `.rb`

Earlopain:
It's just checking that the path ends with `.rb`, which can happen with folders.

For example: https://github.com/chef/chef/tree/1ea245124173f803af628abd70adc87baa70a873/knife/spec/data/client.d_02/foo.rb


iliabylich:
Thanks

PR #1048, opened 2024-12-17
+ ruby34.y: reject `return` in singleton class

### LINKED ISSUE #1029: ruby/ruby@e642ddf
https://github.com/ruby/ruby/commit/e642ddf7ae86e306674559ae745823fdbf56ea10

Earlopain:
This tracks upstream commit https://github.com/ruby/ruby/commit/e642ddf7ae86e306674559ae745823fdbf56ea10

Closes #1029.

Note that `in_class` is now unused for 3.4 as far as I can tell. Ruby uses it in `begin_definition` helper but I don't think there's something equivalent with ragel here, or I just don't know about it. Kept around to keep it closer to upstream.

Tested against https://github.com/eliotsykes/real-world-rails and https://github.com/jeromedalbert/real-world-ruby-apps, got no unexpected syntax errors.

### DIFF
--- a/lib/parser/context.rb
+++ b/lib/parser/context.rb
@@ -24,6 +24,7 @@ module Parser
       in_class
       in_block
       in_lambda
+      cant_return
     ]
 
     def initialize
@@ -38,6 +39,7 @@ module Parser
       @in_class = false
       @in_block = false
       @in_lambda = false
+      @cant_return = false
     end
 
     attr_accessor(*FLAGS)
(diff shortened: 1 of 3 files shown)


iliabylich:
Thanks. Can't say I like that there's a version-specific shape of the `context` now but I don't see a better solution (branching `context` would most probably only make it worse), so 👍

Earlopain:
Could try to backport this to the older parsers. But tbh that seems pretty daunting, especially when they start to deviate more and more. I think I chose a pretty easy issue to start out with otherwise, so I'm glad you didn't insist on it :sweat_smile:

PR #1049, opened 2024-12-18
- builder.rb: correctly handle `...` forwarding to super with explicit block

### LINKED ISSUE #1004: ruby/ruby@a850cd1
https://github.com/ruby/ruby/commit/a850cd1a87bef738c40d9c550fb8823699083f2e

Earlopain:
Closes #1004

Upstream commit https://github.com/ruby/ruby/commit/a850cd1a87bef738c40d9c550fb8823699083f2e is specific to 3.3, previous versions already did not allow this.
See https://bugs.ruby-lang.org/issues/20392 for a bit more info.
    
But since the super node has has a different shape, it behaved unexpectedly in earlier ruby versions already.
    
It allowed forwarding with zero and one arguments and failed for more. Now it correctly rejects as invalid syntax, except in Ruby 3.3, that fix has not been backported.

### DIFF
--- a/lib/parser/builders/default.rb
+++ b/lib/parser/builders/default.rb
@@ -1120,15 +1120,19 @@ module Parser
     end
 
     def block(method_call, begin_t, args, body, end_t)
-      _receiver, _selector, *call_args = *method_call
-
       if method_call.type == :yield
         diagnostic :error, :block_given_to_yield, nil, method_call.loc.keyword, [loc(begin_t)]
       end
 
-      last_arg = call_args.last
+      if method_call.type == :super
+        *_args, last_arg = *method_call
+      else
+        _receiver, _selector, *_args, last_arg = *method_call
+      end
       if last_arg && (last_arg.type == :block_pass || last_arg.type == :forwarded_args)
-        diagnostic :error, :block_and_blockarg, nil, last_arg.loc.expression, [loc(begin_t)]
+        if (@parser.version == 33 && method_call.type != :super) || @parser.version != 33
+          diagnostic :error, :block_and_blockarg, nil, last_arg.loc.expression, [loc(begin_t)]
+        end
       end
 
       if args.type == :numargs
--- a/test/test_parser.rb
+++ b/test/test_parser.rb
@@ -8163,6 +8163,30 @@ class TestParser < Minitest::Test
       SINCE_3_1 - SINCE_2_7)
   end
 
+  def test_forward_args_super_with_block
+    [
+      %q{def foo(...) super(...) {}; end},
+      %q{def foo(...) super(a, ...) {}; end},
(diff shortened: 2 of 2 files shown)


lib/parser/builders/default.rb
```diff
@@ -1120,15 +1120,19 @@ def call_lambda(lambda_t)
...
 
-      last_arg = call_args.last
+      if method_call.type == :super
+        *_args, last_arg = *method_call
+      else
+        _receiver, _selector, *_args, last_arg = *method_call
+      end
       if last_arg && (last_arg.type == :block_pass || last_arg.type == :forwarded_args)
-        diagnostic :error, :block_and_blockarg, nil, last_arg.loc.expression, [loc(begin_t)]
+        if (@parser.version == 33 && method_call.type != :super) || @parser.version != 33
```
iliabylich, line 1133: Should it be `>= 33` here and `< 33` on the right side?

Earlopain: `super(...) {}` is a syntax error already starting on 2.7, and it's only a bug in 3.3 that it is accepted (still present in latest patch version).

```sh
# super
$ ruby -cve "def foo(...) super(...) {}; end"
ruby 2.7.8p225 (2023-03-30 revision 1f4d455848) [x86_64-linux]
-e:1: both block arg and actual block given

$ ruby -cve "def foo(...) super(...) {}; end"
ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [x86_64-linux]
Syntax OK

$ ruby -cve "def foo(...) super(...) {}; end"
ruby 3.4.0dev (2024-12-11T11:30:36Z master 2335ecb7fa) +PRISM [x86_64-linux]
ruby: -e:1: syntax error found (SyntaxError)
> 1 | def foo(...) super(...) {}; end
    |                         ^~ both block arg and actual block given; only one block is allowed
  2 | 

# plain method
$ ruby -cve "def foo(...) bar(...) {}; end"
ruby 2.7.8p225 (2023-03-30 revision 1f4d455848) [x86_64-linux]
-e:1: both block arg and actual block given

$ ruby -cve "def foo(...) bar(...) {}; end"
ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [x86_64-linux]
-e:1: both block arg and actual block given
ruby: compile error (SyntaxError)

$ ruby -cve "def foo(...) bar(...) {}; end"
ruby 3.4.0dev (2024-12-11T11:30:36Z master 2335ecb7fa) +PRISM [x86_64-linux]
ruby: -e:1: syntax error found (SyntaxError)
> 1 | def foo(...) bar(...) {}; end
    |                       ^~ both block arg and actual block given; only one block is allowed
  2 | 
```

Under these circumstances, I think this is correct? Basically it should exclude checking `super` only in one version and otherwise check it always, including normal method calls.

PR #1060, opened 2025-01-11
Numbered parameters are valid for pattern matching pinning

Earlopain:
The following is how different versions behave:
```rb
# Valid only in 2.7
case 0; in _1; end
1.then { 1 in _1 }

# Valid in no versions
case 0; in ^_1; end

# Valid in all versions
1.then { 1 in ^_1 }
```

`@builder.accessible` makes the numparam visible to the environment, so it must be called first

### DIFF
--- a/lib/parser/ruby27.y
+++ b/lib/parser/ruby27.y
@@ -2152,11 +2152,12 @@ opt_block_args_tail:
        p_var_ref: tCARET tIDENTIFIER
                     {
                       name = val[1][0]
+                      lvar = @builder.accessible(@builder.ident(val[1]))
+
                       unless static_env.declared?(name)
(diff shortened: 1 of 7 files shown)


Earlopain:
@koic would you merge this fix and the others I currently have open? If not, I'm just going to close these PRs.

koic:
OK, I'll likely review it sometime next week and cut a new release. Please wait a moment.

Earlopain:
Thanks, that would be nice.

I'm actually running into this again with the prism tests, where a fixture contains such code but the whole file is getting skipped because `parser` doesn't parse it.

koic:
Parser 3.3.7.4 has been released with the previously open PRs. Thank you!

Earlopain:
Thank you as well!

Issue #1046, opened 2024-11-21
Soft deprecation of the project?


iliabylich:
So, Prism is out and it's:

1. correct
2. comes out of the box with MRI (and IIRC with other Ruby implementations)
3. faster (I think on TruffleRuby/JRuby it's absurdly faster because it doesn't run in interpreted mode like it does with `parser`)
4. always up-to-date
5. written in a language with perfect interop capabilities

On the other side backporting changes to `parser` becomes more and more difficult. MRI doesn't use yacc/bison anymore, instead there's Lrama that is also an LALR(1) parser generator but it has more "syntax sugar" that Racc doesn't have, and so the grammar will diverge more and more over time. And if (or once) `parse.y` is deprecated it will be even more complex to "translate" grammar changes.

I personally believe that `parser` should be soft-deprecated. What do you guys think? @kddnewton @bbatsov @koic @palkan @marcandre @mbj

And if you agree do you see anything we could do to `parser` to make transition easier?

mbj:
The real use case of this gem left for me is multi version capability. Prism is not shipped with all the ruby versions my use cases still have to support, basically all non EOL rubies.

If this gem where to stop supporting future ruby versions I'd be fine. Cannot speak for the rest of the ecosystem OFC.

kddnewton:
In general I think this is the right long-term direction, but I'm uncomfortable soft deprecating until we mirror the tree rewriting functionality, since so many consumers use that. I think it could be added to the Ruby API of Prism _relatively_ easily, but I haven't had time to look into this yet.

I'm not sure if there are any other APIs that I'm not thinking about, but that's the main one I know we are missing.

bbatsov:
Indeed. I don't see how we can deprecated `parser` without the rewriting functionality. RuboCop also relies on the multi-version capability quite a lot, although I guess we can live without it if push comes to shove. 

One other question - if `parser` gets soft deprecated will it still be updated to track newer Ruby releases? I think that a realistic time-frame for full migration to Prism is probably 2-3 years (for the projects that are still active, that is), so `parser` will definitely require some maintenance is the mean time.

iliabylich:
> One other question - if parser gets soft deprecated will it still be updated to track newer Ruby releases?

My initial thought was to stop doing it. At this point `parser` looks to me a lot like a polyfill and extending it more to be compliant with new grammar features feels like a disservice to the community.

bbatsov:
[...]

PR #1069, opened 2025-03-13
Document the current state of the project in the readme

Earlopain:
Because of https://github.com/whitequark/parser/issues/1046. It will look like this:

> [!WARNING]  
> The `parser` gem can only correctly parse Ruby code up to and including Ruby 3.3. For Ruby 3.4 and later, please use the [Prism translation parser](https://github.com/ruby/prism/blob/main/docs/parser_translation.md) instead.
> Starting in Ruby 3.4, Prism is the parser used in Ruby itself and can produces AST that is identical to the output of `parser`. If you only need to parse Ruby 3.3 (or greater), also consider using the native Prism AST.
> See this [GitHub issue](https://github.com/whitequark/parser/issues/1046) for more details.

I think this makes sense?

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,11 @@ equivalent source code from Parser's ASTs.
 Sponsored by [Evil Martians](http://evilmartians.com).
 MacRuby and RubyMotion support sponsored by [CodeClimate](http://codeclimate.com).
 
+> [!WARNING]  
+> The `parser` gem is only compatible with the syntax of Ruby 3.3 and lower. For Ruby 3.4 and later, please use the [`Prism::Translation::Parser`](https://github.com/ruby/prism/blob/main/docs/parser_translation.md) instead.
+> Starting in Ruby 3.4, Prism is the parser used in Ruby itself and can produce AST that is identical to the output of the `parser` gem. If you only need to parse Ruby 3.3 (or greater) and don't require compatibility with the `parser` gem AST, also consider using the native Prism AST.
+> See this [GitHub issue](https://github.com/whitequark/parser/issues/1046) for more details.
+
 ## Installation
 
     $ gem install parser


iliabylich:
Yes, LGTM 👍

README.md
```diff
@@ -13,6 +13,11 @@ equivalent source code from Parser's ASTs.
 Sponsored by [Evil Martians](http://evilmartians.com).
 MacRuby and RubyMotion support sponsored by [CodeClimate](http://codeclimate.com).
 
+> [!WARNING]  
+> The `parser` gem can only correctly parse Ruby code up to and including Ruby 3.3. For Ruby 3.4 and later, please use the [Prism translation parser](https://github.com/ruby/prism/blob/main/docs/parser_translation.md) instead.
```
koic, line 17: That's just my two cents.

```suggestion
> The `parser` gem is only compatible parse Ruby code up to and including Ruby 3.3. For Ruby 3.4 and later, please use the [`Prism::Translation::Parser`](https://github.com/ruby/prism/blob/main/docs/parser_translation.md) instead.
```

README.md
```diff
@@ -13,6 +13,11 @@ equivalent source code from Parser's ASTs.
...
+> Starting in Ruby 3.4, Prism is the parser used in Ruby itself and can produces AST that is identical to the output of `parser`. If you only need to parse Ruby 3.3 (or greater), also consider using the native Prism AST.
```
koic, line 18: It might be helpful to add a note clarifying the differences in compatibility between the respective ASTs.

```suggestion
> Starting in Ruby 3.4, Prism is the parser used in Ruby itself and can produces AST that is identical to the output of `parser`. If you only need to parse Ruby 3.3 (or greater) and don't require compatibility with the `parser` gem AST, also consider using the native Prism AST.
```

README.md
Earlopain: I tweaked the first part a bit since a word was missing otherwise

iliabylich:
Let's merge it as is and tweak later?

koic:
Yeah, let's proceed with that.

iliabylich:
Thanks guys!

Earlopain:
Yes, that PR looks ok to me. One not thought, that `Prism::Translation::Parser` is not equivalent to `Parser::CurrentRuby` because it always parses as Ruby 3.4. For the tests, I don't believe that is relevant but I wanted to mention it. You can refer to it via `Prism::Translation::Parser34` to make that clear.

I wrote https://github.com/whitequark/parser/pull/1072 which should cover the most relevant cases. Just recently I also added something like `CurrentRuby` to prism, once that is released I will update the guide to mention that.

And yeah, unfortunately no rewriting with native prism for now as far as I'm aware https://github.com/ruby/prism/issues/3466

BuonOmo:
So maybe we could add a reference to this doc in the warning message, and add a word about rewriting? I could do it next week :)

Earlopain:
Sure, I think it could make sense to add a more detailed warning message specific to 3.4. I'd add a header to the warning in the readme so its linkable and put that in the warning somewhere. Feel free to open a PR.

Earlopain:
Also, I haven't looked much into rewriting yet and am not even sure how it works for parser. But I think it would be nice if you could mention in the prism issue that you need that in order to fully switch to prism

BuonOmo:
> the prism issue

Is it #1046 ? I already saw that people mentionned it, with great ideas (a separate gem for rewriting for instance). I :+1:ed the messages, and didn't feel like we'd need one more message 😅

Earlopain:
Nah, I meant this one https://github.com/ruby/prism/issues/3466, I linked it above.

I'm somewhat sure that the prism maintainer is interested in that as a first-party functionality. I think that has come in the issue you linked sometime.

fschwahn:
A question about that: couldn't the `parser`-gem internally call prism for ruby 3.4+, until the ecosystem has moved on from `parser`? I currently see `warning: parser/current is loading parser/ruby34, which recognizes 3.4.0-dev-compliant syntax, but you are running 3.4.7.` when running for example `erb_lint`.

PR #1071, opened 2025-03-19
+ add prism-specific node types

Earlopain:
It's a bit akward: rubocop and rubocop-ast use this list to generate visitors and node type predicates. Since they both use this, it doesn't make much sense to add this handling in rubocop I think.

On the other hand, adding this to prism doesn't really work since the parser there is lazily loaded, and by rubocop as well since users may just use older ruby versions. Prism itself has no dependecy on parser to eagerly add to this list.

I think adding this to parser is ok, and is also the simplest solution to the problem. The list can (and does) contain entries that are parser-version specific, this just extends that policy a bit farther than it used to be.

cc @koic

### DIFF
--- a/lib/parser/meta.rb
+++ b/lib/parser/meta.rb
@@ -4,9 +4,17 @@ module Parser
   # Parser metadata
   module Meta
 
+    # These are node types required by `Prism::Translation::Parser`, 
+    # which has advanced syntax support ahead of the Parser gem.
+    PRISM_TRANSLATION_PARSER_NODE_TYPES = %i(
+      itarg itblock
+    ).freeze
+    private_constant :PRISM_TRANSLATION_PARSER_NODE_TYPES
+
     # All node types that parser can produce. Not all parser versions
     # will be able to produce every possible node.
-    NODE_TYPES =
+    # Includes node types that are only emitted by the prism parser translator.
+    NODE_TYPES = (
       %i(
         true false nil int float str dstr
         sym dsym xstr regopt regexp array splat
@@ -34,7 +42,7 @@ module Parser
         empty_else find_pattern kwargs
         match_pattern_p match_pattern
         forwarded_restarg forwarded_kwrestarg
-      ).to_set.freeze
-
+      ) + PRISM_TRANSLATION_PARSER_NODE_TYPES
+    ).to_set.freeze
   end # Meta
 end # Parser


lib/parser/meta.rb
```diff
@@ -34,6 +35,8 @@ module class sclass def defs undef alias args
         empty_else find_pattern kwargs
         match_pattern_p match_pattern
         forwarded_restarg forwarded_kwrestarg
+        
+        itarg itblock
       ).to_set.freeze
```
koic, line 40: The approach looks reasonable. For the implementation, it would be better to separate constant instead of relying on blank line for chunk division. e.g,:

```ruby
# They are node types required by `Prism::Translation::Parser`, 
# which has advanced syntax support ahead of the Parser gem.
PRISM_TRANSLATOR_PARSER_NODE_TYPES = %i(
  itarg itblock
).freeze
private_constant :PRISM_TRANSLATOR_PARSER_NODE_TYPES

NODE_TYPES = (
  %i(
    ...
  ) + PRISM_TRANSLATOR_PARSER_NODE_TYPES
).to_set.freeze
```

Earlopain: That's nice, thanks. I wanted to put a comment there but that didn't really work.

Earlopain:
I'm adding `itblock` to `rubocop-ast` at the moment, so no need for a release rightaway in case I find something else like this.

lib/parser/meta.rb
```diff
@@ -34,7 +42,7 @@ module class sclass def defs undef alias args
         empty_else find_pattern kwargs
         match_pattern_p match_pattern
         forwarded_restarg forwarded_kwrestarg
-      ).to_set.freeze
-
+      ) + PRISM_TRANSLATOR_PARSER_NODE_TYPES
```
koic, line 45:
```suggestion
      ) + PRISM_TRANSLATION_PARSER_NODE_TYPES
```

lib/parser/meta.rb
```diff
@@ -4,9 +4,17 @@ module Parser
   # Parser metadata
   module Meta
 
+    # These are node types required by `Prism::Translation::Parser`, 
+    # which has advanced syntax support ahead of the Parser gem.
+    PRISM_TRANSLATOR_PARSER_NODE_TYPES = %i(
+      itarg itblock
+    ).freeze
+    private_constant :PRISM_TRANSLATOR_PARSER_NODE_TYPES
```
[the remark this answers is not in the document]
Earlopain: Oops, good catch

Earlopain:
I think this is the only thing needed, https://github.com/rubocop/rubocop-ast/pull/365 seems to work fine when this PR is pulled in.

bbatsov:
I'm also OK with the proposed approach here, but I think we should also mention this in Parser's docs, so that other Parser consumers are aware of the migration strategy to Parser.

Earlopain:
Hm, were would you mention this? This list has no real effect on users of `parser` (for example, it still contains implementaion-specific node types of alternative ruby implementations like macruby. It can contain anything really, and is not even used in parser itself, expect in tests to check it is exhaustive.

I added a notice to the readme about prism and such last week, did you have something like that in mind?

bbatsov:
> I added a notice to the readme about prism and such last week, did you have something like that in mind?

Yeah, I saw it, but I was thinking more of something like adding a section "Using Parser with Ruby 3.4+" or something along those lines that mentions something of things we've been discussing in the context of RuboCop. I think that'd be useful for every project currently using Parser.

Earlopain:
Ah, right. Yes, I think that would make sense. I can write something up in a separate PR.

koic:
To advance https://github.com/rubocop/rubocop-ast/pull/365 and resolve https://github.com/rubocop/rubocop-ast/issues/366, this PR will be merged and released. Documentation can be handled separately in https://github.com/whitequark/parser/pull/1072.

bbatsov:
@koic Sure, feel free to cut a new release.

Earlopain:
It looks like @iliabylich has `ruby28.rb` from a few years ago still around, so now the new release is 40kb lighter (:

iliabylich:
True 😄

PR #1072, opened 2025-03-19
Document parser/prism migration path

Earlopain:
Ref https://github.com/whitequark/parser/pull/1071#issuecomment-2736283700

cc @bbatsov

I'm not a good technical writer so this can probably use some grease here and there. I think I've got the main points that can come up. I also added `itblock` to the ast docs and marked it was prism-only.

This only focuses on the translation layer, not on using prism standalone.


Earlopain:
I openend https://github.com/ruby/prism/pull/3508 to add something like `Parser::Current` to prism since it currently has nothing like that.

bbatsov:
The doc updates look great to me. Thanks!

PR #1076, opened 2025-04-08
* YARD docs for Parser::CurrentRuby and Parser::Base#version

apiology:
Documentation changes:

* Add YARD docs for `Parser::CurrentRuby`:
  * Document the transition for users seeking Ruby 3.4 support
  * Allow IDE tools like Solargraph to understand the invocation and results from Parser::Current.parse() to provide auto-completion and type checking

* Add type info for `Parser::Base#version`


iliabylich:
Thanks!

PR #1078, opened 2025-05-06
Add a few more type annotations to Parser::Builders::Default

apiology:


lib/parser/builders/default.rb
```diff
@@ -2295,10 +2295,14 @@ def collapse_string_parts?(parts)
           [:str, :dstr].include?(parts.first.type)
     end
 
+    # @param [::Parser::AST::Node] token
```
iliabylich, line 2298: Sorry for a late review but I think this signature is wrong. This method takes a token, not a node. Same for the method below.

